PR #5, opened 2017-10-11
EZP-24699: Enrich policy limitations view with human readable names

adamwojs:
> https://jira.ez.no/browse/EZP-24699
> **Requires merge first**: ~https://github.com/ezsystems/repository-forms/pull/159~ and https://github.com/ezsystems/repository-forms/pull/164

## Description 

This PR makes the values of policy limitations readable to humans. See https://github.com/ezsystems/repository-forms/pull/159 for more details.

zrzut ekranu z 2017-10-11 20-21-51

### DIFF
--- a/src/bundle/Resources/views/admin/policy/list.html.twig
+++ b/src/bundle/Resources/views/admin/policy/list.html.twig
@@ -24,13 +24,20 @@
                 <td>{{ policy.module }}</td>
                 <td>{{ policy.function }}</td>
                 <td>
-                    {% for limitation in policy.limitations %}
-                        {{ limitation.identifier }}
-
-                        ({{ limitation.limitationValues | join(',') }})
-                    {% else %}
-                        No limitations
-                    {% endfor %}
+                    {%- if policy.limitations is not empty -%}
+                        <ul class="list-unstyled">
+                            {%- for limitation in policy.limitations -%}
+                                <li>
+                                    <span class="font-weight-bold" title="{{ 'policy.limitation.identifier_tooltip' | trans({'%identifier%': limitation.identifier}) }}">
+                                        {{ ('policy.limitation.identifier.' ~ limitation.identifier | lower) | trans({}, 'ezrepoforms_policies') }}:
+                                    </span>
+                                    {{ ez_render_limitation_value(limitation) }}
+                                </li>
+                            {%- endfor -%}
+                        </ul>
+                    {%- else -%}
+                        {{- 'policy.limitation.none'|trans -}}
+                    {%- endif -%}
                 </td>
                 <td class="text-right">
                     <a href="{{ path('ezplatform.policy.update', { roleId: role.id, policyId: policy.id }) }}" class="btn btn-info">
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/views/admin/policy/list.html.twig
```diff
@@ -25,9 +25,9 @@
                 <td>{{ policy.function }}</td>
                 <td>
                     {% for limitation in policy.limitations %}
-                        {{ limitation.identifier }}
-
-                        ({{ limitation.limitationValues | join(',') }})
+                        <strong>{{ limitation.identifier }}</strong>:
```
dew326: `<strong>` tag has own meaning, which is "this is more important than other text". I think it's not that case.
I suggest using just `<span>` with css class to bold text.

src/bundle/Resources/views/admin/policy/list.html.twig
```diff
@@ -25,9 +25,9 @@
...
+                        {{ ez_render_limitation_value(limitation) }}
+                        {% if not loop.last %}<br />{% endif %}
```
dew326: `<br>` it to break line of solid text, in this case, I believe each line is own statement. If we look at this as a list of limitation we can do something like: 
```html
<ul>
    <li><span>{{ limitation.identifier }}</span>: {{ ez_render_limitation_value(limitation) }}</li>
    <li><span>{{ limitation.identifier }}</span>: {{ ez_render_limitation_value(limitation) }}</li>
</ul>
```

Some CSS classes might be needed to style.

src/bundle/Resources/views/admin/policy/list.html.twig
sunpietro: Another option would be using `<b></b>` tag :)

lserwatka:
@adamwojs your PR to repository-forms was merged, could you address comments here?

adamwojs:
PR updated according to @sunpietro and @dew326 suggestions.

src/bundle/Resources/views/admin/policy/list.html.twig
```diff
@@ -24,13 +24,20 @@
...
-                        {{ limitation.identifier }}
-
-                        ({{ limitation.limitationValues | join(',') }})
-                    {% else %}
-                        No limitations
-                    {% endfor %}
+                    {%- if policy.limitations is not empty -%}
+                        <ul class="list-unstyled">
+                            {%- for limitation in policy.limitations -%}
+                            <li>
```
dew326: is this indentation ok?

lserwatka:
@adamwojs what else is needed here to merge your PR?

bdunogier:
For what it's worth, I really like the approach of using the translation system for that. Simple, and very flexible. I have suggested the same approach for another feature recently.

👍

SylvainGuittard:
@bdunogier FYI, translation has also been discussed on the other PR for PlatformUI (https://github.com/ezsystems/PlatformUIBundle/pull/907#issuecomment-336053058).
There is a follow up story: https://jira.ez.no/browse/EZP-28072

adamwojs:
PR updated, but it requires https://github.com/ezsystems/repository-forms/pull/164 to be merged first.

PR #8, opened 2017-10-13
EZP-27934: Remove CacheService now that spi cache got fix

andrerom:
> Issue: https://jira.ez.no/browse/EZP-27934


lserwatka:
@andrerom could share a ticket where this was fixed in kernel? This is great news.

andrerom:
@lserwatka Handled it as part of this issue as it was a bit of mix of everything, and as 7.0 kernel is still rebased for clean history until first public beta release. But the kernel change was addtiona of these lines: https://github.com/ezsystems/ezpublish-kernel/blob/7.0/eZ/Publish/Core/Persistence/Cache/ContentTypeHandler.php#L242-L244

lserwatka:
@Nattfarinn I think we can merge this one.

Nattfarinn:
(Removing ezrobot complain about coding standards, as it's fixed on master already)

PR #10, opened 2017-10-16
Added bottom spacing to the location view

sunpietro:
Added bottom spacing to the location view, so the **Show more results** button doesn't stick to the bottom.

### DIFF
--- a/src/bundle/Resources/views/content/locationview.html.twig
+++ b/src/bundle/Resources/views/content/locationview.html.twig
@@ -34,7 +34,7 @@
                 Trash
             </a>
         </div>
-        <div class="col-sm-10 px-0">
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/views/content/locationview.html.twig
```diff
@@ -34,7 +34,7 @@
                 Trash
             </a>
         </div>
-        <div class="col-sm-10 px-0">
+        <div class="col-sm-10 px-0 ez-location-info">
```
dew326: If we already have `px-0` here I would go and use `pb-4` instead of introducing new css class.

lserwatka:
@Nattfarinn we can merge this one too.

PR #11, opened 2017-10-16
EZP-27890: As a Editor, I want to edit a Text Block field type with the new UI guidelines

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-27890

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/eztext.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/eztext.js
@@ -1,0 +1,40 @@
+(function (global) {
+    const SELECTOR_FIELD = '.ez-field-edit-eztext';
+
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/scripts/fieldType/eztext.js
```diff
@@ -0,0 +1,40 @@
...
+        /**
+         * Validates the textarea field value
+         *
+         * @method validateInput
+         * @param {Event} event
+         * @returns {Object}
+         * @memberof EzTextValidator
+         */
+        validateInput(event) {
+            const isError = event.target.required && !event.target.value;
```
sunpietro: What about multiple spaces? Maybe `trim()`?

PR #12, opened 2017-10-17
EZP-27909: As an Editor, I want to edit a URL field type with the new UI guidelines

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-27909


src/bundle/Resources/views/layout.html.twig
```diff
@@ -10,8 +10,10 @@
                 // needs translations
                 errors: {
                     emailRegexp: /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/,
+                    urlRegexp: /(https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|www\.[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9]\.[^\s]{2,}|www\.[a-zA-Z0-9]\.[^\s]{2,})/,
```
adamwojs: Are you sure that this regular expression is consistent with validation on the backend? https://github.com/ezsystems/ezpublish-kernel-ee/pull/161/files#diff-981d67ab6f6be0f1d1ff989637c9e357R20

dew326: We are aware this should be provided by backend (not hardcoded in the twig).

PR #23, opened 2017-10-19
EZP-27938: Add tests and perform cleanup in RequestListener

mikadamczyk:
> https://jira.ez.no/browse/EZP-27938

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -1,0 +1,1 @@
+<?php require_once __DIR__ . '/vendor/autoload.php';
--- a/src/lib/EventListener/RequestListener.php
+++ b/src/lib/EventListener/RequestListener.php
@@ -7,7 +7,6 @@
 namespace EzSystems\EzPlatformAdminUi\EventListener;
 
 use eZ\Publish\Core\MVC\Symfony\SiteAccess;
-use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
@@ -16,7 +15,16 @@ use Symfony\Component\HttpKernel\KernelEvents;
 
 class RequestListener implements EventSubscriberInterface
 {
-    use ContainerAwareTrait;
+    /** @var array */
+    private $groupsBySiteAccess;
+
+    /**
+     * @param $groupsBySiteAccess
+     */
+    public function __construct(array $groupsBySiteAccess)
+    {
+        $this->groupsBySiteAccess = $groupsBySiteAccess;
+    }
 
     /**
      * Returns an array of event names this subscriber wants to listen to.
@@ -36,7 +44,7 @@ class RequestListener implements EventSubscriberInterface
      *
      * @return array The event names to listen to
      */
-    public static function getSubscribedEvents()
+    public static function getSubscribedEvents(): array
     {
         return [
             KernelEvents::REQUEST => ['onKernelRequest', 13],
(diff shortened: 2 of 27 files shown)


bootstrap.php
```diff
@@ -0,0 +1 @@
+<?php require_once __DIR__ . '/vendor/autoload.php';
```
andrerom: can be placed in test folder, example: https://github.com/ezsystems/repository-forms/blob/master/tests/bootstrap.php

src/lib/EventListener/RequestListener.php
```diff
@@ -16,7 +15,12 @@
 
 class RequestListener implements EventSubscriberInterface
 {
-    use ContainerAwareTrait;
+    private $groupsBySiteAccess;
+
+    public function __construct($groupsBySiteAccess)
+    {
+        $this->groupsBySiteAccess = $groupsBySiteAccess;
```
andrerom: note that siteaccess can change. If we don't take from container, ideally class should be adapted to expose setter to refresh setting on siteaccess changes, there are some examples around but I'm not the export on this.

bdunogier: I'm not sure that is relevant in this context: what gets injected is the list of groups per siteaccess, not the _current_ siteaccess.

The current siteaccess may change dynamically, but the list of siteaccess/siteaccess groups may not, right ?

bootstrap.php
mikadamczyk: yes, but possibly we can have tests also in /src/bundle directory, then the file would be duplicated

src/lib/EventListener/RequestListener.php
mikadamczyk: as I know, current SiteAccess can be changed, but here we fetch only list all of available groups by SiteAccess. Can it be changed during master request?

andrerom: you are right, if it is the global map then it is not subject to issues on this.

bdunogier: Based on the name, it is the list of groups for each SA.

mikadamczyk: I think it is a global map, not  groups of the current SiteAccess

Nattfarinn:
Please rebase with master (conflicts). :)

src/lib/Tests/EventListener/RequestListenerTest.php
```diff
@@ -0,0 +1,106 @@
...
+     */
+    private $request;
+
+    /**
+     * @var GetResponseEvent
+     */
+    private $event;
+
+    /**
+     * @var HttpKernelInterface|\PHPUnit_Framework_MockObject_MockObject
```
webhdx: CS, `/** @var (...) */` one liner as above.

src/lib/Tests/EventListener/RequestListenerTest.php
```diff
@@ -0,0 +1,106 @@
...
+            $this->request,
+            HttpKernelInterface::SUB_REQUEST
+        );
+        $this->assertNull($this->requestListener->onKernelRequest($this->event));
+    }
+
+    public function testOnKernelRequestAllowAccessWithoutSiteAccess()
+    {
+        $this->request->attributes->set('siteaccess', 'not_siteaccess_object');
+        $this->assertNull($this->requestListener->onKernelRequest($this->event));
```
webhdx: Empty line. Also check other occurrences below.

mikadamczyk:
Hi @webhdx, I updated PR.

src/lib/EventListener/RequestListener.php
```diff
@@ -16,7 +15,12 @@
 
 class RequestListener implements EventSubscriberInterface
 {
-    use ContainerAwareTrait;
+    private $groupsBySiteAccess;
```
webhdx: As far as I understand this is an array, please enclose type hinting information.

src/lib/EventListener/RequestListener.php
```diff
@@ -16,7 +15,12 @@
...
+
+    public function __construct($groupsBySiteAccess)
```
webhdx: Type hinting + docblock

src/lib/Tests/EventListener/RequestListenerTest.php
```diff
@@ -0,0 +1,100 @@
...
+    /** @var RequestListener */
+    private $requestListener;
+
+    /** @var Request */
+    private $request;
+
+    /** @var GetResponseEvent */
+    private $event;
+
+    /** @var HttpKernelInterface|\PHPUnit_Framework_MockObject_MockObject */
```
webhdx: New PHPUnit has namespaces support which we should use.

mikadamczyk: As I see, createMock() return PHPUnit_Framework_MockObject_MockObject

PR #24, opened 2017-10-20
EZP-27938: Add missing ez-support-tools dependency, tests and perform cleanup in SystemInfoTabGroupListener

mikadamczyk:
> https://jira.ez.no/browse/EZP-27938

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -1,0 +1,1 @@
+<?php require_once __DIR__ . '/vendor/autoload.php';
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -1,0 +1,26 @@
+<phpunit
+        backupGlobals="false"
+        backupStaticAttributes="false"
+        bootstrap="bootstrap.php"
+        convertErrorsToExceptions="true"
+        convertNoticesToExceptions="true"
+        convertWarningsToExceptions="true"
+        colors="true">
+    <testsuites>
+        <testsuite name="EzSystems\EzPlatformAdminUiBundle">
+            <directory>src/bundle/Tests/</directory>
+        </testsuite>
+        <testsuite name="EzSystems\EzPlatformAdminUi">
+            <directory>src/lib/Tests/</directory>
+        </testsuite>
+    </testsuites>
+    <filter>
+        <whitelist>
+            <directory>src</directory>
+            <exclude>
+                <directory>src/bundle/Tests/</directory>
+                <directory>src/lib/Tests/</directory>
+            </exclude>
+        </whitelist>
+    </filter>
+</phpunit>
--- a/src/lib/Tests/EventListener/SystemInfoTabGroupListenerTest.php
+++ b/src/lib/Tests/EventListener/SystemInfoTabGroupListenerTest.php
@@ -1,0 +1,120 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformAdminUi\Tests\EventListener;
+
+use EzSystems\EzPlatformAdminUi\EventListener\SystemInfoTabGroupListener;
+use EzSystems\EzPlatformAdminUi\Tab\SystemInfo\SystemInfoTab;
+use EzSystems\EzPlatformAdminUi\Tab\TabGroup;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+use PHPUnit\Framework\TestCase;
+use EzSystems\EzPlatformAdminUi\Tab\Event\TabEvents;
+use EzSystems\EzPlatformAdminUi\Tab\Event\TabGroupEvent;
+use EzSystems\EzPlatformAdminUi\Tab\SystemInfo\TabFactory;
+use EzSystems\EzPlatformAdminUi\Tab\TabRegistry;
+use EzSystems\EzSupportToolsBundle\SystemInfo\SystemInfoCollectorRegistry;
+use PHPUnit_Framework_MockObject_MockObject as MockObject;
+
+class SystemInfoTabGroupListenerTest extends TestCase
+{
+    /** @var Request */
+    private $request;
+
+    /** @var TabGroupEvent */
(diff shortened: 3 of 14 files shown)


bootstrap.php
```diff
@@ -0,0 +1 @@
+<?php require_once __DIR__ . '/vendor/autoload.php';
```
webhdx: It lacks `@copyright` but my question is different: Do we need `bootstrap.php`? You can just point to the `autoload.php` which will reduce file overhead.

phpunit.xml
```diff
@@ -0,0 +1,29 @@
+<phpunit
+        backupGlobals="false"
+        backupStaticAttributes="false"
+        bootstrap="bootstrap.php"
+        convertErrorsToExceptions="true"
+        convertNoticesToExceptions="true"
+        convertWarningsToExceptions="true"
+        colors="true">
+    <php>
+        <ini name="error_reporting" value="-1"/>
```
webhdx: I'm not sure about this one. I'd not suppress errors completely as this is also a good indicator that some parts of the code may be deprecated or invalid.

src/lib/Tests/EventListener/SystemInfoTabGroupListenerTest.php
```diff
@@ -0,0 +1,116 @@
...
+     */
+    private $request;
+
+    /**
+     * @var TabGroupEvent
+     */
+    private $event;
+
+    /**
+     * @var HttpKernelInterface|\PHPUnit_Framework_MockObject_MockObject
```
webhdx: Do we use PHPUnit 6.x? I believe it brings the support for namespaces.

src/lib/Tests/EventListener/SystemInfoTabGroupListenerTest.php
```diff
@@ -0,0 +1,116 @@
...
+        $this->event->setData($tabGroup);
+        $systemInfoTabGroupListener->onTabGroupPreRender($this->event);
+    }
+
+    /**
+     * @dataProvider dataProvider
+     */
+    public function testOnTabGroupPreRender($identifiers)
+    {
+        $tabRegistry = $this->createMock(TabRegistry::class);
```
adamwojs: You have already created `TabRegistry` mock  in `setUp()`

src/lib/Tests/EventListener/SystemInfoTabGroupListenerTest.php
```diff
@@ -0,0 +1,116 @@
...
+        $tabRegistry->expects(self::exactly(count($identifiers)))
+            ->method('addTab');
+        $tabFactory = $this->createMock(TabFactory::class);
```
adamwojs: You have already created `TabFactory` mock in `setUp()`

src/lib/Tests/EventListener/SystemInfoTabGroupListenerTest.php
```diff
@@ -0,0 +1,116 @@
...
+        $tabFactory->expects(self::exactly(count($identifiers)))
+            ->method('createTab')
+            ->willReturn($this->createMock(SystemInfoTab::class));
```
adamwojs: As we talk on last Friday, this test will be more valuable if it will check that the proper tabs were created and registered. I'm not sure if there is the better way to write this, but following code snippet should be acceptable:


```php
foreach ($identifiers as $i => $identifier) {
    $tab = $this->createMock(SystemInfoTab::class);

    $this->tabFactory
        ->expects($this->at($i))
        ->method('createTab')
        ->with($identifier)
        ->willReturn($tab);

    $this->tabRegistry
        ->expects($this->at($i))
        ->method('addTab')
        ->with($tab, 'systeminfo');
}
```

src/lib/Tests/EventListener/SystemInfoTabGroupListenerTest.php
andrerom: yes, please aim for PHPUNit 5.7+ support and use the namespeaces so we don't have to change it later. All other repos are current undergoing lots of changes for this..

phpunit.xml
mikadamczyk: Of course I can change it, but in ez packages and symfony, in phpunit.xml files is the same config

bootstrap.php
mikadamczyk: It's just a convention. In another phpunit.xml file is the same. @Nattfarinn, @webhdx  what should we do?

src/lib/Tests/EventListener/SystemInfoTabGroupListenerTest.php
mikadamczyk: ok, I will change this.

mikadamczyk:
Hi, I updated PR

lserwatka:
@mikadamczyk fixer jeszcze coś znalazł :)

mikadamczyk:
Tak, właśnie poprawiłem, muszę sobie wyrobić dobre odruchy ;)

lserwatka:
@Nattfarinn @webhdx @adamwojs are you up for final review?

src/lib/Tests/EventListener/SystemInfoTabGroupListenerTest.php
```diff
@@ -0,0 +1,120 @@
...
+
+class SystemInfoTabGroupListenerTest extends TestCase
+{
+    /** @var Request */
+    private $request;
+
+    /** @var TabGroupEvent */
+    private $event;
+
+    /** @var HttpKernelInterface|PHPUnit_Framework_MockObject_MockObject */
```
webhdx: Can we use namespaces? I believe we are using PHPUnit 6.x which brought support for these.

phpunit.xml
lserwatka: I think we should remove this line.

src/lib/Tests/EventListener/SystemInfoTabGroupListenerTest.php
mikadamczyk: Of course, I will change it

phpunit.xml
mikadamczyk: OK, no problem :)

mikadamczyk:
I fixed last remarks @webhdx

PR #25, opened 2017-10-24
EZP-28092 As an editor I want an Edit content interaction including the choice of the language

sunpietro:
> JIRA: https://jira.ez.no/browse/EZP-28092

ezp-2811

The GIF doesn't show the fancy animation the button panel.


src/bundle/Resources/public/js/scripts/sidebar/extra.actions.btn.js
```diff
@@ -0,0 +1,9 @@
+(function () {
+    const btns = document.querySelectorAll('.ez-btn--extra-actions');
+
+    btns.forEach(btn => {
```
dew326: Is forEach on NodeList finally supported by all browser or you are missing `[...btn]`?

src/bundle/Resources/public/scss/_custom.scss
```diff
@@ -104,3 +104,6 @@ $link-color: $ez-color-hyperlink;
 
 // Body
 $body-bg: $ez-ground-base-pale;
+
+// Transition function
+$admin-transition: cubic-bezier(.25,.8,.25,1);
```
dew326: Here are bootstrap variables. I would suggest to put it on top of the file and use prefixed name `$ez-admin-transition`

src/bundle/Resources/public/scss/_buttons-extra-actions.scss
```diff
@@ -0,0 +1,75 @@
...
+}
+
+.ez-btn-container--extra-actions {
+    .ez-extra-actions {
+        position: absolute;
+        left: 0;
+        top: 50%;
+        z-index: 2;
+        opacity: 1;
+        transform: translate3d(calc(-100% - 30px), -50%, 0) scaleX(1);
```
dew326: I know that `translate3d` is better for performance but it also makes text blurry. I think we have to stick with `translate`.

src/bundle/Resources/public/js/scripts/sidebar/extra.actions.btn.js
sunpietro: I've found an interesting approach here: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/10638731/#comment-5

Do you agree with this proposal? IMO it's a bit risky (always adding/overriding native methods is risky).

src/bundle/Resources/public/scss/_buttons-extra-actions.scss
sunpietro: Didn't know about it. I'll change it.

sunpietro:
ping @dew326

src/bundle/Resources/public/js/scripts/sidebar/extra.actions.btn.js
dew326: I also think this kind of hack is risky and would avoid if can. Here we easily can do `[...btns]` so I would stay with it.

sunpietro:
The implementation will be revised upon information from @webhdx that cannot apply this HTML markup of edit button with current codebase.

webhdx:
ping @dew326 @sunpietro I rewritten the history separating frontend from backend.

Anyone else: Please start your code reviews.

m-tyrala:
QA test ok.

barbaragr:
I've found some issues:
1. Click "Create" button
2. Choose one content type, for example "Article", form is opened
3. Go back with web browser's "Back" button
4. Type a word in an input (menu of "Create" btn is still opened), make sure that nothing will be found.
5. Use "Enter" key and the application will go to create Article form again.

In addition: when you type something in input and close the "Create" button menu, the word you've typed is still there. Maybe this field should be cleared. In reference to this, we can move forward with steps above:
6. Go back and delete manually content of the input.
7. Click once again on "Article" - nothing happened. You can click on another option, for instance "Folder" and it will work fine.

PR #35, opened 2017-11-01
Removed hardcoded language_code from content type create/edit/view templates

adamwojs:
> JIRA: -

## Description 

Removed hardcoded `language_code` from `src/bundle/Resources/admin/content_type/{create,edit,view}.html.twig`. 

Disclaimer: we don't support translating content types in UI yet, so in fact we always use `mainLanguageCode`


andrerom:
@adamwojs is this also present in 1.x? If so let's backport it

adamwojs:
@andrerom No, this isn't present in 1.X 😉

PR #38, opened 2017-11-02
EZP-28091: As an editor I want a new Create content interaction including the choice of the language

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28091
> Depends on: #25 

UI

# Description
This PR adds Create button in right sidebar on Location View page. Create widget allows to select a language (not visible on the above screen capture, explanation below) and a content type. It automatically redirects to content editing screen.
* If the system has only one language added, language selector is not visible.
* Button is being disabled if user has no permissions to create any content.
* Content type list is not filtered taking permissions into consideration due to inefficient Permission Resolver functionality in kernel.

# TODO:
 - [x] Rebase after #25 gets merged


webhdx:
Please review only f7a32e9 and a8f74e4 commits. Others are coming from the #25.

Nattfarinn:
@webhdx Rebase please. (https://github.com/ezsystems/ezplatform-admin-ui/pull/25 merged)

barbaragr:
I've found some issues:

1. Click "Create" button
2. Choose one content type, for example "Article", form is opened
3. Go back with web browser's "Back" button
4. Type a word in an input (menu of "Create" btn is still opened), make sure that nothing will be found.
5. Use "Enter" key and the application will go to create Article form again.
In addition: when you type something in input and close the "Create" button menu, the word you've typed is still there. Maybe this field should be cleared. In reference to this, we can move forward with steps above:
6. Go back and delete manually content of the input.
7. Click once again on "Article" - nothing happened. You can click on another option, for instance "Folder" and it will work fine.

webhdx:
@lserwatka pls merge

PR #39, opened 2017-11-03
EZP-28018: Add initial Behat structure

mnocon:
> JIRA: https://jira.ez.no/browse/EZP-28018
Requires: https://github.com/ezsystems/ezplatform/pull/223
# Description:

This PR adds the initial Behat project skeleton and 3 tests for the Login page.

*General idea:*
I've decided to partially implement Page Object pattern (without modelling each action based on their result, there are no "I login successfully" and "I fail to login" methods).

There are two types of Contexts:
* Utility context - a wrapper for Mink methods so that finding/waiting for Elements is easier
* Business contexts - contexts that have steps with logic involving interactions with the site

All files are placed in `src/lib/Tests/Behat` directory, I hope it's ok.

# Running
Set `base_url` and `wd_host` in behat.yml.dist to your v2 installation and Selenium Server URL (you can use selenium-chrome-debug:3.4.0 Docker image, same as in v1: https://hub.docker.com/r/selenium/standalone-chrome-debug/tags/)

Run with: `vendor/bin/behat --profile=default --suite=adminui`


mnocon:
@Nattfarinn @webhdx please have a look if you are ok with where the tests are placed (and feel free to look at the code as well!)

Nattfarinn:
@mnocon I think we should split `src/lib/Tests/Behat` into two separate directories:
 - `src/features` - scenarios only,
 - `src/lib/Behat` - php context handlers.

src/lib/Tests/Behat/BusinessContext.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    protected $utilityContext;
+
+    /** @BeforeScenario
+     * @param BeforeScenarioScope $scope Behat scope
+     */
+    public function getUtilityContext(BeforeScenarioScope $scope)
+    {
+        $environment = $scope->getEnvironment();
+        $this->utilityContext = $environment->getContext('EzSystems\EzPlatformAdminUi\Tests\Behat\UtilityContext');
```
Nattfarinn: I would use `UtilityContext::class` instead.

Nattfarinn:
Nitpicks: `void` return type missing in few places.

mnocon:
@Nattfarinn I've moved the files as suggested (and added missing return types) in https://github.com/ezsystems/ezplatform-admin-ui/pull/39/commits/078ca8266848cf48a15455ca77bb5a27b0061f10

bdunogier:
About the structure, did you consider having `features` at the same level than `src` ? It isn't really the bundle's source code. We could even use a different namespace for those elements so that they aren't part of the application's one.

mnocon:
Moving the features one level up sounds like a good idea, they could be stored there as well.

About moving them to a different namespace: do you mean creating a separate bundle for features?
I'm not a fan of this approach, as I'm not sure about the advantages and it complicates versioning a bit.

We will be having an internal (QA team) discussion about the test code structure, I can let you know about its outcome if you'd like.

PR #40, opened 2017-11-04
EZP-28173: Return value of EzSystems\...\LanguageCreateData::isEnabled() must be of the type boolean, null returned

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-28173

## Description

This PR provide fix for the following error: `Return value of EzSystems\EzPlatformAdminUi\Form\Data\Language\LanguageCreateData::isEnabled() must be of the type boolean, null returned` on create language form.  Besides of this, by default language should be enabled.


lserwatka:
Thank you @adamwojs  :)

PR #41, opened 2017-11-06
EZP-28184: Location View / Relation Tab is broken

webhdx:
>JIRA: https://jira.ez.no/browse/EZP-28184

# Description
This PR fixes incorrect calls to `ValueFactory::createRelation` as well as some issues in twig i.e. incorrect constants access (single slashes instead of double), wrong bundle name in include template calls.

### DIFF
--- a/src/lib/UI/Dataset/RelationsDataset.php
+++ b/src/lib/UI/Dataset/RelationsDataset.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 namespace EzSystems\EzPlatformAdminUi\UI\Dataset;
 
 use eZ\Publish\API\Repository\ContentService;
+use eZ\Publish\API\Repository\Exceptions\UnauthorizedException;
 use eZ\Publish\API\Repository\Values\Content\VersionInfo;
 use EzSystems\EzPlatformAdminUi\UI\Value\ValueFactory;
 use EzSystems\EzPlatformAdminUi\UI\Value as UIValue;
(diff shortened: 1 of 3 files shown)


bdunogier:
Not directly related to this PR, but while reading the code, I've stumbled upon a few elements that could be improved. I'll list them here for future discussions, if they seem valid to you.

The Dataset classes, such as `VersionsDataset`, violate the newable vs injectable principle: they're both holding values and receiving injected services.

About the custom value objects. First, it is perfectly appropriate to use them, as they greatly simplify the application's code. What I'd suggest we do differently is the loading of those. A class such as [lib/UI/Value/ValueFactory.php](https://github.com/webhdx/ezplatform-admin-ui/blob/master/src/lib/UI/Value/ValueFactory.php) will most likely grow out of control in terms of number of methods, as we need to load more custom objects. Sooner or later, you will have to consider splitting them per Domain (Content, Location, ContentType...). What about using custom implementations of the Repository's services, when applicable, instead ?

src/lib/UI/Dataset/RelationsDataset.php
```diff
@@ -35,23 +36,28 @@ public function __construct(ContentService $contentService, ValueFactory $valueF
     {
         $this->contentService = $contentService;
         $this->valueFactory = $valueFactory;
+        $this->relations = [];
```
mikadamczyk: This is not important, but if we have array as default value, I think it is more readable to set default value with properties declaration. `protected $relations = [];`

webhdx: I don't think we discussed it before. @Nattfarinn and @alongosz what is your take on it? I'm used to define default values in the constructor to keep properties section clean.

Nattfarinn: Both ways are fine if you ask me, and not sure if it brings any value here to "standarize" it as both are clean and readable. I would not create a constructor just to assign default value, but I don't mind assigning default value if we already have some property initialisations in a constructor.

webhdx:
@bdunogier I agree on refactoring Datasets to make them more like a value objects. They are definitely doing too much. I will discuss possible solutions with the team.

Factory will be refactored as well. We have the same problem with https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/lib/Form/Factory/FormFactory.php

We just need to negotiate sprint time for these things before final release.

PR #46, opened 2017-11-07
EZP-28190: ContentView should never be cached

Nattfarinn:
We can't (and we should not) cache ContentView in admin as it contains CSRF tokens.

Symfony Documentation ref: https://symfony.com/doc/current/http_cache/form_csrf_caching.html

### DIFF
--- a/src/bundle/Controller/ContentViewController.php
+++ b/src/bundle/Controller/ContentViewController.php
@@ -52,6 +52,8 @@ class ContentViewController extends Controller
 
     public function locationViewAction(ContentView $view)
     {
+        $view->setCacheEnabled(false);
+
         $this->supplyPathLocations($view);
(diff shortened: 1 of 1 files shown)


src/bundle/Controller/ContentViewController.php
```diff
@@ -52,6 +52,8 @@ public function __construct(
 
     public function locationViewAction(ContentView $view)
     {
+        $view->setCacheEnabled(false);
```
andrerom: should probably have had a comment here to explain why we disable it.

Nattfarinn: @andrerom A bit late but here it is: https://github.com/ezsystems/ezplatform-admin-ui/commit/aab8de44534b537f31ac8409b517e36ddc063656 (already on master)

PR #50, opened 2017-11-07
EZP-28163: 500 Internal Server Error on admin/trash/list

Nattfarinn:
**JIRA**: https://jira.ez.no/browse/EZP-28163

### DIFF
--- a/src/lib/Form/Type/Trash/TrashItemChoiceType.php
+++ b/src/lib/Form/Type/Trash/TrashItemChoiceType.php
@@ -6,9 +6,12 @@
  */
 namespace EzSystems\EzPlatformAdminUi\Form\Type\Trash;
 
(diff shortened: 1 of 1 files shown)


src/lib/Form/Type/Trash/TrashItemChoiceType.php
```diff
@@ -32,9 +47,14 @@ public function __construct(TrashService $trashService)
     public function configureOptions(OptionsResolver $resolver)
     {
         $resolver->setDefaults([
-            'choices' => $this->trashService->findTrashItems(new Query([
+            'choices' => array_map(function ($item) {
```
webhdx: * add type hinting form `$item` to reduce warnings in IDE
* extract `array_map` call to separate method for better code readability

PR #52, opened 2017-11-07
EZP-28193: Remove translation without selection causes internal sever error

Nattfarinn:
JIRA: https://jira.ez.no/browse/EZP-28193


micszo:
@Nattfarinn there is no error but also there is no information, user is just redirected to Content tab. Compare with behaviour on Locations tab when clicking Delete w/o selection.

Nattfarinn:
@micszo It's intentional, at least for now. Delete button should not be clickable at all in a first place, so this is just a fallback. It can be improved later.

PR #53, opened 2017-11-07
EZP-28195: Cannot open Roles list because of Internal Server Error

Nattfarinn:
JIRA: https://jira.ez.no/browse/EZP-28195

### DIFF
--- a/src/lib/Form/Factory/FormFactory.php
+++ b/src/lib/Form/Factory/FormFactory.php
@@ -524,10 +524,14 @@ class FormFactory
         RoleAssignmentDeleteData $data,
         ?string $name = null
     ): FormInterface {
-        $name = $name ?: sprintf('delete-role-assignment-%s', md5(implode('/', [
-            $data->getRoleAssignment()->getRole()->id,
-            $data->getRoleAssignment()->getRoleLimitation()->getIdentifier(),
-        ])));
+        $role = $data->getRoleAssignment()->getRole()->id;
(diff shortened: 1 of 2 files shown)


src/lib/Form/Factory/FormFactory.php
```diff
@@ -525,8 +525,9 @@ public function deleteRoleAssignment(
         ?string $name = null
     ): FormInterface {
         $name = $name ?: sprintf('delete-role-assignment-%s', md5(implode('/', [
-            $data->getRoleAssignment()->getRole()->id,
-            $data->getRoleAssignment()->getRoleLimitation()->getIdentifier(),
+            $data->getRoleAssignment()->getRole()->id, !empty($data->getRoleAssignment()->getRoleLimitation())
```
bdunogier: It's very personal, but I find that complicated to read when going through the code. A ternary method call inside an md5 of an implode of... :)

Nattfarinn: Because you're not so good in reading kung-fu!

Nattfarinn:
@bdunogier Fixed :)

src/lib/Form/Factory/FormFactory.php
bdunogier: You know how it goes: maintaining code is twice as hard than initially implementing it. If you're at 100% when writing code, you need to be at 200% when maintaining it.

Conclusion: keep it simple :)

PR #54, opened 2017-11-07
EZP-28150: In preview of Content type, field Container shows "yes_no"

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-28150

## Description 

I've regenerated translation by running in project root directory:  

```bash
bin/console translation:extract en --output-format=xliff --dir=vendor/ezsystems/ezplatform-admin-ui/src --output-dir=vendor/ezsystems/ezplatform-admin-ui/src/bundle/Resources/translations/
```
Also I've needed to manually set source and target to `{0}No|{1}Yes`, because `JMSTranslationBundle` wasn't able to do this automatically (not sure why).

### DIFF
--- a/src/bundle/Resources/translations/content.en.xliff
+++ b/src/bundle/Resources/translations/content.en.xliff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:jms="urn:jms:translation" version="1.2">
-  <file date="2017-11-06T16:28:43Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
+  <file date="2017-11-07T21:29:28Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
     <header>
       <tool tool-id="JMSTranslationBundle" tool-name="JMSTranslationBundle" tool-version="1.1.0-DEV"/>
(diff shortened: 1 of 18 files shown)


webhdx:
Question: What will happen when we run `translation:extract` again? Will `{0}No|{1}Yes` remain in tact?

adamwojs:
@webhdx @Nattfarinn Yes, you are right. I should use `translation:update` instead.

webhdx:
Nope, `translation:update` isn't an answer. As @Nattfarinn said you have to fix your code so that the extractor can extract translations itself.

barbaragr:
Retest ok, I'm waiting for a final version.

adamwojs:
I've investigated the behavior of `translation:extract` command and:

**`translation:extract` don't overwrite anything. It's adding translations for new keys, update references for existing ones and remove translations which are unused anymore.** 

The translation wasn't updated after adding `desc` filter in source file because the key has already existed in the translation file.  

Final conclusions: 

1. There is no bug in the `translation:extract` command. 
2. Execute `translation:extract` command doesn't touch existing translations (besides updating references)
3. If you are adding `desc` in the source file for a key that already has a translation, you need to update translation file manually, because of point 2.

lserwatka:
@Nattfarinn so are we good to merge this one, given explanation and research from @adamwojs ?

PR #58, opened 2017-11-08
EZP-27903: As an Editor, I want to edit a Time field type

sunpietro:
https://jira.ez.no/browse/EZP-27903

EZP-27903

**Requires:**
- https://github.com/ezsystems/ezplatform-admin-ui-assets/pull/1
- https://github.com/ezsystems/repository-forms/pull/169


sunpietro:
ping @barbaragr @dew326

sunpietro:
Before testing it, please run composer update to download the latest (v0.2.0) version of admin-ui-assets bundle. Otherwise it won't work.

src/bundle/Resources/views/content/content_edit.html.twig
```diff
@@ -13,9 +13,18 @@
...
         '@EzPlatformAdminUiBundle/Resources/public/js/scripts/fieldType/base/base-field.js'
         '@EzPlatformAdminUiBundle/Resources/public/js/scripts/fieldType/*'
     %}
         <script type="text/javascript" src="{{ asset_url }}"></script>
     {% endjavascripts %}
 {% endblock %}
+
+{% block stylesheets %}
+    {%  stylesheets
+        'bundles/ezplatformadminuiassets/vendors/flatpickr/dist/flatpickr.min.css'
```
dew326: `@EzPlatformAdminUiAssetsBundle/Resources/public/vendors/flatpickr/dist/flatpickr.min.css`? I never know which one we should use.

sunpietro: Corrected

webhdx:
@barbaragr @mikadamczyk Please hold on testing because the backend part isn't ready yet.

bdunogier:
For what it's worth, for those pull-requests that have complex requirements, I have sometimes used a temporary branch on the meta-repository. All of the requirements can be set there so that those who want to test don't need to check this or that in. Usually, I named it after the branch being tested (here `ezp-27903-time-field-type-design`).

barbaragr:
Here are some issues:

1. There is a difference between time in field Time and page title
Steps:
a. Create a new content type with field types: Time and Richtext for ex. In Time settings set up a default value: current time. Content type pattern should be filled with identifier of Time field type.
b. Create new content item with content type added in previous step. Current time will be loaded automatically.
c. Press "Publish" - there is a 1 hour difference between page title and time field. Furthermore there are two different formats of time. 
screen shot 2017-11-13 at 09 44 02

2. Even if default value is empty, current date appears.
a. Create a new content type with field types: Time and textline for ex. In Time settings set up a default value: empty. Content type pattern should be filled with identifier of Time field type.
b. Create new content item with content type added in previous step. Current time is loaded automatically.
c. Press "Publish" - there is an empty time field in page preview.
screen shot 2017-11-13 at 11 55 11

webhdx:
@barbaragr 1: Time difference happens due to the different timezones used by your browser and server. Most likely your servers is configured to use UTC timezone thus -1 hour difference. We cannot do anything about it for the time being.
Different time format is something I also noticed but I'm unsure why does it happen. Especially because we are using kernel for rendering. 

@alongosz Do you have any idea how to fix different time format issue?

alongosz:
@webhdx content name (which is displayed at the top in breadcrumbs) comes from `FieldType::getName` and [is hardcoded for `eztime`](https://github.com/ezsystems/ezpublish-kernel/blob/v6.7.5/eZ/Publish/Core/FieldType/Time/Type.php#L62-L71). It probably should use `Time::$stringFormat`. Please report a bug and create PR.

As it goes for the issue with empty value in form / edit view - empty `eztime` is equal to `null`. Instance of `new \DateTime(null)` corresponds to the current time.

sunpietro:
Everything is done. 
@webhdx There's still an issue with storing default _current time_ as the field value and with restoring a correct time after getting back to content edit.

src/bundle/Resources/public/js/scripts/fieldType/eztime.js
```diff
@@ -0,0 +1,88 @@
...
+        const enableSeconds = sourceInput.dataset.seconds === '1';
+        let defaultDate;
+
+        sourceInput.classList.add('ez-data-source__input--visually-hidden');
+
+        if (sourceInput.value) {
+            const date = new Date(sourceInput * 1000);
+
+            if (isNaN(date.getTime())) {
+                validator.validateInput({target: sourceInput});
```
dew326: When the date is `NaN`?
This validation will not display any error message and I don't know why it's done here.
If you want to validate input probably you need to fire an `valueChanged` event.

barbaragr:
I've created a ticket with 1-hour difference: https://jira.ez.no/browse/EZP-28244

PR #64, opened 2017-11-13
EZP-27974: Default sort parameter for sub-items table in the Detail content view tab

mikadamczyk:
> https://jira.ez.no/browse/EZP-27974

For a specific location, the editor can define the sort order of the sub-items. 

screen shot 2017-11-14 at 11 38 15 am

### DIFF
--- a/src/bundle/Controller/ContentViewController.php
+++ b/src/bundle/Controller/ContentViewController.php
@@ -92,15 +92,19 @@ class ContentViewController extends Controller
         $locationCopyType = $this->formFactory->copyLocation(
             new LocationCopyData($location)
         );
+
         $locationMoveType = $this->formFactory->moveLocation(
             new LocationMoveData($location)
         );
+
         $locationTrashType = $this->formFactory->trashLocation(
             new LocationTrashData($location)
         );
+
         $contentDraftCreateType = $this->formFactory->createContentDraft(
             new ContentDraftCreateData($content->contentInfo, $versionInfo)
         );
+
         $contentCreateType = $this->formFactory->createContent(
             $this->getContentCreateData($location)
         );
--- a/src/bundle/Controller/LocationController.php
+++ b/src/bundle/Controller/LocationController.php
@@ -11,6 +11,7 @@ use eZ\Publish\API\Repository\ContentTypeService;
 use eZ\Publish\API\Repository\LocationService;
 use eZ\Publish\API\Repository\TrashService;
 use eZ\Publish\API\Repository\Values\Content\ContentInfo;
+use eZ\Publish\API\Repository\Values\Content\LocationUpdateStruct;
 use eZ\Publish\Core\Base\Exceptions\InvalidArgumentException;
 use EzSystems\EzPlatformAdminUi\Form\Data\Content\Location\ContentLocationAddData;
 use EzSystems\EzPlatformAdminUi\Form\Data\Content\Location\ContentLocationRemoveData;
@@ -18,6 +19,7 @@ use EzSystems\EzPlatformAdminUi\Form\Data\Location\LocationCopyData;
 use EzSystems\EzPlatformAdminUi\Form\Data\Location\LocationMoveData;
 use EzSystems\EzPlatformAdminUi\Form\Data\Location\LocationSwapData;
 use EzSystems\EzPlatformAdminUi\Form\Data\Location\LocationTrashData;
+use EzSystems\EzPlatformAdminUi\Form\Data\Location\LocationUpdateData;
 use EzSystems\EzPlatformAdminUi\Form\Factory\FormFactory;
 use EzSystems\EzPlatformAdminUi\Form\SubmitHandler;
 use EzSystems\EzPlatformAdminUi\Notification\NotificationHandlerInterface;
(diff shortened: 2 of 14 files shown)


src/bundle/Controller/LocationController.php
```diff
@@ -395,4 +397,51 @@ public function addAction(Request $request): Response
...
+                        /** @Desc("Location '%name%' updated.") */
+                        'location.update.success',
+                        ['%name%' => $location->getContentInfo()->name],
+                        'location'
+                    )
+                );
+
+                return new RedirectResponse($this->generateUrl('_ezpublishLocation', [
+                        'locationId' => $location->getContentInfo()->mainLocationId,
+                    ]) . '#ez-tab-location-view-details');
```
adamwojs: I think you can use `_fragment` parameter instead of string concatenation: https://symfony.com/blog/new-in-symfony-3-2-routing-improvements

src/bundle/Controller/LocationController.php
```diff
@@ -395,4 +397,51 @@ public function addAction(Request $request): Response
...
+     * Handles update existing location.
+     *
+     * @param Request $request
+     *
+     * @return Response
+     */
+    public function updateAction(Request $request): Response
+    {
+        $form = $this->formFactory->updateLocation(
+            new LocationUpdateData()
```
webhdx: You don't need to initialize `LocationUpdateData()` here.

src/bundle/Controller/LocationController.php
```diff
@@ -395,4 +397,51 @@ public function addAction(Request $request): Response
...
+        );
+        $form->handleRequest($request);
+
+        $location = $form->getData()->getLocation();
+
+        if ($form->isSubmitted()) {
```
webhdx: Should be `$form->isSubmitted() && $form->isValid()`

src/bundle/Controller/LocationController.php
webhdx: Use magic parameter `_fragment` as @adamwojs said

src/bundle/Resources/translations/search.en.xliff
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:jms="urn:jms:translation" version="1.2">
-  <file date="2017-11-07T21:29:28Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
+  <file date="2017-11-09T13:48:46Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
```
webhdx: Can we clean up this PR from unrelated translation files? I think during extraction you can limit domains.

src/bundle/Controller/ContentViewController.php
```diff
@@ -92,15 +93,23 @@ private function supplyContentActionForms(ContentView $view): void
...
+
         $locationMoveType = $this->formFactory->moveLocation(
             new LocationMoveData($location)
         );
+
         $locationTrashType = $this->formFactory->trashLocation(
             new LocationTrashData($location)
         );
+
+        $locationUpdateType = $this->formFactory->updateLocation(
```
webhdx: This form should be a part of `DetailsTab` as far as I understand. Doing this way you have to pass your form object to all location view tabs.

src/lib/Form/Type/ContentType/SortFieldChoiceType.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $this->translator = $translator;
+    }
+
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver->setDefaults([
+            'choices' => $this->getSortFieldChoices(),
+            'choices_as_values' => true,
+            'translation_domain' => 'ezrepoforms_content_type',
```
webhdx: `ezrepoforms` domain name is reserved for `ezsystems/repository-forms`. You can use `forms` domain name.

src/lib/Form/Type/ContentType/SortFieldChoiceType.php
```diff
@@ -0,0 +1,90 @@
...
+     *
+     * @param string $sortField
+     *
+     * @return string
+     *
+     * @throws InvalidArgumentException
+     */
+    private function getSortFieldLabel(string $sortField): string
+    {
+        return $this->translator->trans(/** @Ignore */'content_type.sort_field.' . $sortField, [], 'ezrepoforms_content_type');
```
webhdx: Doesn't it make labels unextractable by `bin/console translation:extract`? In this case you should provide an extractor. You can do this in similar fashion as I did with menu builders: https://github.com/ezsystems/ezplatform-admin-ui/blob/a90408f631e0d982db48fa2b71b9639af2428246/src/lib/Menu/MainMenuBuilder.php#L145

adamwojs: AFAIK this is reuse of translations that are already defined in repository-forms and they have custom extractor: https://github.com/ezsystems/repository-forms/blob/master/lib/Translation/SortingTranslationExtractor.php

mikadamczyk: Yes, this is

src/bundle/Resources/translations/search.en.xliff
mikadamczyk: Yes, I will try

lserwatka:
@mikadamczyk could you resolve conflicts here

PR #67, opened 2017-11-14
EZP-28094: Set the main location of a content item

mikadamczyk:
> https://jira.ez.no/browse/EZP-27974

The main location will be displayed at the top of the list and rest of the locations underneath the main.

set-main-location

### DIFF
--- a/src/bundle/Controller/ContentController.php
+++ b/src/bundle/Controller/ContentController.php
@@ -9,17 +9,22 @@ namespace EzSystems\EzPlatformAdminUiBundle\Controller;
 use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\Exceptions as ApiException;
 use eZ\Publish\API\Repository\Exceptions\UnauthorizedException;
-use eZ\Publish\API\Repository\LocationService;
+use EzSystems\EzPlatformAdminUi\Exception\InvalidArgumentException as AdminInvalidArgumentException;
 use EzSystems\EzPlatformAdminUi\Form\Data\Content\Draft\ContentCreateData;
 use EzSystems\EzPlatformAdminUi\Form\Data\Content\Draft\ContentDraftCreateData;
+use EzSystems\EzPlatformAdminUi\Form\Data\Content\Location\ContentMainLocationUpdateData;
+use EzSystems\EzPlatformAdminUi\Form\DataMapper\ContentMainLocationUpdateMapper;
 use EzSystems\EzPlatformAdminUi\Form\Factory\FormFactory;
 use EzSystems\EzPlatformAdminUi\Form\SubmitHandler;
 use EzSystems\EzPlatformAdminUi\Notification\NotificationHandlerInterface;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\OptionsResolver\Exception\InvalidOptionsException;
 use Symfony\Component\Translation\Exception\InvalidArgumentException;
 use Symfony\Component\Translation\TranslatorInterface;
+use eZ\Publish\API\Repository\Exceptions\InvalidArgumentException as APIRepositoryInvalidArgumentException;
+use Symfony\Component\Translation\Exception\InvalidArgumentException as TranslationInvalidArgumentException;
 
 class ContentController extends Controller
 {
@@ -28,9 +33,6 @@ class ContentController extends Controller
 
     /** @var ContentService */
     private $contentService;
-
-    /** @var LocationService */
-    private $locationService;
 
     /** @var FormFactory */
     private $formFactory;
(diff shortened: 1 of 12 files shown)


src/bundle/Controller/ContentController.php
```diff
@@ -157,4 +156,62 @@ public function createDraftAction(Request $request): Response
...
+     */
+    public function setMainLocationAction(Request $request): Response
+    {
+        $form = $this->formFactory->updateContentMainLocation();
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            $result = $this->submitHandler->handle($form, function (ContentUpdateMainLocationData $data) {
+                $contentInfo = $data->getContentInfo();
+                $contentMetadataUpdateStruct = new ContentMetadataUpdateStruct();
```
webhdx: As far as I remember we use mappers to do the conversion between repository structs and our DTOs? @Nattfarinn am I right?

mikadamczyk: I will check this and fix if needed

src/lib/Form/Data/Content/ContentUpdateMainLocationData.php
```diff
@@ -0,0 +1,79 @@
...
+namespace EzSystems\EzPlatformAdminUi\Form\Data\Content;
+
+use eZ\Publish\API\Repository\Values\Content\Location;
+use eZ\Publish\API\Repository\Values\Content\ContentInfo;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * @todo Add validation.
+ */
+class ContentUpdateMainLocationData
```
webhdx: Name doesn't follow our notation, correct name should be `ContentMainLocationUpdateData`.

src/bundle/Resources/config/routing.yml
```diff
@@ -342,3 +342,9 @@ ezplatform.search:
     methods: ['GET']
     defaults:
         _controller: 'EzPlatformAdminUiBundle:Search:search'
+
+ezplatform.content.set_main_location:
```
webhdx: Declaration is alright but placement in the file not so much. As you can see above in the same file I've tried to keep it organized using `# Comments` to separate sections. Can you place your route declaration in more appropriate section? Might be at the end of `# LocationView / Locations tab` section.

src/lib/Form/Factory/FormFactory.php
```diff
@@ -121,6 +123,28 @@ public function createContentDraft(
...
 
     /**
+     * @param ContentUpdateMainLocationData|null $data
+     * @param string|null $name
+     *
+     * @return FormInterface
+     *
+     * @throws InvalidOptionsException
+     */
+    public function updateContentMainLocation(
```
webhdx: Code placement isn't logical to me, there are methods like `removeLocation()`, `swapLocation()`, somewhere around that area seems like a better fit.

src/lib/Form/Type/Content/Location/ContentLocationSetMainType.php
```diff
@@ -0,0 +1,67 @@
...
+use EzSystems\EzPlatformAdminUi\Form\Type\Content\ContentInfoType;
+use EzSystems\EzPlatformAdminUi\Form\Type\Content\LocationType;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\Extension\Core\Type\SubmitType;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use eZ\Publish\API\Repository\LocationService;
+use eZ\Publish\API\Repository\Values\Content\Location;
+
+class ContentLocationSetMainType extends AbstractType
```
webhdx: We do not have any formal guidelines for Type/Data class naming but to keep things streamlined:

* Corresponding data class is named `ContentUpdateMainLocationData` or as proposed by me: `ContentMainLocationUpdateData`. The type should be named after the data class: `ContentMainLocationUpdateType`.
* With above change you should also change submit button name as it should be named after action name (which now is `set` but should be `update`).

src/lib/Tab/LocationView/LocationsTab.php
```diff
@@ -91,12 +92,14 @@ public function renderView(array $parameters): string
...
         $formLocationRemove = $this->createLocationRemoveForm($location, $locations);
         $formLocationSwap = $this->createLocationSwapForm($location);
+        $formLocationSetMain = $this->createLocationSetMainForm($contentInfo, $location);
 
         $viewParameters = [
             'locations' => $locations,
             'form_content_location_add' => $formLocationAdd->createView(),
             'form_content_location_remove' => $formLocationRemove->createView(),
             'form_content_location_swap' => $formLocationSwap->createView(),
+            'form_content_location_set_main' => $formLocationSetMain->createView(),
```
webhdx: If you decided to streamline names to use `update` instead of `set` action name, please also update all occurrences like the ones above: array key and variable name.

src/bundle/Controller/ContentController.php
Nattfarinn: Mapper is overkill and will be misleading. Revert it.

PR #70, opened 2017-11-15
EZP-28229: As an Editor, I want to have styled login form

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-28229

**Screenshot:**
login-form


src/bundle/Resources/views/Security/login.html.twig
```diff
@@ -2,33 +2,37 @@
...
-                            <button type="submit" class="btn btn-primary">{{ 'authentication.submit'|trans|desc('Login') }}</button>
-                        </fieldset>
-                    </form>
-                </div>
+                        <button type="submit" class="btn btn-primary ez-btn ez-btn--login">{{ 'authentication.submit'|trans|desc('Login') }}</button>
+                    </fieldset>
+                </form>
+            </div>
+            <div class="ez-login__footer">
+                <span>eZ Platform ©1999-2017 <a href="http://ez.no">eZ Systems AS</a> and others.</span>
```
adamwojs: Can we use `{{ "now"|date("Y") }}` instead of hardcoding current year ? 😉

lserwatka: Yes, this is good idea!

dew326: I didn't know how to do it in PHP. I will add it ;)

dew326:
@adamwojs @sunpietro done

src/bundle/Resources/public/scss/_login.scss
```diff
@@ -0,0 +1,34 @@
...
+        color: $ez-white;
+
+        &-wrapper {
+            width: 18rem;
+            height: 90%;
+            display: flex;
+            flex-direction: column;
+            justify-content: center;
+
+            .ez-login__form-label {
```
sunpietro: Does it have to be nested like this? I think can be placed one level up without any issues :)

src/bundle/Resources/public/scss/_login.scss
```diff
@@ -0,0 +1,34 @@
...
+                margin-bottom: 0;
+                margin-top: 1rem;
+            }
+
+            .ez-login__form-input {
```
sunpietro: Same here.

m-tyrala:
QA test ok.

PR #78, opened 2017-11-17
EZP-28244: Difference between time/date in edit and preview mode

sunpietro:
https://jira.ez.no/browse/EZP-28244

There's still a know issue with the eztime fieldtype. It's a backend issue.

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
@@ -53,7 +53,10 @@
         formatDate: (date) => (new Date(date)).toLocaleDateString()
     };
     const updateInputValue = (sourceInput, date) => {
-        sourceInput.value = Math.floor((new Date(date)).getTime() / 1000);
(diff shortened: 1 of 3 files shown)


barbaragr:
For DateTime field date issue works fine.

For Date:
a. Create a new content type with field type: Date and In Date settings set up a default value: current date. 
b. Create new content item with content type added in previous step. Current date will be loaded automatically and it is one day earlier than the current date.

In addition: when I am in create mode date format is mm/dd/yyyy (both for Date and DateTime), but after publishing date format is dd/mm/yyyy.

webhdx:
My observation: There is something wrong in `Time\Value::fromTimestamp()` implementation as it messes up timezones. Switching to the same implementation as in `DateAndTime\Value::fromTimestamp()` will fix the issue. I need to spend more time on it to come up with final verdict and hopefully a fix in kernel repository.

lserwatka:
We need to fix issue with eztime field type. Ref. to https://jira.ez.no/browse/EZP-28269

PR #79, opened 2017-11-17
EZP-27974: Fix for deleted `locationId` in listener.

mikadamczyk:
Fix for deleted `locationId` in listener. There is still a bug with route generated in breadcrumb.

### DIFF
--- a/src/bundle/EventListener/RequestAttributesListener.php
+++ b/src/bundle/EventListener/RequestAttributesListener.php
@@ -19,7 +19,7 @@ use eZ\Publish\API\Repository\Repository;
  */
 class RequestAttributesListener implements EventSubscriberInterface
 {
-    /** @var \eZ\Publish\API\Repository\Repository */
+    /** @var Repository */
     private $repository;
 
     /** @var array */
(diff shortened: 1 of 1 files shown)


src/bundle/EventListener/RequestAttributesListener.php
```diff
@@ -43,20 +43,21 @@ public static function getSubscribedEvents()
     /**
      * Adds all the request attributes to the parameters.
      *
-     * @param \eZ\Publish\Core\MVC\Symfony\View\Event\FilterViewBuilderParametersEvent $e
+     * @param \eZ\Publish\Core\MVC\Symfony\View\Event\FilterViewBuilderParametersEvent $event
```
webhdx: We tend to not use FQCN in docblocks, can you import this class?

src/bundle/EventListener/RequestAttributesListener.php
```diff
@@ -43,20 +43,21 @@ public static function getSubscribedEvents()
...
+
+        if (!$this->isAdmin($request)) {
             return;
         }
 
-        $parameterBag = $e->getParameters();
+        $parameterBag = $event->getParameters();
 
-        if ($parameterBag->has('locationId')) {
+        if ($parameterBag->has('locationId') && $request->get('_route') === '_ezpublishLocation') {
```
adamwojs: IMHO `$request->get('_route') === '_ezpublishLocation'`  is precondition to manipulate `$parameterBag` so if you move it into the first "if" statement, the code will be more readable 😉

mikadamczyk: Yes, we discussed this with @webhdx. The only reason for this is that first check is lightweight

PR #83, opened 2017-11-21
EZP-28289: Updated UI colors in Content views

inakijv:
Most important aspects:
Updated UI colors and UI colors of Content Views (next step will target Editing views);

Work In Progress due to the following blockers:

- Conflict updating `ez-tabs` when in _admin/systeminfo_ view, due to it has a different block structure.
- Conflict when content is not displayed within tables. Specifically between [1] `#ez-tab-location-view-relations section.ez-table-header` & `#ez-tab-location-view-relations section.ez-relations-box-list-no-content` and [2] `locations:panel_swap.html.twig location_swap`. Both should have same colors and styling.
- Conflict in dashboard. Background-color not updated yet. It seems that there's no content added to `.ez-header` ("My dashboard", header according to the design). 

ez platform


inakijv:
ping @dew326 @sunpietro @SylvainGuittard

lserwatka:
Hi @inakijv , could you create JIRA ticket and amend commit message before we commit your PR?

src/bundle/Resources/public/scss/_custom.scss
```diff
@@ -3,31 +3,38 @@ $ez-admin-transition: cubic-bezier(.25, .8, .25, 1);
...
+$ez-ground-base: #f3f3f3;
+$ez-ground-base-medium: #ededed;
+$ez-ground-base-dark: #e5e3e3;
+$ez-color-base-pale: #dbdbdb;
+$ez-color-base-light: #878787;
+$ez-color-base-medium: #646464;
+$ez-color-base-dark: #555;
+$ez-black: #333;
 
+$ez-secondary-ground-pale: #e1f5ff;
```
SylvainGuittard: It makes more sense to group `ez-ground-primary*` and `ez-secondary-ground*`
And also better to be consistent on naming: `ground-primary` or `primary-ground` / `ground-secondary` or `secondary-ground`. :)

inakijv: We are basically arranging colors in three groups: ez colors (branding) and grays & blues palettes (for background composition). Your comment would be related to the third group. I reused existing variables and when needing a new one I have followed the current labeling logic.

+1 to consistency on naming. We should tackle this too.

inakijv:
@lserwatka I have applied changes requested.

src/bundle/Resources/public/scss/_custom.scss
inakijv: Ping @dew326 & @sunpietro What do you think?

src/bundle/Resources/public/scss/_breadcrumb.scss
```diff
@@ -0,0 +1,16 @@
...
+.breadcrumb-item {
+    cursor: not-allowed;
+    
+    + .breadcrumb-item::before {
+        color: $ez-color-base-medium;
+    }
+
+    &.active {
+        color: $ez-color-base-medium;
+        cursor: not-allowed;
```
sunpietro: You're duplicating the `cursor: not-allowed`. Is it intentional?

dew326: Why we need `cursor: not-allowed` on `.active`? The same rule is without `.active`.

src/bundle/Resources/public/scss/_custom.scss
```diff
@@ -3,31 +3,38 @@ $ez-admin-transition: cubic-bezier(.25, .8, .25, 1);
...
 /* eZ colors */
 $ez-color-primary: #f15a10;
-$ez-color-primary-hover: #cf4c0c;
-$ez-color-secondary: #0f6d95;
-$ez-color-secondary-hover: #0c5472;
+$ez-color-primary-hover: darken($ez-color-primary, 15%);
+$ez-color-secondary: #106d95;
+$ez-color-secondary-hover: darken($ez-color-secondary, 15%);
 $ez-color-hyperlink: #0645ad;
 $ez-color-hyperlink-hover: #032863;
```
dew326: I really like to use the `darken`. Can we use it for hyperlinks as well?

sunpietro:
@inakijv please resolve conflicts here

inakijv:
@sunpietro comparaison table as requested :smile:
ui_colors_comparaison_table

inakijv:
Thanks @sunpietro & @dew326 for the review. I have applied your suggestions 👍

lserwatka:
@inakijv we can merge this PR but you must resolve conflicts here

src/bundle/Resources/public/scss/_custom.scss
dew326: The naming here is a little bit misleading. I would expect that `ground-primary` is orange and `ground-secondary` is blue but it seems all to be blues. I think we should use one common name and add some suffix.

src/bundle/Resources/public/scss/_breadcrumb.scss
```diff
@@ -0,0 +1,15 @@
...
+}
+
+.breadcrumb-item {
+    cursor: not-allowed;
+    
+    + .breadcrumb-item::before {
+        color: $ez-color-base-medium;
+    }
+
+    &.active {
```
sunpietro: This selector can be assigned to the rule above, so:
```
+ .breadcrumb-item::before,
&.active {
```

src/bundle/Resources/public/scss/_tabs.scss
```diff
@@ -1,24 +1,25 @@
...
 
     .nav-link {
-        color: $ez-color-base-pale;
+        color: $ez-color-base-medium;
 
         &:hover {
-            background-color: $ez-ground-base-dark;
-            border-color: $ez-ground-base-dark;
+            background-color: $ez-color-base-pale;
+            border-color: $ez-color-base-pale;
```
sunpietro: What's the point in setting the border color the same as the background color?
From what I can see in this PR: `.nav-tabs` and `.nav-link` items share the same colors for border and background. We might skip the border CSS properties, do you agree?

inakijv: Because when removed `background-color` and `border-color` are different.

inakijv:
Conflicts solved and changes applied. Ping @dew326 @sunpietro @SylvainGuittard

src/bundle/Resources/public/scss/_custom.scss
lserwatka: @inakijv could you explain why all colors are blue variants?

inakijv: @lserwatka for the composition of the background, we are using both gray and blue hues. The former group is used for the background composition. The latter adds the soft accent to that background (as well as the some of the interactions) in order to give all the stage for the brand (and more saturated) colors that we selected.

The screenshot below with most of the colors listed out (based on `_custom.scss` file can help to explain this.
ui_colors_update

PR #85, opened 2017-11-21
EZP-28284: Do not run admin siteaccess filter in there are no groups defined

emodric:
It appears that this filter is ran for every file which has `ezpublish` configuration defined.

So, if there are any files with `ezpublish` configuration, but without siteaccess groups config, this causes an exception.


lserwatka:
Hi @emodric, could you create JIRA ticket and amend commit message before we commit your PR?

emodric:
Done!

PR #88, opened 2017-11-21
EZP-28273: Make v2 admin sidebar buttons sticky

janit:
https://jira.ez.no/browse/EZP-28273

In the administration interface the views can be very long. This causes the user to scroll down and then the sidebar buttons are lost. This is especially prominent when editing content with multiple field.

This PR adds a wrapper element and some styles to set the buttons in the sidebar to be sticky when scrolling. See video for a preview - sticky functionality in video has white border.

https://www.youtube.com/watch?v=GZoD7GJ99Y4

### DIFF
--- a/src/bundle/Resources/public/scss/_general.scss
+++ b/src/bundle/Resources/public/scss/_general.scss
@@ -32,6 +32,10 @@ a {
     padding-right: 0;
     padding-left: 0;
 
+    .ez-button-container {
+        position: fixed;
+    }
+
     .ez-icon {
         display: block;
         margin: 0 auto;
--- a/src/bundle/Resources/public/scss/_general.scss
+++ b/src/bundle/Resources/public/scss/_general.scss
@@ -32,11 +32,6 @@ a {
     padding-right: 0;
     padding-left: 0;
 
-    .ez-button-container {
-        position: sticky;
-        top: 1em;
-    }
-
     .ez-icon {
         display: block;
         margin: 0 auto;
@@ -126,4 +121,9 @@ a {
 .ez-section-list-view,
 .ez-trash-list-view {
     background-color: $ez-ground-base-light;
+}
+
+.ez-sticky-container {
+    position: sticky;
+    top: 1rem;
 }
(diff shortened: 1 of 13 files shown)


src/bundle/Resources/public/scss/_general.scss
```diff
@@ -25,6 +25,9 @@ body {
 
 .ez-side-menu,
 .ez-context-menu {
+    .ez-button-container {
+        position: fixed;
```
sunpietro: Are you sure it's cross-browser safe? We're supporting the latest browser versions.
Also, this change will affect the right side menu in the location view, not only in the edit view. Is it desired?
What if there are many buttons? What's your experience with a such case.
Also, to improve browser performance you might want to add a property `will-change: transform`.

janit: I actually switched to `position:sticky`, which won't lock the container to the initial position, rather it will stick to a fixed position from top only when scrolling. I also added your suggestion for `will-change`. As for support for this feature, it is in newest browser versions and I don't think we need a browser polyfill since this is only an enhancement.

While I looked at this briefly today, I noticed there is actually an issue with the content type selector, etc. widgets (see [video](https://youtu.be/mFSTTP6DqkA)), which will also need to be handled for this to be an improvement over current functionality.

lserwatka:
@janit could you amend commit message with JIRA ticket?

janit:
Ok, I added the link to the PR description

janit:
I looked at this a bit today, but couldn't figure out a smart way to make the extra tools sticky. 

The `.ez-extra-actions-container` seems to be positioned an collapsed underneath the buttons, see image underneath. Do you have any ideas how we could make it sticky and follow?

screen shot 2017-11-28 at 17 06 24

dew326:
Branch rebased with master and I fixed the solution.

The sticky container is added for sidebars in:
- location view
- content edit/create
- trash
- content type edit/create
- content type group edit/create
- language edit/create
- policy edit/create
- role edit/create
- role assignment 
- sections edit/create

Known limitation: Sticky container cannot contain bootstrap modal. The modal will be unclickable because it's under the backdrop.

janit:
Thank you for the improvements @dew326! I tried it on my install and it works just as expected now, so from my part this is good to go.

PR #89, opened 2017-11-22
EZP-28227: As a v2 editor I want to preview content in Edit mode

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-28227

**TODO**
- [x] create preview button in sidebar in content edit
- [x] create a proper route for preview content
- [x] add href to back button
- [x] add proper src to iframe


src/bundle/Resources/public/scss/_icons.scss
```diff
@@ -95,3 +95,16 @@
         }
     }
 }
+
+.ez-preview__action {
+    .ez-icon {
+        width: 1rem;
```
sunpietro: I think it's a good time to introduce icons mixins like: `icon-small`, `icon-medium`, `icon-big` that will define the icon size in a specific context. Or maybe, we predefine icons CSS classes instead.

src/bundle/Resources/public/scss/_preview.scss
```diff
@@ -0,0 +1,87 @@
...
+                line-height: 0;
+            }
+
+            &--actions {
+                flex-basis: auto;
+                display: flex;
+                justify-content: center;
+                border: 1px solid $ez-color-base-medium;
+                border-radius: 5px;
+                padding: .1rem;
```
sunpietro: `.1rem` is `0.0625px` if font size is set to `16px`. Maybe it should be `0.125rem` instead, which translates to `1px`? Some browsers can do subpixel rendering.

src/bundle/Resources/public/scss/_preview.scss
```diff
@@ -0,0 +1,87 @@
...
+            &-text {
+                text-overflow: ellipsis;
+                width: 80%;
+                display: block;
+                overflow: hidden;
+                white-space: nowrap;
+            }
+
+            .ez-preview__action {
+                line-height: 0;
```
sunpietro: Why is it se to `line-height: 0`?

src/bundle/Resources/public/scss/_preview.scss
```diff
@@ -0,0 +1,87 @@
...
+                display: block;
+                padding: .5rem;
+                border: none;
+                border-radius: initial;
+                cursor: pointer;
+                border-radius: 5px;
```
sunpietro: Duplicated `border-radius` property.

PR #93, opened 2017-11-22
EZP-27900: As an Editor, I want to edit a Country field type

sunpietro:
https://jira.ez.no/browse/EZP-27900

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezcountry.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezcountry.js
@@ -1,0 +1,45 @@
+(function (global) {
+    const SELECTOR_FIELD = '.ez-field-edit--ezcountry';
+
+    class EzCountryValidator extends global.eZ.BaseFieldValidator {
+        /**
+         * Validates the select box value
+         *
+         * @method validateSelection
+         * @param {Event} event
+         * @returns {Object}
+         * @memberof EzCountryValidator
+         */
+        validateSelection(event) {
+            const isRequired = event.target.required;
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/scripts/fieldType/ezcountry.js
```diff
@@ -0,0 +1,45 @@
...
+         * @memberof EzCountryValidator
+         */
+        validateSelection(event) {
+            const isRequired = event.target.required;
+            const isEmpty = !event.target.value.trim().length;
+            const isError = (isEmpty && isRequired);
+            const label = event.target.closest(SELECTOR_FIELD).querySelector('.ez-field-edit__label').innerHTML;
+            const result = { isError };
+
+            if (isEmpty) {
```
dew326: I think it should be `isError`

sunpietro: I disagree. I'll keep that like it is now. By checking if a field value is empty I'm able to attach a correct error message.

sunpietro:
@m-tyrala deafult empty value added

m-tyrala:
Now, when there's possibility to select empty option, i see that theres no lose-focus validation too.

src/bundle/Resources/public/js/scripts/fieldType/ezcountry.js
dew326: If we want to be precise with adding correct error message then we should use `isEmpty && isError` because when the field is not required the value can be empty.

sunpietro:
@m-tyrala please, retest it

PR #94, opened 2017-11-22
EZP-28344: Twig template extensibility via Components (Dashboard, CSS & JS)

Nattfarinn:
- Extensibility: https://github.com/ezsystems/ezplatform-admin-ui/pull/94/commits/5e0f2071c18d8d90ebecca8928a09d3c81c24ad4
- Dashboard: https://github.com/ezsystems/ezplatform-admin-ui/pull/94/commits/2201b1333d98b218518d33ad2380667d33e9f06f
- CSS & JS: https://github.com/ezsystems/ezplatform-admin-ui/pull/94/commits/818a5bae8403b6f30bcdafaa576b3ad50a2fbd02

### DIFF
--- a/src/bundle/Templating/Twig/ComponentExtension.php
+++ b/src/bundle/Templating/Twig/ComponentExtension.php
@@ -1,0 +1,66 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformAdminUiBundle\Templating\Twig;
+
+use EzSystems\EzPlatformAdminUi\Component\Registry as ComponentRegistry;
+use EzSystems\EzPlatformAdminUi\Component\Renderable;
+use EzSystems\EzPlatformAdminUi\Exception\InvalidArgumentException;
+use Twig_Extension;
+use Twig_SimpleFunction;
+
+class ComponentExtension extends Twig_Extension
+{
+    protected $registry;
+
+    public function __construct(
+        ComponentRegistry $registry
+    ) {
+        $this->registry = $registry;
+    }
+
+    public function getFunctions()
+    {
+        return [
+            new Twig_SimpleFunction(
+                'ezplatform_admin_ui_component_group',
+                [$this, 'renderComponentGroup'],
+                ['is_safe' => ['html']]
+            ),
+            new Twig_SimpleFunction(
+                'ezplatform_admin_ui_component',
+                [$this, 'renderComponent'],
+                ['is_safe' => ['html']]
+            ),
+        ];
+    }
+
+    public function renderComponentGroup(string $group, array $parameters = [])
+    {
+        $components = $this->registry->getComponents($group);
+        $outputs = array_map(function (Renderable $component) use ($parameters) {
+            return $component->render($parameters);
+        }, $components);
+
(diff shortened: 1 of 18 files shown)


src/lib/Component/Registry.php
```diff
@@ -0,0 +1,24 @@
...
+class Registry
+{
+    protected $components = [];
+
+    public function addComponent(string $group, string $serviceId, Renderable $component): void
+    {
+        $this->components[$group][$serviceId] = $component;
+    }
+
+    public function getComponents($group): array
```
webhdx: Can you add a docblock with proper return type? IDE users will benefit from that.

src/lib/Component/Registry.php
```diff
@@ -0,0 +1,24 @@
...
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformAdminUi\Component;
+
+class Registry
+{
+    protected $components = [];
```
webhdx: A docblock would be advised.

src/bundle/Templating/Twig/ComponentExtension.php
```diff
@@ -0,0 +1,66 @@
...
+
+    public function renderComponent(string $group, string $id, array $parameters = [])
+    {
+        $components = $this->registry->getComponents($group);
+
+        if (!isset($components[$id])) {
+            throw new InvalidArgumentException('id', sprintf("Can't find Component '%s' in group '%s'", $id, $group));
+        }
+
+        /** @var Renderable $component */
```
webhdx: If you added proper docblock in Registry class you wouldn't have to make `@var` comments here ;)

src/bundle/Templating/Twig/ComponentExtension.php
```diff
@@ -0,0 +1,66 @@
...
+        ComponentRegistry $registry
+    ) {
+        $this->registry = $registry;
+    }
+
+    public function getFunctions()
+    {
+        return [
+            new Twig_SimpleFunction(
+                'ezplatform_adminui_component_group',
```
webhdx: I believe it would be more streamlined with other parts of Admin UI if the name was `ezplatform_admin_ui_component_group`

src/bundle/Templating/Twig/ComponentExtension.php
```diff
@@ -0,0 +1,66 @@
...
+
+use EzSystems\EzPlatformAdminUi\Component\Registry as ComponentRegistry;
+use EzSystems\EzPlatformAdminUi\Component\Renderable;
+use EzSystems\EzPlatformAdminUi\Exception\InvalidArgumentException;
+use Twig_Extension;
+use Twig_SimpleFunction;
+
+class ComponentExtension extends Twig_Extension
+{
+    protected $registry;
```
webhdx: Missing docblock

lserwatka:
@Nattfarinn can you add event dispatching as @webhdx suggested?

Nattfarinn:
Isn't it a feature we can add later? We don't need it yet.

lserwatka:
@Nattfarinn if you are on it and it is not much work, I would add it now, as there will be always something more important to do in the future.

lserwatka:
We also need possibility to define ordering for Dashboard blocks.

lserwatka:
@Nattfarinn @webhdx what is the status here?

PR #96, opened 2017-11-22
EZP-27999: Updated typography in content and edit views

inakijv:
Most important aspects:
- Reviewed and updated typography, `headings`;
- Reviewed typography in Content and Edit views.

Next updates in typography will be reviewed by component.

ez platform 2
ez platform 3


inakijv:
Thank you @SylvainGuittard for the review and feedback :-)

src/bundle/Resources/public/scss/_typography.scss
```diff
@@ -0,0 +1,37 @@
...
+    font-size: 1.25rem;
+}
+
+h5 {
+    font-size: 1.125rem;
+}
+
+.ez-dashboard .ez-main-row {
+    h4 {
+        font-weight: bold;
```
sunpietro: Isn't `bold` set by default for all `hx` elements?

src/bundle/Resources/public/scss/_typography.scss
```diff
@@ -0,0 +1,37 @@
...
+    }
+}
+
+.ez-page-title {
+    h1 {
+        font-weight: bold;
```
sunpietro: Isn't `bold` set by default for all `hx` elements?

src/bundle/Resources/public/scss/_typography.scss
```diff
@@ -0,0 +1,37 @@
...
+    }
+}
+
+.ez-content-edit {
+    h1 {
+        font-weight: bold;
```
sunpietro: Isn't `bold` set by default for all `hx` elements?

src/bundle/Resources/views/parts/page_title.html.twig
```diff
@@ -1,13 +1,13 @@
 {% if title is defined %}
     <div class="ez-page-title pt-3">
-        <h2>
+        <h1>
```
sunpietro: Instead of changing the header types we should use CSS classes to style the elements, so we won't have to care about the type of HTML element in our selectors. Our app is not built in a semantic way because there's no point doing it here. Google robots will never crawl the app ;)

SylvainGuittard: I disagree. Even if Google does not crawl our app, we should follow the basic recommendations of web accessibility.
This code is public and should be an example for our community and other developers. 
I would rather have `h1` than `h2` for the page title.

src/bundle/Resources/public/scss/_typography.scss
```diff
@@ -0,0 +1,37 @@
...
+h3 {
+    font-size: 1.375rem;
+}
+
+h4 {
+    font-size: 1.25rem;
+}
+
+h5 {
+    font-size: 1.125rem;
```
dew326: This `hx` font-size can be set by bootstrap variables. In `_custom.scss` you should set variables, for default in boostrap it is: 
```
$h1-font-size: 2.5rem !default;
$h2-font-size: 2rem !default;
$h3-font-size: 1.75rem !default;
$h4-font-size: 1.5rem !default;
$h5-font-size: 1.25rem !default;
$h6-font-size: 1rem !default;
```

src/bundle/Resources/views/parts/page_title.html.twig
sunpietro: Our app is not accessible then.

src/bundle/Resources/public/scss/_typography.scss
```diff
@@ -0,0 +1,37 @@
...
+}
+
+.ez-page-title {
+    h1 {
+        font-weight: bold;
+    }
+}
+
+.ez-content-edit {
+    h1 {
```
dew326: I think the selector is too generic it will have an impact on every `h1` in content edit. Why only `h1` and only in content edit (and some other places)?
Maybe it's good idea to set it by boostrap variable for every `hx` in the app? 
```
$headings-font-weight:   500 !default;
```

src/bundle/Resources/public/scss/_field-group.scss
```diff
@@ -1,29 +1,44 @@
...
+        .ez-content-field-name {
+            background-color: $ez-ground-base;
+            margin-bottom: 0;
+            padding: .5rem 1.5rem;
+            border-top-right-radius: 5px;
+            border-top-left-radius: 5px;
+        }
+
+        .ez-content-field-value {
+            background-color: $white;
```
dew326: `$ez-white`

src/bundle/Resources/views/dashboard/dashboard.html.twig
```diff
@@ -7,12 +7,32 @@
...
+                <div class="card border-light mb-4">
+                    <div class="card-body">
+                        <h2 class="mb-3">{{ 'everyone'|trans|desc('Everyone') }}</h2>
+                        {{ ez_platform_tabs('dashboard-everyone', []) }}
+                    </div>
+                </div>
             </div>
         </div>
     </div>
+    <div id="react-udw"></div>
```
dew326: I think we add this in `layout.html.twig`. Not needed.

inakijv:
Most important aspects of this update:
- Updated headings size & weight according to design Guidelines.
- Applied basic HTML structure to each view based on Typography & [Web Accessibility standards](https://www.w3.org/WAI/tutorials/page-structure/headings/).
- PRs involved: UI Colors #83 

screen shot 2017-11-30 at 9 24 49 am
screen shot 2017-11-30 at 8 43 17 am
When in Admin section, Content Type, creating or editing a CT, `background-color` sass variable is pending of #83 to be merged (that's why the screenshot below shows a contrast that it's not the expected one)
screen shot 2017-11-30 at 8 41 56 am

inakijv:
Conflicts solved and added changes suggested. Ping @dew326 @sunpietro @SylvainGuittard

src/bundle/Resources/public/scss/_field-group.scss
```diff
@@ -1,29 +1,44 @@
...
 
-		.ez-content-field-name {
-			background-color: $ez-ground-base;
-			margin-bottom: 0;
-			padding: .5rem 1.5rem;
-			border-top-right-radius: 5px;
-			border-top-left-radius: 5px;
-		}
+    h3 {
+        &.ez-fieldgroup-name {
```
sunpietro: What do you think about making a generic selector: `.ez-fieldgroup-name` which contains color, border, padding-bottom and font-weight declarations? Then, in specific cases like: `h3.ez-fieldgroup-name` you'll apply the font-size rules only.

src/bundle/Resources/public/scss/_field-group.scss
```diff
@@ -1,29 +1,44 @@
...
-	}
+    .ez-content-field {
+        padding-left: 2rem;
+        margin-bottom: 1.5rem;
+
+        .ez-content-field-name {
+            background-color: $ez-ground-base;
+            margin-bottom: 0;
+            padding: .5rem 1.5rem;
+            border-top-right-radius: 5px;
```
sunpietro: Nitpick: There's a shorthand for border-radius styles: `border-radius: 5px 0 0 5px;`

src/bundle/Resources/public/scss/_field-group.scss
SylvainGuittard: I think you're right. It can be useful.
At some point we will have to rework `.ez-fieldgroup-name`
because it has to be specific to fieldgroup only (IMHO)
(cf. https://github.com/ezsystems/ezplatform-admin-ui/pull/113/files#diff-54b008e73d6d6a4eab6497a8dcb64a46R2)

lserwatka:
@inakijv could you re-base here and fix conflicts?

inakijv:
@lserwatka reviewed and ready.

PR #99, opened 2017-11-23
EZP-28312: [Date and Time fields] Backspace key doesn't clear field properly

sunpietro:
https://jira.ez.no/browse/EZP-28312

EZP-28312


sunpietro:
@m-tyrala issue fixed

src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
```diff
@@ -55,7 +55,17 @@
         formatDate: (date) => (new Date(date)).toLocaleString()
     };
     const updateInputValue = (sourceInput, date) => {
+        const event = new CustomEvent(EVENT_VALUE_CHANGED);
+
         date = new Date(date);
+
+        if (isNaN(date)) {
```
webhdx: You should validate `date` argument before initializing `Date` object with it.

sunpietro: I get empty Date object that's why I try to create a new Date and then with an empty date object I get `NaN`

webhdx: It's a code smell which only works because JavaScript doesn't support exceptions... You should validate input in the beginning before creating *invalid* objects. From what I noticed if field is empty `date` is an empty array, it's better to check for an empty array earlier.

sunpietro: Then it should be fixed in the flatPickr library itself. I just do the checking on change event.

PR #101, opened 2017-11-24
EZP-28317: Can create content despite ContentType's Container setting

webhdx:
> JIRA: https://jira.ez.no/browse/EZP-28317

This PR disabled Create button if content type `container` setting is set to `no/false`. Also I passed more data to the contextual sidebar builder to avoid accessing these objects from services.

### DIFF
--- a/src/lib/Menu/ContentRightSidebarBuilder.php
+++ b/src/lib/Menu/ContentRightSidebarBuilder.php
@@ -9,6 +9,7 @@ namespace EzSystems\EzPlatformAdminUi\Menu;
 use eZ\Publish\API\Repository\Exceptions as ApiExceptions;
 use eZ\Publish\API\Repository\PermissionResolver;
 use eZ\Publish\API\Repository\Values\Content\Location;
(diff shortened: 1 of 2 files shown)


src/lib/Menu/ContentRightSidebarBuilder.php
```diff
@@ -67,10 +71,17 @@ public function createStructure(array $options): ItemInterface
...
         /** @var Location $location */
         $location = $options['location'];
+        /** @var ContentType $contentType */
+        $contentType = $options['content_type'];
         /** @var ItemInterface|ItemInterface[] $menu */
         $menu = $this->factory->createItem('root');
-        $canCreate = $this->permissionResolver->hasAccess('content', 'create');
-        $canEdit = $this->permissionResolver->canUser('content', 'edit', $location->getContentInfo());
+        $canCreate = $this->permissionResolver->hasAccess('content', 'create')
+            && $contentType->isContainer;
```
andrerom: nitpick: can do these in opposite order to avoid a permission lookup.

PR #102, opened 2017-11-25
removed adding admin access to siteaccess groups (separate scopes for…

skonrad:
| Question      | Answer
| ------------- | ---
| Tickets       | N/A
| Bug fix?      | maybe
| New feature?  | no
| BC breaks?    | not sure
| Tests pass?   | yes
| Doc needed?   | no

This PR removes the addition of the admin access to the regular group it is based on.
Currently the system adds the admin siteaccess to e.g. `frontend`-siteaccess group and therefore gets the override-configuration from this group.

This breaks the admin UI as soon as you have multiple siteaccess' using the same group for override definition.

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/lib/SiteAccess/AdminFilter.php
+++ b/src/lib/SiteAccess/AdminFilter.php
@@ -44,7 +44,6 @@ class AdminFilter implements SiteAccessConfigurationFilter
                 ? str_replace('_group', '', $groupName) . '_admin'
                 : 'admin';
             $configuration['list'][] = $adminSiteAccessName;
-            $configuration['groups'][$groupName][] = $adminSiteAccessName;
             $configuration['groups'][self::ADMIN_GROUP_NAME][] = $adminSiteAccessName;
         }
 


bdunogier:
I see why this needs to be improved, indeed. My only concern with the change is that we lose the information that maps an admin to the group it is an admin for. Adding that info would require that we add a config key so that filters can add this information.

skonrad:
Maybe adding a copy of the siteaccess-group (instead of adding the admin directly to the group) with the added admin siteacccess would be sufficient?

Another option imo would be to check if the admin siteaccess is defined so users could define their own admin siteaccess with a different matching (if required).
This way one could change the matching to something like `admin.example.com`.

emodric:
Usecase where one group needs one admin siteaccess is fairly trivial and assuming one. Most of our installations are a single site with multiple siteaccess groups which only need one admin siteaccess. So I'm all for this change :+1:

lserwatka:
I'm closing this one. It is not relevant anymore. admin siteaccess is now a native product configuration in ezplatform.yml

PR #103, opened 2017-11-27
EZP-28323: Added pagination to Trash

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28323
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   |no

This PR add pagination to Trash and contain refactor of trash restore form.

#### Checklist:
- [x] Ready for Code Review


lserwatka:
@mikadamczyk could you create a JIRA ticket for this task?

src/bundle/Resources/config/services/controllers.yml
```diff
@@ -34,3 +34,10 @@ services:
         public: true
         arguments:
             $defaultSiteaccess: '%ezpublish.siteaccess.default%'
+
+    EzSystems\EzPlatformAdminUiBundle\Controller\TrashController:
+        public: true
+        arguments:
+            $defaultLimit: '$trash_limit$'
+
+
```
sunpietro: No extra line required :P

src/bundle/Resources/config/ezplatform_default_settings.yml
```diff
@@ -3,3 +3,4 @@ parameters:
     ezsettings.default.location_ids.media: 43
     ezsettings.default.location_ids.users: 5
     ezsettings.default.search_limit: 10
+    ezsettings.default.trash_limit: 10
```
sunpietro: What does it say? Display only 10 items on the list?

src/bundle/Resources/views/admin/trash/list.html.twig
```diff
@@ -66,6 +72,16 @@
                     </tbody>
                 </table>
                 {{ form_end(form_trash_item_restore) }}
+
+                <div class="row justify-content-center align-items-center">
+                    <h6>{{ 'trash.viewing'|trans({'%viewing%': pager.currentPageResults|length, '%total%': pager.nbResults})|desc('Viewing %viewing% out of %total% items') }}</h6>
+                </div>
+                <div class="row justify-content-center align-items-center">
+                    {% if pager.haveToPaginate %}
```
dew326: I would not render wrapper when pagination is not displayed.

barbaragr:
1. There is a missing translation:
screen shot 2017-11-27 at 15 26 07

2. When I restore an item, current page number is placed over a message:
screen shot 2017-11-27 at 15 25 32

3. If we need a tab title, it's missing here (for ex. in Admin/Languages it exists)
screen shot 2017-11-27 at 15 29 53

4. When trash is empty, there is also a label "Trash items". I didn't see it on a master branch, so maybe it appears now.
screen shot 2017-11-27 at 15 28 58

mikadamczyk:
All issue fixed. Current title is 'Trash', as 'Language' in Admin->Languages

composer.json
```diff
@@ -20,7 +20,8 @@
         "ezsystems/repository-forms": "^2.0@dev",
         "ezsystems/ezplatform-admin-ui-modules": "^1.0@dev",
         "ezsystems/ez-support-tools": "^0.2@dev",
-        "knplabs/knp-menu-bundle": "^2.1"
+        "knplabs/knp-menu-bundle": "^2.1",
+        "white-october/pagerfanta-bundle": "^1.1"
```
alongosz: Please move it up one line, so we see diff in line that actually got added ;)

src/bundle/Resources/config/services/pagination.yml
```diff
@@ -0,0 +1,12 @@
+services:
+    ez_template:
+        class: EzSystems\EzPlatformAdminUiBundle\View\Template\EzPagerfantaTemplate
+        arguments:
+            - "@translator"
```
alongosz: Please use single quotes

src/bundle/Resources/config/services/pagination.yml
```diff
@@ -0,0 +1,12 @@
...
+
+    EzSystems\EzPlatformAdminUiBundle\View\EzPagerfantaView:
+        public: false
+        arguments:
+            - "@ez_template"
```
alongosz: Please use single quotes

src/lib/Pagination/Pagerfanta/TrashItemAdapter.php
```diff
@@ -0,0 +1,80 @@
...
+    public function getNbResults()
+    {
+        if (isset($this->nbResults)) {
+            return $this->nbResults;
+        }
+
+        $countQuery = clone $this->query;
+        $countQuery->limit = 0;
+
+        return $this->nbResults = $this->trashService->findTrashItems($countQuery)->count;
```
alongosz: Hmm, this probably needs PAPI improvement, because in case of 10000 items we still need to fetch them all to get count.
// cc @andrerom 

@mikadamczyk question for now: is it possible to use pagination w/o information about count? Yeah, that's kinda silly, but maybe...?

src/bundle/Resources/config/services/pagination.yml
webhdx: `ez_template` doesn't feel like proper name. Please use more descriptive name.

src/lib/Form/Type/Trash/TrashRestoreType.php
```diff
@@ -0,0 +1,58 @@
...
+
+use EzSystems\EzPlatformAdminUi\Form\Data\Trash\TrashItemRestoreData;
+use EzSystems\EzPlatformAdminUi\Form\Type\UniversalDiscoveryWidget\UniversalDiscoveryWidgetType;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\CollectionType;
+use Symfony\Component\Form\Extension\Core\Type\SubmitType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class TrashRestoreType extends AbstractType
```
webhdx: Name is misleading to me. Should be `TrashItemRestoreType` - you are restoring TrashItems not the Trash.

src/lib/Form/Type/Trash/TrashItemCheckboxType.php
```diff
@@ -0,0 +1,64 @@
...
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\Form\FormView;
+use EzSystems\EzPlatformAdminUi\Form\DataTransformer\TrashItemTransformer;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class TrashItemCheckboxType extends AbstractType
+{
+    /**
+     * @var TrashService
```
webhdx: Use single line docblock.

src/lib/Form/Type/Trash/TrashItemCheckboxType.php
```diff
@@ -0,0 +1,64 @@
...
+            'compound' => false,
+        ]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getBlockPrefix()
+    {
+        return 'trash_item_checkbox';
```
webhdx: Is it required? We keep it out of our `Type` classes so that the names are generated from class names.

src/lib/Pagination/Pagerfanta/TrashItemAdapter.php
andrerom: @alongosz Given what is available, API use seems okish, and we can with not to much effort do work under the hood to across 6.x and 7.x to improve it to just do a count in this case.

That said the api end point should ideally be either deprecated in favor of a much simpler one, or we need to model some kind of query that should be supported for trash, as the query itself is ignored for trash atm.

PR #105, opened 2017-11-27
EZP-27905: As an Editor, I want to edit a Keyword field type

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-27905
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes/no
| Doc needed?   | yes/no

Created the Keyword field type.

EZP-27905

Requires: https://github.com/ezsystems/ezplatform-admin-ui-assets/pull/2

## THINGS DONE:
- [x] implemented creation of keywords/tags,
- [x] returning to edit mode
- [x] validation of input.


src/bundle/Resources/public/js/scripts/fieldType/ezkeyword.js
```diff
@@ -0,0 +1,70 @@
+(function (global) {
+    const SELECTOR_FIELD = '.ez-field-edit--ezkeyword';
+    const SELECTOR_LABEL_WRAPPER = '.ez-field-edit__label-wrapper';
+
+    class EzKeywordValidator extends global.eZ.BaseFieldValidator {
+        validateKeywords(event) {
```
dew326: missing doc

src/bundle/Resources/public/js/scripts/fieldType/ezkeyword.js
```diff
@@ -0,0 +1,70 @@
...
+            if (isError) {
+                result.errorMessage = global.eZ.errors.emptyField.replace('{fieldName}', label);
+            }
+
+            return result;
+        }
+    }
+
+    const updateValue = (input, event) => {
+        input.value = event.detail.tags.map(tag => tag.label).join();
```
dew326: Is it correct? I think it will join tags `Platform` and `Studio` to `PlatformStudio` or the `,` is part of the tag (I assume backend wants comma separated value)?

src/bundle/Resources/public/js/scripts/fieldType/ezkeyword.js
```diff
@@ -0,0 +1,70 @@
...
+                    callback: 'validateKeywords',
+                    errorNodeSelectors: [SELECTOR_LABEL_WRAPPER],
+                }
+            ],
+        });
+        const taggify = new window.Taggify({
+            containerNode: taggifyContainer,
+            displayLabel: false,
+            displayInputValues: false,
+            hotKeys: [188]
```
dew326: It might be a good idea to explain in the comment which key is the `188`.

src/bundle/Resources/public/js/scripts/fieldType/ezkeyword.js
sunpietro: The `join()` function adds a coma by default if no delimiter is set, so there's no reason to add it as a param.

src/bundle/Resources/views/fieldtypes/ezkeyword.html.twig
```diff
@@ -0,0 +1,6 @@
+{%- block ezplatform_fieldtype_ezkeyword_widget -%}
+    <div class="ez-data-source__taggify form-control"></div>
+    <div class="ez-data-source__input-wrapper" hidden>
+        {{- block('form_widget') -}}
```
webhdx: Can't we just hide the input itself instead of creating a boilerplate `div`?

sunpietro: It might be done in many ways. It's just one of them. What do you suggest instead?
Please, provide a code sample.

sunpietro:
We will need to bump `ezplatform-admin-ui-assets` in order to incorporate the latest changes in the Taggify library: https://github.com/sunpietro/taggify/releases/tag/v1.2.1

It will fix the issue with: _When we enter word in field without adding comma at the end, and then publish, then value is not saved._

sunpietro:
@m-tyrala all the issues are fixed

src/bundle/Resources/views/fieldtypes/ezkeyword.html.twig
webhdx: To hide the input itself:
```
{%- block ezplatform_fieldtype_ezkeyword_widget -%}
    {% set attr = attr|merge({'hidden': 'hidden'}) %}
    <div class="ez-data-source__taggify form-control"></div>
    {{- block('form_widget') -}}
{%- endblock -%}
```

lserwatka:
@sunpietro if there is dependency on ezsystems/ezplatform-admin-ui-assets you must always make a new tag, otherwise it is impossible to test it by our QA.

sunpietro:
@lserwatka it's not true. Anyway, I'll make a new tag :)

webhdx:
@sunpietro It would be easier for everyone, trust me. I wasted 30 minutes yesterday to setup keyword fieldtype just to check something for you. Alternatively you can just provide a brief instruction in PR's description.

lserwatka:
@sunpietro it is absolutely true. I also lost 20 min with Maciej. Pointing to PR that requires bower to install the newer version of the package is very difficult to test.

lserwatka:
@m-tyrala what is the status here?

lserwatka:
@sunpietro could you re-base here?

PR #111, opened 2017-11-29
EZP-27894: As an Editor I want to have access to the user menu in Platform UI v2

webhdx:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-27894
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | yes

#### Checklist:
- [x] Ready for Code Review


barbaragr:
Clickable area could be smaller (marked below). Apart from that everything works fine.
screen shot 2017-11-29 at 12 44 28

src/bundle/DependencyInjection/EzPlatformAdminUiExtension.php
```diff
@@ -41,10 +41,30 @@ public function load(array $configs, ContainerBuilder $container)
...
     public function prepend(ContainerBuilder $container)
+    {
+        $this->prependViews($container);
+        $this->prependImageVariations($container);
+    }
+
+    /**
+     * @param ContainerBuilder $container
+     */
+    private function prependViews(ContainerBuilder $container): void
```
sunpietro: Just curiosity, what does `void` at the end means? This is something new for me.

src/bundle/Resources/public/js/scripts/admin.user.menu.js
```diff
@@ -0,0 +1,22 @@
+(function (global, doc) {
+    const SELECTOR_MENU_ITEMS = '.ez-user-menu__items';
+    const SELECTOR_NAME_WRAPPER = '.ez-user-menu__name-wrapper';
+    const clickOutsideHandler = event => {
```
sunpietro: CS: Please, wrap `event` in parens :)

src/bundle/Resources/public/js/scripts/admin.user.menu.js
```diff
@@ -0,0 +1,22 @@
...
+        if (event.target.closest(SELECTOR_NAME_WRAPPER)) {
+            return;
+        }
+
+        doc.querySelector(SELECTOR_MENU_ITEMS).classList.add('ez-user-menu__items--hidden');
+        doc.querySelector('body').removeEventListener('click', clickOutsideHandler, false);
+    };
+    const toggleMenuItems = () => {
+        const menuItems = doc.querySelector('.ez-user-menu__items');
+        const methodName = menuItems.classList.contains('ez-user-menu__items--hidden') ? 'addEventListener' : 'removeEventListener';
```
sunpietro: This could be cached: `ez-user-menu__items--hidden`

src/bundle/Resources/config/services/ui_config.yml
```diff
@@ -0,0 +1,26 @@
...
+        autowire: true
+        autoconfigure: true
+        public: false
+
+    EzSystems\EzPlatformAdminUi\UI\Config\Provider\ContentTypeMappings:
+        arguments:
+            $locationMappings: "%ezplatform.multifile_upload.location.mappings%"
+            $defaultMappings: "%ezplatform.multifile_upload.location.default_mappings%"
+            $fallbackContentType: "%ezplatform.multifile_upload.fallback_content_type%"
+            $maxFileSize: "%ez_systems.multifile_upload.max_file_size%"
```
alongosz: Nitpick: single quotes in entire file

src/bundle/Resources/public/scss/_user-menu.scss
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    &__items {
+        position: absolute;
+        top: 100%;
+        right: 0;
+        min-width: 100%;
+        background-color: $ez-white;
+        border-radius: 5px;
+        box-shadow: 0px 0px 10px 0px rgba(0,0,0,.25);
```
sunpietro: `px` could be omitted in `0px`

src/bundle/Resources/views/parts/navigation.html.twig
```diff
@@ -14,14 +14,23 @@
...
-                    <a class="nav-link" href="{{ url('logout') }}">
-                        {{ 'authentication.logout'|trans|desc('Logout') }}
-                    </a>
-                </li>
-            </ul>
+            <div class="ez-user-menu ml-md-auto">
+                <div class='ez-user-menu__name-wrapper'>
+                    {% set user = admin_ui_config.user.user%}
+                    <span class="ez-user-menu__name">{{ user.name }}</span>
+                    {% if admin_ui_config.user.profile_picture_field.value.uri is not null %}
```
alongosz: Q: @webhdx what will happen if Field is null, not just uri property (you're returning it from provider in some edge case)?
(not a Twig expert).

src/bundle/DependencyInjection/EzPlatformAdminUiExtension.php
webhdx: This hints that the method won't return any value. Can be omitted but we enforce type hinting everywhere.

sunpietro: That's interesting. It seems you're not using it in other places in this file.
Thanks, good to know.

PR #113, opened 2017-11-30
EZP-27533: Switch to another translation in the Main content view tab

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-27533
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   | no

This PR add language switcher for the available translations of the content item.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/config/routing.yml
+++ b/src/bundle/Resources/config/routing.yml
@@ -4,8 +4,17 @@ ezplatform.dashboard:
         _controller: 'FrameworkBundle:Template:template'
         template: 'EzPlatformAdminUiBundle:dashboard:dashboard.html.twig'
 
+_ezpublishLocation:
+    path: /content/location/{locationId}/{languageCode}/{viewType}/{layout}
+    defaults:
+        _controller: ez_content:viewLocation
+        viewType: full
+        layout: true
+        languageCode: null
+
 ezplatform.content.edit:
     path: /todo/content/edit
+
 ezplatform.content.view:
     path: /todo/content/view
 
--- a/src/bundle/Resources/views/content/locationview.html.twig
+++ b/src/bundle/Resources/views/content/locationview.html.twig
@@ -81,6 +81,7 @@
 
 {% block javascripts %}
     {%  javascripts
+        'bundles/ezplatformadminui/js/scripts/admin.location.change_language.js'
         'bundles/ezplatformadminui/js/scripts/admin.location.view.js'
         'bundles/ezplatformadminui/js/scripts/admin.location.tab.js'
         'bundles/ezplatformadminui/js/scripts/admin.location.visibility.js'
--- a/src/bundle/Resources/views/content/tab/content.html.twig
+++ b/src/bundle/Resources/views/content/tab/content.html.twig
@@ -1,19 +1,38 @@
-{% for group in fieldDefinitionsByGroup if group.fieldDefinitions|length > 0 %}
-    <section class="ez-fieldgroup">
-        <h5 class="ez-fieldgroup-name">{{ group.name }}</h5>
-        {% for fieldDefinition in group.fieldDefinitions %}
-            <div class="ez-content-field">
-                <p class="ez-content-field-name">{{ ez_field_name(content, fieldDefinition.identifier) }}:</p>
-                <div class="ez-content-field-value">
-                    {% if ez_is_field_empty(content, fieldDefinition.identifier) %}
-                        <em>{{'fieldview.field.empty'|trans({}, 'fieldview')|desc('This field is empty')}}</em>
-                    {% else %}
-                        {{ ez_render_field(content, fieldDefinition.identifier, {
-                            template: '@EzPlatformAdminUi/content_fields.html.twig'
-                        }) }}
-                    {% endif %}
+<section class="ez-fieldgroup ">
+    <div class="ez-fieldgroup-name d-flex justify-content-between mb-3">
+        <h5>
+            {{ 'tab.view.content'|trans()|desc('Content') }}
+        </h5>
+        {% if languages|length > 1  %}
+        <form class="form-inline">
+            <select class="form-control ez-location-language-change">
+                {% for language in languages %}
+                    <option value="{{ path('_ezpublishLocation', {'locationId': location.id, 'languageCode': language.languageCode}) }}"
+                            {% if app.request.get('languageCode') == language.languageCode %} selected="selected" {% endif %}
+                    >{{ language.name }}</option>
+                {% endfor %}
+            </select>
+        </form>
+        {% endif %}
+    </div>
+
+    {% for group in fieldDefinitionsByGroup if group.fieldDefinitions|length > 0 %}
+        <section class="ez-fieldgroup container">
+            <h5 class="ez-fieldgroup-name">{{ group.name }}</h5>
+            {% for fieldDefinition in group.fieldDefinitions %}
+                <div class="ez-content-field">
+                    <p class="ez-content-field-name">{{ ez_field_name(content, fieldDefinition.identifier) }}:</p>
+                    <div class="ez-content-field-value">
+                        {% if ez_is_field_empty(content, fieldDefinition.identifier) %}
+                            <em>{{ 'fieldview.field.empty'|trans({}, 'fieldview')|desc('This field is empty') }}</em>
+                        {% else %}
+                            {{ ez_render_field(content, fieldDefinition.identifier, {
+                                template: '@EzPlatformAdminUi/content_fields.html.twig'
+                            }) }}
+                        {% endif %}
+                    </div>
                 </div>
-            </div>
-        {% endfor %}
-    </section>
-{% endfor %}
+            {% endfor %}
+        </section>
+    {% endfor %}
+</section>
--- a/src/lib/Tab/LocationView/ContentTab.php
+++ b/src/lib/Tab/LocationView/ContentTab.php
@@ -10,6 +10,8 @@ namespace EzSystems\EzPlatformAdminUi\Tab\LocationView;
 
 use eZ\Publish\API\Repository\Values\Content\Content;
 use eZ\Publish\API\Repository\Values\ContentType\ContentType;
+use eZ\Publish\API\Repository\LanguageService;
+use eZ\Publish\API\Repository\Values\Content\Language;
 use eZ\Publish\Core\Helper\FieldsGroups\FieldsGroupsList;
 use EzSystems\EzPlatformAdminUi\Tab\AbstractTab;
 use EzSystems\EzPlatformAdminUi\Tab\OrderedTabInterface;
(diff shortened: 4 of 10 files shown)


src/bundle/Resources/public/js/scripts/admin.location.changeLanguage.js
```diff
@@ -0,0 +1,7 @@
+(function (global, doc) {
+    const changeLocationLanguage = event => {
+        window.location = event.target.options[event.target.selectedIndex].value;
+    };
+
+    doc.querySelector('.change-location-language').addEventListener('change', changeLocationLanguage, false);
```
dew326: the CSS class should be prefixed with `ez-`

src/bundle/Resources/public/js/scripts/admin.location.changeLanguage.js
```diff
@@ -0,0 +1,7 @@
+(function (global, doc) {
+    const changeLocationLanguage = event => {
+        window.location = event.target.options[event.target.selectedIndex].value;
```
dew326: I think select should have it in `value`, so `event.currentTarget.value` should be enough.

src/bundle/Resources/public/js/scripts/admin.location.changeLanguage.js
```diff
@@ -0,0 +1,7 @@
+(function (global, doc) {
+    const changeLocationLanguage = event => {
```
dew326: CS `(event) => {`

src/bundle/Resources/views/content/tab/content.html.twig
```diff
@@ -1,5 +1,21 @@
+<section class="ez-fieldgroup ">
+    <h5 class="ez-fieldgroup-name" style="overflow: hidden">
```
dew326: no inline styles

src/bundle/EventListener/RequestAttributesListener.php
```diff
@@ -60,6 +67,18 @@ public function addRequestAttributes(FilterViewBuilderParametersEvent $event)
...
+            /** @var Location $location */
+            $location = $parameterBag->get('location');
+            $content = $this->loadContent($location->contentInfo->id, $parameterBag->get('language'));
+            $parameterBag->set('content', $content);
+        }
+    }
+
+    private function fetchingTranslatedContent(Request $request, ParameterBag $parameterBag)
+    {
+        return $parameterBag->has('language') && $parameterBag->has('location') && $request->get('_route') === '_ezpublishLocation';
```
webhdx: Wouldn't checking `_route` first be more logical, would it?

src/bundle/EventListener/RequestAttributesListener.php
```diff
@@ -60,6 +67,18 @@ public function addRequestAttributes(FilterViewBuilderParametersEvent $event)
...
+
+        if ($this->fetchingTranslatedContent($request, $parameterBag)) {
+            /** @var Location $location */
+            $location = $parameterBag->get('location');
+            $content = $this->loadContent($location->contentInfo->id, $parameterBag->get('language'));
+            $parameterBag->set('content', $content);
+        }
+    }
+
+    private function fetchingTranslatedContent(Request $request, ParameterBag $parameterBag)
```
webhdx: I don't think this is the best method name. It's a good practice to name methods returning boolean value name with `is` or `has` prefix. In this case I believe `hasContentLanguage` would be more accurate.

You are missing return type hint here and there. Adding docblock wouldn't harm either.

src/bundle/EventListener/RequestAttributesListener.php
```diff
@@ -60,6 +67,18 @@ public function addRequestAttributes(FilterViewBuilderParametersEvent $event)
             $parameterBag->remove('locationId');
             $parameterBag->set('location', $location);
         }
+
+        if ($this->fetchingTranslatedContent($request, $parameterBag)) {
```
webhdx: I'd like to hear others opinion but I wouldn't put that in the same request listener. Put it in separate class or at least in separate method. Also this RequestListener should be moved into `src/lib/EventListener` (this is something we overlooked in one of previous PRs).

src/bundle/EventListener/RequestAttributesListener.php
```diff
@@ -73,6 +92,26 @@ function (Repository $repository) use ($locationId) {
         return $location;
     }
 
+    /**
+     * @param mixed $contentId
```
webhdx: As far as I know `$contentId` will always be `int`.

src/bundle/EventListener/RequestAttributesListener.php
```diff
@@ -73,6 +92,26 @@ function (Repository $repository) use ($locationId) {
...
+     * @param string $language
+     *
+     * @return Content
+     *
+     * @throws UnauthorizedException
+     * @throws NotFoundException
+     */
+    private function loadContent($contentId, string $language): Content
+    {
+        $content = $this->repository->sudo(
```
webhdx: Are you aware that using `sudo` here will load the content ignoring user permissions?

src/bundle/Resources/config/routing.yml
```diff
@@ -4,8 +4,17 @@ ezplatform.dashboard:
         _controller: 'FrameworkBundle:Template:template'
         template: 'EzPlatformAdminUiBundle:dashboard:dashboard.html.twig'
 
+_ezpublishLocation:
+    path: /content/location/{locationId}/{language}/{viewType}/{layout}
```
webhdx: I would be more specific with parameter names - `language` is too generic, especially because Language uses either unique `languageCode` or `languageId`. You should call it `languageCode` in this case.

src/bundle/Resources/public/js/scripts/admin.location.changeLanguage.js
webhdx: @dew326 Shouldn't the class name use BEM naming convention, should it?

src/bundle/Resources/views/content/tab/content.html.twig
```diff
@@ -1,5 +1,21 @@
...
+        <div class="float-left">
+        {{ 'fieldview.content'|trans({}, 'fieldview')|desc('Content') }}
```
webhdx: Why are you using `fieldview` translation domain? I can't recall we are using it anywhere in the admin UI. Location View tabs are using `locationview` domain name as far as I remember. Also you should check how we are composing translation keys in tabs.

src/bundle/Resources/views/content/tab/content.html.twig
```diff
@@ -1,5 +1,21 @@
...
+        </div>
+        <form class="form-inline float-right">
+            <select class="form-control change-location-language">
+                {% for language in languages %}
+                    <option value="{{ path('_ezpublishLocation', {'locationId': location.id, 'language': language.languageCode}) }}"
+                        {% if app.request.get('language') == language.languageCode %} selected="selected" {% endif %}
```
webhdx: How about passing selected language as a parameter to the template instead of accessing the request?

src/lib/Tab/LocationView/ContentTab.php
```diff
@@ -60,12 +68,21 @@ public function renderView(array $parameters): string
         $contentType = $parameters['contentType'];
         $fieldDefinitions = $contentType->getFieldDefinitions();
         $fieldDefinitionsByGroup = $this->groupFieldDefinitions($fieldDefinitions);
+        $contentLanguages = $content->versionInfo->languageCodes;
+
+        $languages = $this->loadLanguages(
```
webhdx: You should probably also filter out disabled Languages.  Language can be disabled but it will be still visible in the content.

src/lib/Tab/LocationView/ContentTab.php
```diff
@@ -60,12 +68,21 @@ public function renderView(array $parameters): string
...
+            function (Language $language) use ($contentLanguages) {
```
webhdx: Do we need to pass a callable here? Can't you put this in `loadLanguages` method directly?

src/bundle/EventListener/RequestAttributesListener.php
mikadamczyk: `_route` check is the heaviest one

src/bundle/EventListener/RequestAttributesListener.php
mikadamczyk: In `loadContent` method it is typed as mixed, but apparently it is integer in this case

mikadamczyk:
@dew326 @webhdx could you take a look at this?

src/bundle/Resources/config/routing.yml
```diff
@@ -4,8 +4,17 @@ ezplatform.dashboard:
         _controller: 'FrameworkBundle:Template:template'
         template: 'EzPlatformAdminUiBundle:dashboard:dashboard.html.twig'
 
+_ezpublishLocation:
+    path: /content/location/{locationId}/{languageCode}/{viewType}/{layout}
+    defaults:
+        _controller: ez_content:viewLocation
+        viewType: full
+        layout: true
+        languageCode: 'eng-GB'
```
webhdx: You can't hardcode `eng-GB` here. It can be completely different depending on the setup.

src/bundle/Resources/views/content/tab/content.html.twig
```diff
@@ -1,19 +1,38 @@
...
+    <h5 class="ez-fieldgroup-name clearfix">
+        <div class="float-left">
+            {{ 'tab.view.content'|trans()|desc('Content') }}
+        </div>
+        {% if languages|length > 1  %}
+        <form class="form-inline float-right">
+            <select class="form-control ez-location-language-change">
+                {% for language in languages %}
+                    <option value="{{ path('_ezpublishLocation', {'locationId': location.id, 'languageCode': language.languageCode}) }}"
+                            {% if app.request.get('languageCode') == language.languageCode %} selected="selected" {% endif %}
```
webhdx: How about passing selected language as a parameter to the template instead of accessing the request?

src/bundle/Resources/views/content/tab/content.html.twig
```diff
@@ -1,19 +1,38 @@
...
-                <div class="ez-content-field-value">
-                    {% if ez_is_field_empty(content, fieldDefinition.identifier) %}
-                        <em>{{'fieldview.field.empty'|trans({}, 'fieldview')|desc('This field is empty')}}</em>
-                    {% else %}
-                        {{ ez_render_field(content, fieldDefinition.identifier, {
-                            template: '@EzPlatformAdminUi/content_fields.html.twig'
-                        }) }}
-                    {% endif %}
+<section class="ez-fieldgroup ">
+    <h5 class="ez-fieldgroup-name clearfix">
```
dew326: This HTML is invalid. According to the HTML specification, `h5` cannot have `div` inside.

src/bundle/Resources/views/content/locationview.html.twig
```diff
@@ -81,6 +81,7 @@
 
 {% block javascripts %}
     {%  javascripts
+        'bundles/ezplatformadminui/js/scripts/admin.location.changeLanguage.js'
```
dew326: Please do not use camelCase in files name for js.

src/bundle/Resources/public/js/scripts/admin.location.changeLanguage.js
```diff
@@ -0,0 +1,11 @@
+(function (global, doc) {
+    const changeLocationLanguage = (event) => {
+        window.location = event.currentTarget.value;
+    }
```
dew326: missing `;`

src/bundle/Resources/public/js/scripts/admin.location.changeLanguage.js
```diff
@@ -0,0 +1,11 @@
...
+
```
dew326: empty line not needed

src/bundle/Resources/public/js/scripts/admin.location.changeLanguage.js
```diff
@@ -0,0 +1,11 @@
...
+    const locationLanguageSwitcher = doc.querySelector('.ez-location-language-change');
+
+    if(locationLanguageSwitcher){
```
dew326: CS `if (locationLanguageSwitcher) {`

lserwatka:
@webhdx are you OK with latest changes?

PR #121, opened 2017-12-02
Add some useful blocks to main layout

emodric:
| Question      | Answer
| ------------- | ---
| Tickets       | N/A
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

When integrating third party solutions into Admin UI, sometimes, it's quite useful to inject a piece of markup just before and after `content` Twig block to further refine the integration, in cases where such modification is not possible in the files extending the main layout.

Also, `pageHead` block has been added to be able to inject stuff to page head, like meta tags and so on. 

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/layout.html.twig
+++ b/src/bundle/Resources/views/layout.html.twig
@@ -4,6 +4,9 @@
         <meta charset="UTF-8" />
         <meta name="CSRF-Token" content="{{ csrf_token('authenticate') }}" />
         <meta name="SiteAccess" content="{{ app.request.get('siteaccess').name }}" />
+
+        {% block pageHead %}{% endblock %}
+
         <script>
             window.eZ = {
                 adminUiConfig: {{ admin_ui_config|json_encode|raw }},
--- a/src/bundle/Resources/views/layout.html.twig
+++ b/src/bundle/Resources/views/layout.html.twig
@@ -56,6 +56,7 @@
             </div>
         </div>
         <div class="container-fluid ez-main-container">
+            {% block contentWrapper %}
             {% block content %}
 
                 {% block left_sidebar %}
@@ -65,6 +66,7 @@
                 {% endblock left_sidebar %}
 
             {% endblock content %}
+            {% endblock contentWrapper %}
         </div>
         <div class="ez-notifications-container">
             {% for label, messages in app.flashes %}
--- a/src/bundle/Resources/views/layout.html.twig
+++ b/src/bundle/Resources/views/layout.html.twig
@@ -5,7 +5,7 @@
         <meta name="CSRF-Token" content="{{ csrf_token('authenticate') }}" />
         <meta name="SiteAccess" content="{{ app.request.get('siteaccess').name }}" />
 
-        {% block pageHead %}{% endblock %}
+        {% block page_head %}{% endblock %}
 
         <script>
             window.eZ = {
@@ -59,7 +59,7 @@
             </div>
         </div>
         <div class="container-fluid ez-main-container">
-            {% block contentWrapper %}
+            {% block content_wrapper %}
             {% block content %}
 
                 {% block left_sidebar %}
@@ -69,7 +69,7 @@
                 {% endblock left_sidebar %}
 
             {% endblock content %}
-            {% endblock contentWrapper %}
+            {% endblock content_wrapper %}
         </div>
         <div class="ez-notifications-container">
             {% for label, messages in app.flashes %}


webhdx:
Can't you just extend `layout.html.twig` and use:
```twig
{% block content %}
<p>pre content</p>
{{ parent() }}
<p>post content</p>
{% endblock %}
```

emodric:
I'm going to try it, but I doubt it. In this case `{{ parent() }}` would be empty, would it not, since the actual parent block in `layout.html.twig` from Admin UI is empty?

emodric:
@webhdx Nope, it doesn't work. While the central `content` block gets used from the template my controller renders, `<p>pre content</p>` and `<p>post content</p>` get ignored. The only way to make it work is with separate blocks, like this: https://github.com/netgen/TagsBundle/blob/master/bundle/Resources/views/ezadminui/pagelayout.html.twig#L21-L44

wizhippo:
How about something like the following instead
```
            {% block content_wrapper %}
            {% block content %}

                {% block left_sidebar %}
                    <div class="col-sm-1 bg-dark pt-4 ez-side-menu"> {# @todo sidebars should be moved to layout.html.twig !! #}
                        {{ knp_menu_render('ezplatform_admin_ui.menu.content.sidebar_left', {'template': '@EzPlatformAdminUi/parts/menu/sidebar_left.html.twig'}) }}
                    </div>
                {% endblock left_sidebar %}

            {% endblock content %}
            {% endblock content_wrapper %}
```

Then you can

```
{% block content_wrapper %}
    <div class="ng-tags-app">
        {% include '@NetgenTags/admin/header.html.twig' %}

        <section class="ng-tags-container">
            <div class="tags-sidebar tags-sidebar-resizable">
                <div class="main-menu">
                    {% include '@NetgenTags/admin/main_menu.html.twig' %}
                </div>
            </div>

            <div class="tags-content">
                <div class="path">
                    {% include '@NetgenTags/admin/path.html.twig' %}
                </div>
                {% block content %}{% endblock content %}
            </div>
        </section>
    </div>
{% endblock %}

```

emodric:
@wizhippo I guess that works too. I don't really have a preference, as long as it works :)

webhdx:
I'm more keen on @wizhippo solution as giving `pre` and `post` blocks you are encouraging users to put unclosed tags in the blocks.

emodric:
Okay, I'll fix it ASAP!

emodric:
@webhdx @wizhippo Updated code to add `contentWrapper` block.

webhdx:
Sorry for being picky but can you also use underscore for block names?

emodric:
Sure, I can, but I picked camel case because most of the blocks in `layout.html.twig` are in camel case.

emodric:
@webhdx Changed to snake case.

webhdx:
@Nattfarinn What do you think of introducing components here?

emodric:
> not to mention it will only work with template inheritance

This is the whole point :)

> and will be a pain if two different bundles would like to use same extensibility point.

Not sure what you mean? The point is to allow seamless 3rd party admin interface integrations, like TagsBundle.

Defining two Symfony services just to inject a piece of markup before and after `content` block seems like an overkill. What if that piece of markup needs to access some variable available in the layout template itself? Besides, I do not want to do that globally, but just when displaying TagsBundle admin interface and using its layout: https://github.com/netgen/TagsBundle/blob/master/bundle/Resources/views/ezadminui/pagelayout.html.twig

Nattfarinn:
> > not to mention it will only work with template inheritance
>
> This is the whole point :)
> 
> > and will be a pain if two different bundles would like to use same extensibility point.
> Not sure what you mean? The point is to allow seamless 3rd party admin interface integrations, like TagsBundle.

Inheritance doesn't sound seamless to me if it comes to having multiple bundles trying to extend same template. Let's say, there's another bundle along with TagsBundle, different 3rd party, that would like to add their own markup above content block. End-user expect to have both extensions. Should the 3rd party extend TagsBundle layout, or should you extend theirs, and why when only dependency of each is nothing but AdminUI? Isn't it better to inject custom piece of markup instead of inheriting layout and replace the wrapper? :)

> What if that piece of markup needs to access some variable available in the layout template itself?

Just pass the context. It's passed to every group component when creating extensibility point (and can be overwritten from service argument if you like).

```
{{ ezplatform_admin_ui_component_group('pre-content', {
    'content': content,
    'something': somethingElse,
}) }}
```

> Defining two Symfony services just to inject a piece of markup before and after content block seems like an overkill.

Anyone can write any number of such services (even more complex ones), and just mark them with desired component group to make them appear where they should. :) It's just more flexible even if it seems to be an overkill at first.

emodric:
TagsBundle doesn't extend Admin UI Bundle, only a template from Tags Bundle extends a template from Admin UI Bundle so there won't be any 3rd party bundles colliding. Besides, extending bundles is deprecated in Symfony 3.4 and was removed in Symfony 4.

My usecase, and the point of this PR, is using that additional markup purely in routes that I, or rather, Tags Bundle, fully controls, so injecting a custom piece of markup globally is not an answer. In our case, that markup injects a whole tags tree to the left of the main content block. You wouldn't wan't the tags tree appearing everywhere?

> Anyone can write any number of such services (even more complex ones), and just mark them with desired component group to make them appear where they should. :) It's just more flexible even if it seems to be an overkill at first.

As I said, I don't want to inject the markup in all pages, only when displaying TagsBundle routes.

Nattfarinn:
And the purpose of these components feature is to achieve exactly what you want. `TwigComponent` is just a basic component, nothing else but wrapper for twig render, but you can create ie. `TagsTreeComponent` (only requirement is to implement `Renderable` interface), and decide inside of it how, when and why it will be displayed. Yet, you wont block any other extensions to do the same (and it would be a case using only twig blocks) while `block content_wrapper` solves just a single use case.

emodric:
I'm sorry, but I'm still not convinced. The need for creating aditional Symfony services to render a couple of aditional lines of markup is counterintuitive to the whole DX experience which should allow for much easier integration than before.

I'll go ahead and close this PR and integrate Tags Bundle when the support 3rd party admin interface integrations is finalized :)

wizhippo:
@Nattfarinn any spec or docs on these components?  How would you add javascript/css required by a component?  You would have to create a component for a css component and a javascript component and a component  for the component itself?  Seeing these are components it would be nice if you could define the resources needed in the component itself.
@emodric just wondering why the left tree is part of the layout and not just part of the content?

emodric:
@wizhippo Because that was the only way we could make it work when integrating into eZ Platform UI and Netgen Admin UI with a constraint that we didn't want to touch the controller result templates. But other than that, it's natural that it's a part of the layout and not content part, since it's used everywhere, no?

Nattfarinn:
@wizhippo It's done already. Currently we have three Components by default: `TwigComponent`, `ScriptComponent` and `LinkComponent`. In layout there are already extensibility points defined:
```
         {{ ezplatform_admin_ui_component_group('stylesheet-head') }}
         {{ ezplatform_admin_ui_component_group('script-head') }}
         {{ ezplatform_admin_ui_component_group('stylesheet-body') }}
         {{ ezplatform_admin_ui_component_group('script-body') }}
```

No documentation on that yet, but you can give it a try even now (figuring out what's going on from the code is very easy).

wizhippo:
@emodric I'm just looking at the examples that admin ui v2 is providing where the left/right panels are part of the content and not the layout.  So following that would suggest we should do the same.  For example you could put the tree in left_sidebar.  But I do understand the challenge now supporting both v1 and v2 admin.

emodric:
@wizhippo Maybe, I'll look into it. Btw, it's not just eZ admin v1 and v2, but Netgen Admin and standalone interface, so definititelly a challenge :)

lserwatka:
@emodric Creating service is a standard DX for every Symfony developer. I can't imagine any difficulties for Symfony developer in this approach. This is just basics of basics. Thus I  fully disagree with a statement that it is counter intuitive. We will always favour flexible extensibility point over the basic that serves  one purpose only. This extensibility point will be officially documented.

Nattfarinn:
> True, but using them in Twig templates feels wrong, nothing else.

You're not using them in Twig template. They serve the purpose to be rendered into HTML. It's no different to KNP Menu Bundle. Your need is very simple, but the tool is powerful.

PR #124, opened 2017-12-05
EZP-28397: Update icons for table toolbar in AlloyEditor

dew326:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28397
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Screenshot:
table-icons

#### Checklist:
- [x] Ready for Code Review


src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-tablecell.js
```diff
@@ -0,0 +1,50 @@
...
+export default class EzBtnTableCell extends AlloyEditor.ButtonTableCell {
+    static get key() {
+        return 'eztablecell';
+    }
+
+    /**
+     * Lifecycle. Renders the UI of the button.
+     *
+     * @method render
+     * @return {Object} The content which should be rendered.
```
sunpietro: I think we can skip the docs related to lifecycle methods in React components.

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-tablecell.js
```diff
@@ -0,0 +1,50 @@
...
+     */
+    render() {
+        let buttonCommandsList;
+        let buttonCommandsListId;
+
+        if (this.props.expanded) {
```
sunpietro: I would move it to a separate method to make `render` function body as small as possible.

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-tablecolumn.js
```diff
@@ -0,0 +1,51 @@
...
+     * Lifecycle. Renders the UI of the button.
+     *
+     * @method render
+     * @return {Object} The content which should be rendered.
+     */
+    render() {
+        let buttonCommandsList;
+        let buttonCommandsListId;
+
+        if (this.props.expanded) {
```
sunpietro: Same as above. The similar suggestion applies to all the situations similar to this one.

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-tablecell.js
dew326: Most of the `render` function is copied from AlloyEditor, I would like to keep it like they do for easier maintenance.

sunpietro: Ok, let's keep it then.

PR #128, opened 2017-12-05
EZP-28356: Improved Visibility switcher style

inakijv:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-28356](https://jira.ez.no/browse/EZP-28356)
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

Most important aspects:
- Improved Visibility switcher style according to design guidelines.

screen shot 2017-12-05 at 5 07 38 pm

#### Checklist:
- [x] Ready for Code Review


src/bundle/Resources/public/scss/_mixins.scss
```diff
@@ -1,34 +1,37 @@
...
+        height: 2rem;
         top: 50%;
         left: 0;
         background-color: $ez-color-secondary;
         z-index: -1;
-        border-radius: 5px;
+        border-radius: .25rem;
         transform: translate(calc(100% + .3rem), -50%);
-        transition: all .2s $ez-admin-transition;
+        transition: all .2s ease-in;
```
sunpietro: Just curiosity, what was wrong with the previous timing function?

dew326: Q: shouldn't we have one transition style across whole application?

lserwatka: I agree, we should have one type for consistency.

lserwatka:
@inakijv rebase is needed here

src/bundle/Resources/public/scss/_mixins.scss
inakijv: There’s nothing wrong about the cubic bezier option :-)

I am more inclined to options like ease-in because they depict a natural movement that is aligned with user expectations when an event is triggered; and also a movement that doesn’t demand breaking user’s attention.

Ease-in with a long timing undermines usability. Here is proposed .2s, but at the same time the effect is not that fast that breaks user’s attention.

inakijv: I am not totally sure about if we should apply the same kind of animation and timing for all cases. We have to be consistent, but differentiate cases based on the experience we would like to define.

sunpietro: I believe that for now, we should use the same transition timing function everywhere and do some research what works better in each case. Transitions should be meaningful (in some way). For me, using standard easing functions makes UI less responsive.

PR #129, opened 2017-12-06
EZP-28403: Design improvements for Sections

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28403
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Improve design for Section create and edit in Admin Panel

screen shot 2017-12-06 at 9 24 00 am
screen shot 2017-12-06 at 9 23 50 am

#### Checklist:
- [x] Ready for Code Review
- [x] Add right sidebar with action buttons
- [x] Keep Main Menu active on these sub-actions


lserwatka:
@webhdx could you check this code for menu?

lserwatka:
@sunpietro & @dew326 could you review frontend part here?

src/bundle/Resources/public/scss/_general.scss
```diff
@@ -39,7 +39,7 @@ body {
     height: auto;
 }
 
-button {
+button, a {
```
sunpietro: Maybe we should remove the `button, a` part and make the `.btn` styles more generic? cc @dew326

dew326: Good point, +1

webhdx:
Looks good. We might think of separating menu builders into different namespaces as `Menu\` is getting really bloated by the Builders. But this is something we can do with another menu once we need to add one.

PR #131, opened 2017-12-06
EZP-28414: Design improvements for Role list in Admin Panel

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28414
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

Design improvements for Section create and edit in Admin Panel

screen shot 2017-12-06 at 4 24 53 pm

#### Checklist:
- [ ] Add bulk delete for roles
- [ ] Ready for Code Review


src/bundle/Resources/public/scss/_icons.scss
```diff
@@ -10,6 +10,10 @@
         width: 1.5rem;
         height: 1.5rem;
         fill: $ez-color-primary;
+
+        &--dark {
+            fill: $ez-secondary-ground-dark
```
inakijv: Could you please change it to `$ez-color-secondary`. Thanks!

mikadamczyk: Of course, done

dew326: I would also change the CSS class name to `ez-icon--secondary`

src/bundle/Resources/public/js/scripts/admin.role.list.js
```diff
@@ -0,0 +1,17 @@
+(function () {
+    const checkboxes = [...document.querySelectorAll('form[name="bulk-remove-role"] input[type="checkbox"]')];
+    const buttonRemove = document.querySelector('#bulk-remove-role');
+
+    const enableButtons = (event) => {
+        const isNonEmptySelection = checkboxes.some(el => el.checked);
```
dew326: you could write something like this:
```js
const methodName = checkboxes.some(el => el.checked ? 'removeAttribute' : 'setAttribute';

buttonRemove[methodName]('disabled', true);
```

src/bundle/Resources/public/js/scripts/admin.role.list.js
```diff
@@ -0,0 +1,17 @@
+(function () {
+    const checkboxes = [...document.querySelectorAll('form[name="bulk-remove-role"] input[type="checkbox"]')];
+    const buttonRemove = document.querySelector('#bulk-remove-role');
+
+    const enableButtons = (event) => {
```
dew326: what do you think to name it `toggleButtonState`?

src/bundle/Controller/RoleController.php
```diff
@@ -79,7 +80,16 @@ public function listAction(): Response
     {
         $roles = $this->roleService->loadRoles();
 
+        $rolesNumbers = array_column($roles, 'id');
+
+        $roleBulkRemoveData = new RoleBulkRemoveData(
```
webhdx: Why do you need another data class instead of refactoring `RoleDeleteData` to accept multiple roles? This would reduce code duplication as you could share this class between removing single role via Role View page and removing multiple roles in the Role List page.

src/bundle/Resources/public/js/scripts/admin.role.list.js
```diff
@@ -0,0 +1,13 @@
+(function () {
+    const checkboxes = [...document.querySelectorAll('form[name="delete-roles"] input[type="checkbox"]')];
+    const buttonRemove = document.querySelector('#delete-roles');
+
+    const toggleButtonState = (event) => {
+        console.log(checkboxes);
```
sunpietro: console log

src/bundle/Resources/public/js/scripts/admin.role.list.js
```diff
@@ -0,0 +1,13 @@
+(function () {
+    const checkboxes = [...document.querySelectorAll('form[name="delete-roles"] input[type="checkbox"]')];
+    const buttonRemove = document.querySelector('#delete-roles');
+
```
sunpietro: No extra line needed

src/bundle/Resources/public/scss/_buttons.scss
```diff
@@ -35,3 +35,9 @@
         margin-right: 0;
     }
 }
+
+.ez-table-header {
+    .ez-icon-create, .ez-icon-trash {
```
sunpietro: CS: In this case, each selector should be in a separate line

src/bundle/Resources/public/scss/_tables.scss
```diff
@@ -22,12 +22,11 @@
 
     &.round {
         border-radius: 4px 4px 0 0;
-        padding: 0.5em 0;
     }
 
     &--label {
         margin: 0;
-        padding: 0 1.4rem;
+        padding: 0 .45rem;
```
sunpietro: Maybe `.4375rem`? Which is equal to 7px.

src/bundle/Resources/public/js/scripts/admin.role.list.js
mikadamczyk: This line is already removed

PR #132, opened 2017-12-06
EZP-28418: Design improvements for Role create and edit actions

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28418
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

Design improvements for Role create and edit actions in Admin Panel

screen shot 2017-12-06 at 5 22 31 pm
screen shot 2017-12-06 at 5 22 18 pm

#### Checklist:
- [x] Ready for Code Review


src/bundle/Resources/views/admin/role/add.html.twig
```diff
@@ -1,39 +1,39 @@
-{% extends "@EzPlatformAdminUi/layout.html.twig" %}
+{% extends 'EzPlatformAdminUiBundle:admin/role:base.html.twig' %}
 
 {% trans_default_domain 'role' %}
 
-{% block breadcrumbs %}
+{% block breadcrumbsAdmin %}
```
webhdx: I know this is all over the place but we should use underscore for block names as this is more common notation.

src/bundle/Resources/views/admin/role/add.html.twig
```diff
@@ -1,39 +1,39 @@
...
     {% include '@EzPlatformAdminUi/parts/breadcrumbs.html.twig' with { items: [
         { value: 'breadcrumb.admin'|trans(domain='messages')|desc('Admin') },
         { url: path('ezplatform.role.list'), value: 'role.breadcrumb.list'|trans|desc('Roles') },
         { value: 'role.breadcrumb.add'|trans|desc('Creating a new Role') }
     ]} %}
 {% endblock %}
 
-{% block pageTitle %}
+{% block pageTitleAdmin %}
```
webhdx: Underscore block name

lserwatka: Yes, I agree. Let's change it to underscore.

PR #135, opened 2017-12-07
EZP-27730: Buttons improvement opacity updated

inakijv:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-27730](https://jira.ez.no/browse/EZP-27730)
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

Most important aspect:
- Updated buttons' opacity based on design specifications.

screen shot 2017-12-07 at 12 23 55 am

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_buttons.scss
+++ b/src/bundle/Resources/public/scss/_buttons.scss
@@ -51,3 +51,8 @@
         font-size: .8125rem;
     }
 }
+
+.btn.disabled,
+.btn:disabled {
+    opacity: .3;
+}


src/bundle/Resources/public/scss/_buttons.scss
```diff
@@ -35,3 +35,8 @@
         margin-right: 0;
     }
 }
+
+.btn.disabled, 
+.btn:disabled {
```
sunpietro: Maybe it should be a generic selector like: `.disabled, *:disabled`

inakijv: Reviewed and updated.

inakijv:
@sunpietro I tested your suggestion, but it didn't work. It has to be more specific.

PR #145, opened 2017-12-08
EZP-28450: Design improvements for Policies and User and Groups lists

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28450
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   | no

Design improvements for Policies and User and Groups lists

screen shot 2017-12-08 at 4 39 56 pm

screen shot 2017-12-08 at 4 39 45 pm

#### Checklist:
- [x] Bulk delete in policies
- [x] Bulk delete in user and groups
- [x] Ready for Code Review


src/bundle/Resources/views/admin/role/view.html.twig
```diff
@@ -43,15 +43,13 @@
...
-                {{ 'role.view.view.actions.edit'|trans|desc('Edit') }}
-            </a>
-            {{ form_widget(delete_form.delete, { 'attr': { 'class': 'btn btn-danger' } }) }}
-        </div>
-    </section>
-    {{ form_end(delete_form) }}
+{% block javascripts %}
+    {% javascripts
+    '@EzPlatformAdminUiBundle/Resources/public/js/scripts/admin.policy.list.js'
+    '@EzPlatformAdminUiBundle/Resources/public/js/scripts/admin.role_assignment.list.js'
```
sunpietro: Is there a possibility that these files will be included separately into the page layout?
I mean, maybe we should avoid JS code duplication here and create one file with 2 selectors inside?

src/bundle/Resources/public/js/scripts/admin.role_assignment.list.js
```diff
@@ -0,0 +1,10 @@
+(function () {
+    const checkboxes = [...document.querySelectorAll('form[name="role_assignments_delete"] input[type="checkbox"]')];
+    const buttonRemove = document.querySelector('#delete-role-assignments');
+    const toggleButtonState = (event) => {
+        const methodName = checkboxes.some(el => el.checked) ? 'removeAttribute' : 'setAttribute';
+        buttonRemove[methodName]('disabled', true);
+    }
+
+    checkboxes.forEach(checkbox => checkbox.addEventListener('change', toggleButtonState, false));
```
sunpietro: The whole code here duplicates the code from `admin.policy.list.js` we can avoid it, by using another selector in the file above.

dew326:
ping @sunpietro

lserwatka:
Rebase is needed here.

PR #158, opened 2017-12-12
EZP-28360: Design consistency for tables across the whole application

inakijv:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-28360](https://jira.ez.no/browse/EZP-28360)
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

Most important aspects:
- Standardized style of `.ez-table-header` throughout the application (based also on HTML standards), as well as provided style for interactive icon buttons allocated within it;
- Updated style of Dashboard tables based on design specifications;
- Update design view for Admin > Content types > Specific Content type view based on specifications.

screen shot 2017-12-12 at 2 11 03 am
screen shot 2017-12-12 at 2 11 33 am
screen shot 2017-12-12 at 2 12 00 am
screen shot 2017-12-12 at 2 15 21 am
screen shot 2017-12-12 at 2 38 41 am

Next steps:
- All buttons inside `.ez-table-header` should be styled as icon buttons;
- In tables, delete (icon) button should be placed within `.ez-table-header` too;
- Add hover states to icons within rows (Edit & Assign), as well as align them based on design specifications.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_tables.scss
+++ b/src/bundle/Resources/public/scss/_tables.scss
@@ -5,30 +5,15 @@
     justify-content: space-between;
     background-color: $ez-ground-primary;
     align-items: center;
-    padding: .5rem 1rem;
+    padding: 0 1.25rem;
+    min-height: 50px;
+    border-radius: .25rem .25rem 0 0;
 
-    h5 {
+    &__headline {
+        font-size: 1.0625rem;
         margin-bottom: 0;
         font-weight: bold;
         color: $ez-black;
-    }
-
-    .ez-icon {
-        width: 1rem;
-        height: 1rem;
-        fill: $white;
-    }
-
-    &.round {
-        border-radius: 4px 4px 0 0;
-    }
-
-    &--label {
-        margin: 0;
-        padding: 0 .4375rem;
-        font-size: 1rem;
-        line-height: 2;
-        font-weight: bold;
     }
 }
 
@@ -36,22 +21,41 @@
     background-color: $ez-color-base-pale;
 }
 
+.table {
+    margin-bottom: 3rem;
+    
+    th, td {
+        padding: .7rem 1.4rem;
+        line-height: 20px;
+    }
+
+    thead th {
+        font-size: .9375rem;
+        vertical-align: top;
+    }
+
+    td {
+        font-size: .875rem;
+        vertical-align: middle;
+
+        .checkbox label {
+            margin-bottom: 0;
+        }
+    }
+}
+
 .ez-dashboard {
     .table {
+        margin-bottom: 0;
+
         thead {
-            background-color: $ez-ground-primary-inverted-light;
-            color: $white;
+            background-color: $ez-ground-primary;
+            color: $ez-black;
 
             th {
                 border-bottom: none;
             }
         }
-    }
-}
-
-.ez-content-view {
-    .table {
-        margin-bottom: 3rem;
     }
 }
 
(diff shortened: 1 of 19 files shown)


src/bundle/Resources/public/scss/_tables.scss
```diff
@@ -5,42 +5,65 @@
     justify-content: space-between;
     background-color: $ez-ground-primary;
     align-items: center;
-    padding: .5rem 1rem;
+    padding: 0 1.25rem;
+    height: 50px;
```
dew326: I'm not sure if we want to have fixed height of the header, what if the text will have 2 lines? I think `min-height` would be better.

src/bundle/Resources/public/scss/_tables.scss
```diff
@@ -5,42 +5,65 @@
...
+    border-radius: .25rem .25rem 0 0;
 
-    h5 {
+    p {
```
dew326: I know you just change `h5` to `p` but we should avoid tag selectors. Could you change it to class: `ez-table-header__text`?

src/bundle/Resources/public/scss/_tables.scss
```diff
@@ -5,42 +5,65 @@
...
-        width: 1rem;
-        height: 1rem;
-        fill: $white;
-    }
-
-    &.round {
-        border-radius: 4px 4px 0 0;
-    }
+    .btn {
+        line-height: .85;
```
dew326: The style for `.btn` should be in `_buttons` with parent `.ez-table-header`

src/bundle/Resources/public/scss/_tables.scss
```diff
@@ -5,42 +5,65 @@
...
+        margin-left: .15rem;
+        margin-right: 0;
 
-    &--label {
-        margin: 0;
-        padding: 0 .4375rem;
-        font-size: 1rem;
-        line-height: 2;
-        font-weight: bold;
+        .ez-icon {
```
dew326: And this should be in `_icons` also with parents.

src/bundle/Resources/public/scss/_tables.scss
```diff
@@ -64,9 +81,9 @@
 
 .ez-info-view {
     .table {
-        th, td {
-            padding: .8rem 1.4rem;
-        }
+        // th, td {
```
dew326: If it's not needed please remove

src/bundle/Resources/views/admin/content_type/list.html.twig
```diff
@@ -2,14 +2,14 @@
 
 <section class="container mt-4 px-5">
     <div class="ez-table-header">
-        <h4>{{ 'content_type.view.list.title'|trans({ '%identifier%': content_type_group.identifier })|desc('Content Types in %identifier%') }}</h4>
+        <p>{{ 'content_type.view.list.title'|trans({ '%identifier%': content_type_group.identifier })|desc('Content Types in %identifier%') }}</p>
```
sunpietro: If it's a `p` tag doesn't it mean its parent container is not a header-like element anymore, does it?
Maybe we should revise the parent container class name?

src/bundle/Resources/public/scss/_tables.scss
sunpietro: I would name it to: `ez-table-header__headline`

inakijv:
Updated based on comments.

PR #160, opened 2017-12-12
EZP-28415: [MapLocation] No validation of value range for longitude and latitude fields

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28415
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes/no
| Doc needed?   | yes/no

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezgmaplocation.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezgmaplocation.js
@@ -16,14 +16,25 @@
         validateCoordInput(input) {
             const value = parseFloat(input.value.replace(',', '.'));
             const result = { isError: false };
+            const label = input.closest('.ez-data-source__field').querySelector('.ez-data-source__label').innerHTML;
+            const isNumber = !isNaN(value);
+            const isCorrectValue = isNumber && (value <= 90 && value >= -90);
 
-            if (!input.required) {
+            if (!input.required && isCorrectValue) {
+                return result;
+            }
+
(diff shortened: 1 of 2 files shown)


mnocon:
One small thing fore **Firefox 57.0.1** (does not occur on Chrome):
There is a difference when entering characters into field that was marked as incorrect compared to the one that had correct input.

**Example 1 (with correct input at first):**
1. Start creating a Content Item with Map field type
2. Enter a valid value, for example "20"
3. Select all the text that has been entered and enter "-40"

Result: "-40" is in the field

**Example (with incorrect input at first):**
1. Start creating a Content Item with Map field type
2. Enter an invalid value into Latitude, for example "-100"
3. Select all the text that has been entered and enter "-40"

Result: "40" is in the field, it should be "-40" (the minus sign is lost)

You should be able to see an example here: https://ezno-my.sharepoint.com/personal/marek_nocon_ez_no/_layouts/15/guestaccess.aspx?guestaccesstoken=VptD2nGH3WkXB6CwfKGCmZfzX7mVYiRO%2B5Vz04X91rk%3D&docid=2_0c7c4817561f24677bc861f55f953ddef&rev=1&e=1f2371431b1448ba835a3309f2fb0111

sunpietro:
@mnocon I'm not able to reproduce the issue you mentioned. If such behaviour occurs then obviously it has to be something with input element implementation in your version of Firefox. As I said, I've just checked that issue and it doesn't occur for me.

PR #170, opened 2017-12-13
 EZP-28493: Interface improvements when assigning roles to user / groups

webhdx:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28493
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Screenshot:
roles

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.role_assignment.add.js
+++ b/src/bundle/Resources/public/js/scripts/admin.role_assignment.add.js
@@ -1,25 +1,38 @@
-(function () {
-    const btns = document.querySelectorAll('.btn--open-udw');
-    const udwContainer = document.getElementById('react-udw');
-    const token = document.querySelector('meta[name="CSRF-Token"]').content;
-    const siteaccess = document.querySelector('meta[name="SiteAccess"]').content;
+(function (global, doc) {
+    const udwContainer = doc.getElementById('react-udw');
+    const limitationsRadio = [...doc.querySelectorAll('.ez-limitations__radio')];
+    const token = doc.querySelector('meta[name="CSRF-Token"]').content;
+    const siteaccess = doc.querySelector('meta[name="SiteAccess"]').content;
     const closeUDW = () => udwContainer.innerHTML = '';
-    const onConfirm = (form, content) => {
-        const field = form.querySelector('#role_assignment_locations_value');
-        field.value = content.map(item => item.ContentInfo.Content._id).join();
+    const selectSubtreeConfirm = (data) => {
+        const selectedItems = data.reduce((total, item) => total + `<li>${item.ContentInfo.Content.Name}</li>`, '');
+
+        doc.querySelector('#role_assignment_create_locations').value = data.map(item => item.id).join();
+        doc.querySelector('.ez-limitations__selected-subtree').innerHTML = selectedItems;
+
         closeUDW();
     };
-    const onCancel = () => closeUDW();
-    const openUDW = (event) => {
+    const selectSubtree = (event) => {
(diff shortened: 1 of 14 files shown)


lserwatka:
@micszo we need very careful QA on this one, including DB entires.

lserwatka:
@webhdx could you attach screenshot?

src/bundle/Resources/public/js/scripts/admin.role_assignment.add.js
```diff
@@ -1,25 +1,38 @@
...
         ReactDOM.render(React.createElement(eZ.modules.UniversalDiscovery, {
-            onConfirm: onConfirm.bind(this, form),
-            onCancel: onCancel,
+            onConfirm: selectSubtreeConfirm.bind(this),
+            onCancel: closeUDW,
+            multiple: true,
             restInfo: {token, siteaccess}
         }), udwContainer);
     };
+    const toggleDisableState = () => {
```
sunpietro: `toggleDisabledState`

src/bundle/Resources/public/js/scripts/admin.role_assignment.add.js
```diff
@@ -1,25 +1,38 @@
...
+        limitationsRadio.forEach(radio => {
+            const disableNode = doc.querySelector(radio.dataset.disableSelector);
```
sunpietro: Do you have to query the whole document?

src/bundle/Resources/public/js/scripts/admin.role_assignment.add.js
```diff
@@ -1,25 +1,38 @@
...
+            if (disableNode) {
+                disableNode[methodName]('disabled', 'disabled');
+            }
+        });
+    };
 
-    btns.forEach(btn => btn.addEventListener('click', openUDW, false));
-})();
+    doc.querySelector('.ez-btn--select-subtree').addEventListener('click', selectSubtree, false);
+    limitationsRadio.forEach(radio => radio.addEventListener('change', toggleDisableState, false));
```
sunpietro: I'm not sure the use case here. If you're clicking on a specific radio button then I would expect to change the state of a clicked button. The `toggleDisabledState` function tries to toggle the state of all radio buttons. It looks quite suboptimal.

src/bundle/Resources/views/admin/role_assignment/create.html.twig
```diff
@@ -38,10 +38,39 @@
                 <div class="ez-card__header ez-card__header--secondary">
                     {{ 'role_assignment.view.add.panel.limitations.title'|trans|desc('Limitations') }}
                 </div>
-                <div class="card-body">
-                    {{ form_row(form.sections) }}
-                    {{ form_widget(form.locations.select_content, {'attr': {'class': 'btn btn-secondary btn--open-udw'}}) }}
-                    {% do form.locations.setRendered %}
+                <div class="card-body ez-limitations">
+                    <div>
```
sunpietro: Do you need the `div` wrapper here?

src/bundle/Resources/public/js/scripts/admin.role_assignment.add.js
dew326: The event change is fired only on radio which gets the checked=true. I have to loop over every to set disable state on other buttons

lserwatka:
@mikadamczyk it looks like you need to run cs-fixer here

PR #185, opened 2017-12-15
EZP-28547: SubItems should not redirect to form on edit item

dew326:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28547
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

**Description**
Fixed bug with handle submits form with a different name than the default. When the name of the form is set to some custom the submit creates a form with the wrong name. In my case I set name to `form_sub_items_content_edit` and the submit creates form with name `content_draft_create`

**Requires:** https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/28

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/ContentController.php
+++ b/src/bundle/Controller/ContentController.php
@@ -126,7 +126,11 @@ class ContentController extends Controller
      */
     public function editAction(Request $request): Response
     {
-        $form = $this->formFactory->contentEdit();
+        /* @todo it shouldn't rely on keys from request */
+        $requestKeys = $request->request->keys();
+        $formName = reset($requestKeys) ?: null;
+
+        $form = $this->formFactory->contentEdit(null, $formName);
         $form->handleRequest($request);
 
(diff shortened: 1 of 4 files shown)


src/bundle/Controller/ContentController.php
```diff
@@ -126,7 +126,7 @@ public function createAction(Request $request): Response
      */
     public function createDraftAction(Request $request): Response
     {
-        $form = $this->formFactory->createContentDraft();
+        $form = $this->formFactory->createContentDraft(null, $request->request->keys()[0]);
```
dew326: Probably not the best solution, backend guys should find better ;)

lserwatka:
@webhdx @Nattfarinn do you have better solution? We need to merge this one ahead of RC1

PR #199, opened 2017-12-18
EZP-28559: Fixed block names in twig templates to snake_case

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28559
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   |no

Fixed block names in twig templates to snake_case

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/Security/login.html.twig
+++ b/src/bundle/Resources/views/Security/login.html.twig
@@ -2,7 +2,7 @@
 
 {% block navigation %}{% endblock %}
 
-{% block bodyClass %}ez-login-view{% endblock %}
+{% block body_class %}ez-login-view{% endblock %}
 
 {% block content %}
     <div class="row ez-login">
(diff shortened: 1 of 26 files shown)


src/bundle/Resources/views/parts/menu/top_menu_2nd_level.html.twig
```diff
@@ -34,6 +34,6 @@
     </a>
 {% endblock %}
 
-{% block spanElement %}
-    {{ block('linkElement') }}
+{% block span_element %}
```
dew326: Isn't that block from KNP? Can we change this? /cc @webhdx

webhdx: I don't think we can change it. This overrides the block from KNP.

mikadamczyk: Ok, I fixed this one @dew326 @webhdx

PR #214, opened 2017-12-18
Missed reference to location_ids.content (EZP-28553)

wizhippo:
EZP-28553: Adds Permission Awareness to all menus

| Question      | Answer
| ------------- | ---
| Tickets       | Related to EZP-28553
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/config/ezplatform_default_settings.yml
+++ b/src/bundle/Resources/config/ezplatform_default_settings.yml
@@ -1,6 +1,5 @@
 parameters:
     # System Location IDs
-    ezsettings.default.location_ids.content: 2
     ezsettings.default.location_ids.media: 43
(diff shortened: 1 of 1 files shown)


emodric:
Why are these IDs (for media and users section) even defined here? Shouldn't they be taken from eZ kernel?

webhdx:
@emodric These settings are not available in kernel.

emodric:
My bad, sorry :)

andrerom:
yup, they probably should be :) _(but then installer will need to modify them depending on installed demo content, in case of future multi page demo)_

PR #226, opened 2017-12-19
EZP-28598: Fixes no filtering langauge on create content UI

webhdx:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28598
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/config/services.yml
+++ b/src/bundle/Resources/config/services.yml
@@ -4,13 +4,13 @@ imports:
     - { resource: services/tabs.yml }
     - { resource: services/menu.yml }
     - { resource: services/application_config.yml }
-    - { resource: services/type_extensions.yml }
     - { resource: services/pagination.yml }
     - { resource: services/ui_config.yml }
     - { resource: services/components.yml }
     - { resource: services/dashboard.yml }
     - { resource: services/modules/subitems.yml }
     - { resource: services/form_processors.yml }
+    - { resource: services/form_types.yml }
 
 parameters:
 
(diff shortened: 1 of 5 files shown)


micszo:
@webhdx it is intentional that error 500 is thrown when the two `languages` arrays in ezplatform.yml are not identical?
screen shot 2017-12-19 at 10 35 09
(one is in system: site_group: and the other in system: admin_group:)

micszo:
Also on this branch with a testeditor user I cannot enter Content / Content structure with a set o policies that allow that on _master_.
screen shot 2017-12-19 at 11 06 08
The policies are:
screen shot 2017-12-19 at 11 08 13

micszo:
@webhdx could you confirm the necessary policies for a editor limited to one language?

lserwatka:
@webhdx rebase is needed

micszo:
@mnocon we reached a deadlock, please help. I still have the second error which Maciek cannot reproduce.

micszo:
List of policies from Maciek:
image

mnocon:
@micszo @webhdx  looks like it works for me - I have created a French language, created a user that is assigned to the Role specified above and I can browse Content -> Content Structure on ezplatform-ee (dev mode). We can go through that tomorrow to double check everything.

webhdx:
Ok, I managed to reproduce the issue. It's caused by `PermissionResolver`:
1. In `canUser()` it overrides `$targets = null` when `$targets` is empty. Later when it checks `SubtreeLimitation`  it expects `$targets` to be an array.

cc @alongosz

webhdx:
Further investigation: Whole PR won't work:
In order to check permissions I have to create `ContentCreateStruct` simulating content user is going to create. In case we are not able to provide ContentType information which is required to perform Permissions check.

For the time being we can only limit language list depending on siteaccess defined languages.

micszo:
Follow-up improvement https://jira.ez.no/browse/EZP-28620.

PR #230, opened 2017-12-19
EZP-28604: Menu is not collapsed when switching tab

dew326:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28604
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_main-nav.scss
+++ b/src/bundle/Resources/public/scss/_main-nav.scss
@@ -7,7 +7,7 @@
     > .container-fluid {
         padding-right: 0;
     }
-    
+
     .navbar-brand {
         margin-top: .3125rem;
     }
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/public/scss/_main-nav.scss
```diff
@@ -52,10 +52,11 @@
             padding: .625rem 1rem;
             color: $ez-black;
 
-            &.active {
+            &.ez-active {
```
sunpietro: Does the class renaming do a thing? Strange without having a broader context.

dew326: It's conflict with naming. The tab is looking for `.active` to hide the submenu but it finds the `Content structure` menu item because it also has class `.active`.

sunpietro: Is it because of Bootstrap scripts?

dew326: Indeed

PR #234, opened 2017-12-19
EZP-28608: Add pagination to the Sections list

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28608
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

Added pagination to the Sections list

screen shot 2017-12-19 at 3 50 45 pm

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/config/ezplatform_default_settings.yml
+++ b/src/bundle/Resources/config/ezplatform_default_settings.yml
@@ -4,8 +4,9 @@ parameters:
     ezsettings.default.location_ids.users: 5
 
     # Pagination limits
-    ezsettings.default.search_pagination_limit: 10
-    ezsettings.default.trash_pagination_limit: 10
+    ezsettings.default.pagination.search_limit: 10
+    ezsettings.default.pagination.trash_limit: 10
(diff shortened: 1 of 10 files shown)


src/bundle/Resources/config/ezplatform_default_settings.yml
```diff
@@ -6,6 +6,7 @@ parameters:
     # Pagination limits
     ezsettings.default.search_pagination_limit: 10
     ezsettings.default.trash_pagination_limit: 10
+    ezsettings.default.section_pagination_limit: 10
```
webhdx: I'm not convinced with having different limit setting for every page? Why do we need to complicate it so much?

More important thing: you are defining parameters here but you haven't provided config parser for them which means it isn't possible to override those with config.yml. Please reference what I did with content ids.

lserwatka: It is more flexibility doing this way, but config should be validated.

PR #240, opened 2017-12-19
EZP-28612: Session is always started for anonymous user

Nattfarinn:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28612
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/config/services.yml
+++ b/src/bundle/Resources/config/services.yml
@@ -64,6 +64,7 @@ services:
 
 
     EzSystems\EzPlatformAdminUi\Notification\FlashBagNotificationHandler:
+        lazy: true
         arguments:
             $flashBag: "@=service('session').getFlashBag()"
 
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/config/services.yml
```diff
@@ -64,6 +64,7 @@ services:
 
 
     EzSystems\EzPlatformAdminUi\Notification\FlashBagNotificationHandler:
+        lazy: true
```
emodric: Would it be better to refactor `FlashBagNotificationHandler` to look something like https://github.com/ezsystems/ezplatform-admin-ui/pull/241, rather than relying on the service being lazy?

Nattfarinn:
Closing in favour of: https://github.com/ezsystems/ezplatform-admin-ui/pull/241

PR #241, opened 2017-12-19
Fix EZP-28612: Session is always started for anonymous user

emodric:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28612
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

This refactors a bit `FlashBagNotificationHandler` in order not to start the session for anonymous user, by calling the `getFlashBag` only when neeeded, since it is responsible for starting the session.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/config/services.yml
+++ b/src/bundle/Resources/config/services.yml
@@ -63,9 +63,7 @@ services:
     EzSystems\EzPlatformAdminUi\Form\Factory\FormFactory: ~
 
 
-    EzSystems\EzPlatformAdminUi\Notification\FlashBagNotificationHandler:
-        arguments:
-            $flashBag: "@=service('session').getFlashBag()"
+    EzSystems\EzPlatformAdminUi\Notification\FlashBagNotificationHandler: ~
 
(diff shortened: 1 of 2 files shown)


andrerom:
@Nattfarinn Guess you forgot to mention on slack you where going to look at it? Ref #240

emodric:
@andrerom @Nattfarinn Damn :)

andrerom:
👍

emodric:
Nice! :)

What about template change in #240 about using `{% if app.request.hasPreviousSession %}` ? Is that needed?

Nattfarinn:
@emodric Even if it would start, it's not required. This "if" is a good practice to avoid anonymous session starts, but in this case it's used inside Admin panel so you do have session anyway. :)

emodric:
@Nattfarinn Well, not always. That layout template is also used as a layout for the login page, so just by displaying a login page, session would start if we were on lower versions of Symfony :)

PR #244, opened 2017-12-20
EZP-28616: Add pagination to the Content Type Groups and Content Type

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28616
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

Added pagination to the Content Type Groups and Content Type

screen shot 2017-12-20 at 12 22 40 pm
screen shot 2017-12-20 at 12 22 28 pm

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/ContentTypeController.php
+++ b/src/bundle/Controller/ContentTypeController.php
@@ -19,6 +19,8 @@ use EzSystems\EzPlatformAdminUi\Notification\NotificationHandlerInterface;
 use EzSystems\RepositoryForms\Data\Mapper\ContentTypeDraftMapper;
 use EzSystems\RepositoryForms\Form\ActionDispatcher\ActionDispatcherInterface;
 use EzSystems\RepositoryForms\Form\Type\ContentType\ContentTypeUpdateType;
+use Pagerfanta\Adapter\ArrayAdapter;
+use Pagerfanta\Pagerfanta;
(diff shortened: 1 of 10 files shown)


src/bundle/Controller/ContentTypeController.php
```diff
@@ -71,22 +82,43 @@ public function __construct(
...
+     * @return Response
+     */
     public function listAction(ContentTypeGroup $group): Response
     {
+        $parentRequest = $this->requestStack->getParentRequest();
+
         $deletableTypes = [];
 
-        $types = $this->contentTypeService->loadContentTypes($group, $this->languages);
+        $page = $parentRequest->query->get('page') ?? 1;
```
webhdx: Why can't you just define `$page` as a part of the route so you don't have to access `RequestStack` as this feels dirty, really dirty?

mikadamczyk: It's not only about page. I need also route name from parent request. I think this is a better approach than pass this variables in `    {{ render(controller('EzPlatformAdminUiBundle:ContentType:list', {
        contentTypeGroupId: content_type_group.id
    })) }}`

Or maybe this should be standalone action instead render controller?

PR #250, opened 2017-12-21
EZP-28634: Top menu disappear on Search and Trash

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28634
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   | no

Top menu disappeared on Search and Trash. Current solution is to show top menu always with selected position "Content" -> "Content structure selected"

screen shot 2017-12-21 at 10 12 47 am
screen shot 2017-12-21 at 10 12 32 am

#### Checklist:
- [x] Ready for Code Review


mnocon:
@mikadamczyk not sure if it's related to this ticket, but this behaviour could not be present before this change:

when you go to "Content -> Media" and then click "Search" the menu switches to "Content -> Content structure). Same applies for "Trash".

Should it be that way?

mikadamczyk:
I think this is the best solution for that moment.

PR #287, opened 2017-12-22
EZP-28670: Certain routes conflict with eZ Publish legacy admin

emodric:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28670
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes?
| Tests pass?   | yes
| Doc needed?   | no

Using `/section/list`, `/section/view/{sectionId}` and `/role/list` conflicts with eZ Publish legacy admin routes, so this renames them in order not to conflict. The rest of the routes are renamed for consistency.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/config/routing.yml
+++ b/src/bundle/Resources/config/routing.yml
@@ -40,39 +40,39 @@ ezplatform.systeminfo.php:
 #
 
 ezplatform.section.list:
-    path: /section/list
+    path: /sections/list
     defaults:
         _controller: 'EzPlatformAdminUiBundle:Section:list'
 
 ezplatform.section.create:
-    path: /section/create
+    path: /sections/create
     defaults:
         _controller: 'EzPlatformAdminUiBundle:Section:create'
 
 ezplatform.section.view:
-    path: /section/view/{sectionId}
+    path: /sections/view/{sectionId}
     defaults:
         _controller: 'EzPlatformAdminUiBundle:Section:view'
 
 ezplatform.section.update:
-    path: /section/update/{sectionId}
+    path: /sections/update/{sectionId}
     defaults:
         sectionId: null
         _controller: 'EzPlatformAdminUiBundle:Section:update'
 
 ezplatform.section.delete:
-    path: /section/delete/{sectionId}
+    path: /sections/delete/{sectionId}
     defaults:
         _controller: 'EzPlatformAdminUiBundle:Section:delete'
 
 ezplatform.section.bulk_delete:
-    path: /section/bulk-delete
+    path: /sections/bulk-delete
     methods: ['POST']
     defaults:
         _controller: 'EzPlatformAdminUiBundle:Section:bulkDelete'
 
 ezplatform.section.assign_content:
-    path: /section/assign-content/{sectionId}
+    path: /sections/assign-content/{sectionId}
     defaults:
         _controller: 'EzPlatformAdminUiBundle:Section:assignContent'
 
@@ -117,13 +117,13 @@ ezplatform.language.bulk_delete:
 #
 
 ezplatform.role.list:
-    path: /role/list
+    path: /roles/list
     methods: ['GET']
     defaults:
         _controller: 'EzPlatformAdminUiBundle:Role:list'
 
 ezplatform.role.view:
-    path: /role/{roleId}/{policyPage}/{assignmentPage}
+    path: /roles/{roleId}/{policyPage}/{assignmentPage}
     methods: ['GET']
     defaults:
         _controller: 'EzPlatformAdminUiBundle:Role:view'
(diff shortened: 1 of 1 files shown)


emodric:
@lserwatka Could this land into final? This is a pretty big conflict with Netgen Admin UI.

lserwatka:
I'm afraid we are too late in the process.

emodric:
Okay. I've overriden them in Netgen Admin UI to make it work: https://github.com/netgen/NetgenAdminUIBundle/commit/b52353c5f078a6132d3fc7fa79a27a88794ab8f6

lserwatka:
We won't integrate this change into the admin UI. It is not our intention to support legacy setups working in a new stack mode.

emodric:
I don't think that's a decision you might make lightly. This is not a case of major incompatiblity which would prevent Netgen Admin UI from working. If it were, I would understand. But this is only a case of a conflict between the routes which is easily solvable by modifying the route paths in order not to conflict with eZ legacy admin routes.

I easily hacked the route in Netgen Admin UI:

https://github.com/netgen/NetgenAdminUIBundle/blob/master/Resources/config/routing.yml#L13-L29

but this will only complicate maintenance both for me (where I would need to keep a close eye on your routes in case they change), and for you (where you will get users which will have Netgen Admin UI installed and will try to ask help for eZ Platform Admin UI). So, I don't think it is in anyones interest to have this hack out there.

lserwatka:
Our intention is to move forward and not carry on with legacy compatibilities besides database (what was initially communicated) which might impact development of new generation Admin UI and Studio. This include not supporting usage of legacy parts in Symfony mode, which is what Netgen Admin UI does, and why you have issues with routes in symfony overloading legacy routes. Reg. to https://doc.ez.no/display/EZP/Legacy+code+and+features#Legacycodeandfeatures-Whatitdoes

We can organise a meeting inc. our PM team to discuss it. This is not the best place.

emodric:
I understand that you don't want to support it and that you wish to discourage it and that is certainly fine, but why would you actively prevent such usecase is beyond me, especially since the fix is contributed (so you don't have to write it) and is so simple, obvious and backwards compatible.

I only hope that waiting for decision from your PM team will not take forever.

lserwatka:
Because this PR "cripples" routes names, by changing them from singular to plural, just to support legacy system. They are like that on purpose. This is not how good engineering should be handled. Change like this will also involve high QA cost on our side, even if change looks simple at the first sight.

emodric:
Okay, so that issue could've been comunicated from the start so I can find a better solution then for the route paths, instead of closing the PR. They don't have to be plural, it was just a most obvious suggestion :)

lserwatka:
Perhaps I was not clear explaining reasons why this PR was closed, thus I will repeat it again:
1. We don't have intention to support legacy parts working in Symfony mode
2. PR was technically against our principles for naming routes
3. We don't have intention to change properly defined routes which were already tested

I hope this clear now. Thank you.

crevillo:
sorry to enter here. but about point 1 of your last comment
> We don't have intention to support legacy parts working in Symfony mode

To me this is contrary to what we can read here https://ez.no/Blog/eZ-Platform-Enterprise-Now-Officially-Supports-Use-with-Legacy-Bridge. 
Or at least, contrary to what I understood when i read that.

lserwatka:
@crevillo no it is not, we will support Legacy Bridge as announced. It is written in the announcement "There are technical limitations (ex: Field Types), and it can be highly confusing for editors, so in the end you need to avoid using both at the same time", this implies for using both admin panels at the same time. @andrerom might explain this in more details.

andrerom:
@crevillo It's a bit different, the blog post goes into being able to run legacy in legacy mode, including legacy site admin. Also being able to have fallbacks in code from symfony stack to legacy stack to reuse code and logic, including being able to have legacy modules still working.

This is the same as what was supported in 5.x.

However we never promised we would avoid overlap on routes in platform/symfony mode, that is somewhat outside of the scope we have aimed for. Evne in 5.x that might have happened if you added thirdparty bundles for instance that might pick similar routes overlapping.

So far on cases where we have managed to limit the reach of our routes by narrowing the match to avoid conflicts with legacy, without compromising our route layout, we have done it. In this case @lserwatka and team have decided against the proposed solution, but there is still option of finding another one.

crevillo:
ok. good to know. i know you have never promised such thing, but in the post i can read 
`we are happy to announce the roll out of **full** support for eZ Publish legacy and legacy bridge`

Full is what confuses me here.

andrerom:
True, well there are some limitations to it that could have been written more clear:
1. The sentences talks about migration scenarios, this is what we professionally support. If people want to do clean install on legacy, that is perfectly fine, but not something we spend efforts on so that is something you would typically get help on from community.
2. Full here implies, as full as in 5.x, so as full as before. With the exception of scope mentioned in 1.

emodric:
> Evne in 5.x that might have happened if you added thirdparty bundles for instance that might pick similar routes overlapping.

If that would've happened, we'd do a PR over there too, to avoid the overlap, just as it was done here, because I still don't think it is in anyones interest to conflict over simple matters like route paths.

> However we never promised we would avoid overlap on routes in platform/symfony mode

@andrerom Your blog post doesn't mention the difference between legacy mode true/false too: that one is being supported and the other not. To me, full implies as full as what is possible to do with configuration, and it's certainly possible to change the siteaccess to legacy mode false.

You say you want to only support migration scenarios. Don't you think that having Netgen Admin UI working on eZ Platform in legacy mode false (meaning, having legacy admin with the possibility of including Symfony based solutions like Netgen Tags or custom customer code written in Symfony) is a valid migration scenario?

PR #296, opened 2017-12-31
As an Editor I want language choices to be ordered by SiteAccess settings

andrerom:
| Question      | Answer
| ------------- | ---
| Tickets       | 
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | n/a
| Doc needed?   | unsure

E.g. If I configure admin to be [nor-NO, eng-GB], then I expect norwegian to be the default
choice so I don't need to click to change it when I create new content.

Review note:
- I'm unsure if this is right layer to do this, but until we move forward on SiteAccess aware
  Repo PR, this is currently only place we can do it.
- As class is autowired with no service config, I opted for type hinting for full config resolver
- EDIT: And besides, we need to filter out disabled languages here anyway

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/Form/Type/Language/LanguageChoiceType.php
+++ b/src/lib/Form/Type/Language/LanguageChoiceType.php
@@ -8,7 +8,6 @@ declare(strict_types=1);
 
 namespace EzSystems\EzPlatformAdminUi\Form\Type\Language;
 
-use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\ConfigResolver;
 use eZ\Publish\API\Repository\LanguageService;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\ChoiceList\Loader\CallbackChoiceLoader;
@@ -25,16 +24,17 @@ class LanguageChoiceType extends AbstractType
     /** @var LanguageService */
     protected $languageService;
 
-    /** @var ConfigResolver */
-    protected $configResolver;
+    /** @var array */
+    protected $siteAccessLanguages;
 
     /**
      * @param LanguageService $languageService
+     * @param array $siteAccessLanguages
      */
-    public function __construct(LanguageService $languageService, ConfigResolver $configResolver)
+    public function __construct(LanguageService $languageService, array $siteAccessLanguages)
     {
         $this->languageService = $languageService;
-        $this->configResolver = $configResolver;
+        $this->siteAccessLanguages = $siteAccessLanguages;
     }
 
     public function getParent()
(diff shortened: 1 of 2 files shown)


lserwatka:
@andrerom could you rebase?
@webhdx could you make code review?
@micszo we will need a round of QA here.

andrerom:
@lserwatka rebased

webhdx:
Hey @andrerom. Took me a while to figure it out what is happening here. Won't using `array_filter` and `array_multisort` be simpler and more logical?

src/lib/Form/Type/Language/LanguageChoiceType.php
```diff
@@ -39,9 +44,31 @@ public function getParent()
 
     public function configureOptions(OptionsResolver $resolver)
     {
+        $languageService = $this->languageService;
+        $siteAccessLanguages = $this->configResolver->getParameter('languages');
         $resolver
             ->setDefaults([
-                'choice_loader' => new CallbackChoiceLoader([$this->languageService, 'loadLanguages']),
+                'choice_loader' => new CallbackChoiceLoader(function () use ($languageService, $siteAccessLanguages) {
```
Nattfarinn: As far as I understand a logic behind this, it's a bit overcomplicated. 

How about this?

```php
'choice_loader' => new CallbackChoiceLoader(function() {
    $languages = [];
    $languagesByCode = [];
    
    foreach ($this->languageService->loadLanguages() as $language) {
        $languagesByCode[$language->languageCode] = $language;
    }

    foreach ($this->siteAccessLanguages as $language) {
        if (!isset($languagesByCode[$language]) || !$languagesByCode[$language]->enabled) {
            continue;
        }

        $languages[] = $languagesByCode[$language];
        unset($languagesByCode[$language]);
    }

    return $languages + $languagesByCode;
}),
```

src/lib/Form/Type/Language/LanguageChoiceType.php
```diff
@@ -24,12 +25,16 @@ class LanguageChoiceType extends AbstractType
...
+    protected $configResolver;
+
     /**
      * @param LanguageService $languageService
      */
-    public function __construct(LanguageService $languageService)
+    public function __construct(LanguageService $languageService, ConfigResolver $configResolver)
     {
         $this->languageService = $languageService;
+        $this->configResolver = $configResolver;
```
Nattfarinn: No need for `ConfigResolver` here, why don't just inject `array $siteAccessLanguages`?

```yml
EzSystems\EzPlatformAdminUi\Form\Type\Language\LanguageChoiceType:
    arguments:
        $siteAccessLanguages: '$languages$'
```

andrerom: I can, but as service was autowired I did not want to having to define it.

Also question is if we allow siteaccess switching at some point (several admins), then it might be worthwhile to get it dynamically via ConfigResolver on-demand.

But that said, let's define it for now, can always come back and change it if we need to.

src/lib/Form/Type/Language/LanguageChoiceType.php
andrerom: Can do it like that, but would probably have to change your last line to something like:
`return array_merge($languages, array_values($languagesByCode));`

ok?

Nattfarinn: @andrerom Sure. :)

andrerom:
@Nattfarinn updated

PR #307, opened 2018-01-15
EZP-28742: Add unit tests to RepositoryForms

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28742
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

Add unit tests to RepositoryForms

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/Tests/RepositoryForms/Form/Processor/PreviewFormProcessorTest.php
+++ b/src/lib/Tests/RepositoryForms/Form/Processor/PreviewFormProcessorTest.php
@@ -1,0 +1,288 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformAdminUi\RepositoryForms\Form\Processor;
+
+use eZ\Publish\API\Repository\Values\Content\ContentInfo;
+use eZ\Publish\Core\Repository\Values\Content\VersionInfo;
+use EzSystems\RepositoryForms\Data\Content\FieldData;
+use PHPUnit\Framework\TestCase;
+use eZ\Publish\API\Repository\ContentService;
+use EzSystems\EzPlatformAdminUi\Form\Event\ContentEditEvents;
+use EzSystems\EzPlatformAdminUi\Notification\NotificationHandlerInterface;
+use EzSystems\RepositoryForms\Data\Content\ContentCreateData;
+use EzSystems\RepositoryForms\Event\FormActionEvent;
+use Symfony\Component\Form\FormConfigInterface;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Symfony\Component\Translation\TranslatorInterface;
+use Symfony\Component\Form\FormInterface;
+use eZ\Publish\Core\Repository\Values\ContentType\FieldDefinition;
+use eZ\Publish\API\Repository\Values\Content\LocationCreateStruct;
+use eZ\Publish\Core\Repository\Values\Content\Content;
+use eZ\Publish\API\Repository\Values\Content\Content as APIContent;
+use PHPUnit\Framework\MockObject\MockObject;
+use eZ\Publish\Core\Repository\Values\ContentType\ContentType;
+
+class PreviewFormProcessorTest extends TestCase
+{
+    /** @var ContentService $contentService */
+    private $contentService;
+
+    /** @var UrlGeneratorInterface $urlGenerator */
+    private $urlGenerator;
+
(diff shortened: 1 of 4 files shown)


src/lib/Tests/RepositoryForms/Form/Processor/PreviewFormProcessorTest.php
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    public function testSubscribedEvents()
+    {
+        $previewFormProcessor = $this->createPreviewFormProcessor();
+
+        $this->assertSame([ContentEditEvents::CONTENT_PREVIEW => ['processPreview', 10]], $previewFormProcessor::getSubscribedEvents());
+    }
+
+    protected function createUntouchableMock(string $class): MockObject
```
webhdx: What's the thinking behind untouchable mock? Any particular reason why you are mocking a class which should never be used by implementation?

src/lib/Tests/RepositoryForms/Form/Processor/PreviewFormProcessorTest.php
```diff
@@ -0,0 +1,307 @@
...
+            ->willReturn('http://content_edit_url');
+
+        return $notificationHandler;
+    }
+
+    private function generateTranslatorMock()
+    {
+        $translator = $this->createMock(TranslatorInterface::class);
+        $translator
+            ->expects(self::once())
```
webhdx: I'm against creating strict mocks that are checking method calls - this is testing implementation detail which should not be in the scope of the test. From my perspective I don't see any point in mocking Translator interface and making sure it was called only once with specific message.

src/lib/Tests/RepositoryForms/Form/Processor/PreviewFormProcessorTest.php
mikadamczyk: Yes, if we have situation like in `ProcessPreview` method, where we want to be sure, that in "positive path" we won't call any method from `notificationHandler`or `translator`, this is usecase for "UntouchableMock". E.g. if someone accidentally delete try-catch block.

src/lib/Tests/RepositoryForms/Form/Processor/PreviewFormProcessorTest.php
mikadamczyk: I think it is important to ensure that we have the right error message, and we do it once. But I would like to hear another opinion @Nattfarinn @ViniTou

src/lib/Tests/RepositoryForms/Form/Processor/PreviewFormProcessorTest.php
webhdx: Sorry but you are restricting implementation details in the test cases which should NEVER be the case. Right now you are making any refactoring really hard because small change can make tests fail. How not using `$translator` is a valid assertion from the test standpoint?

src/lib/Tests/RepositoryForms/Form/Processor/PreviewFormProcessorTest.php
webhdx: Alright but using "expects only once" affects implementation details which in most cases ends up as failing tests even though the changes were correct. Please test only the outcome which in case of a PreviewFormProcessor might be not that explicit but it would be saving a draft if I remember correctly.

src/lib/Tests/RepositoryForms/Form/Processor/PreviewFormProcessorTest.php
```diff
@@ -0,0 +1,290 @@
...
+        $fieldDefinitionIdentifier = 'identifier_1';
+        $fieldDataValue = 'some_value';
+
+        /**
+         * $data variable in PreviewFormProcessor classs.
+         */
+        $contentStruct = $this->generateContentStruct($languageCode, $fieldDefinitionIdentifier, $fieldDataValue);
+
+        $contentDraft = $this->generateContentDraft($contentDraftId, $languageCode);
+
```
webhdx: Can you work on code style a bit more? I feel like separating everything with a new line doesn't make sense here.

src/lib/Tests/RepositoryForms/View/ViewParametersListenerTest.php
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * @param ContentInfo $contentInfo
+     *
+     * @return APIVersionInfo
+     */
+    private function generateVersionInfo(APIContentInfo $contentInfo): APIVersionInfo
+    {
+        return new VersionInfo(
+            [
```
webhdx: For arrays with only one element you can write it inline. No need to expand it as it takes more lines and forces you to scroll more.

lserwatka:
@webhdx are you good now?

PR #313, opened 2018-01-17
EZP-28686: Error 500 while user is going to dashboard after deleting another user

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28686
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   |no

Fixed error 500 while user is going to dashboard after deleting another user. After user deletion we will display text "Can not fetch contributor" in a "Contributor" cell.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/dashboard/tab/all_media.html.twig
+++ b/src/bundle/Resources/views/dashboard/tab/all_media.html.twig
@@ -17,8 +17,11 @@
                 <td><a href="{{ url('_ez_content_view', { 'contentId': row.contentId }) }}">{{ row.name }}</a></td>
                 <td>{{ row.type }}</td>
                 <td>
-                    {{ ez_render_field(row.contributor, 'first_name') }}
-                    {{ ez_render_field(row.contributor, 'last_name') }}
+                    {% if row.contributor is not null %}
+                        {{ row.contributor.name }}
+                    {% else %}
+                        {{ 'dashboard.table.contributor.not_found'|trans|desc('Can’t fetch contributor') }}
+                    {% endif %}
                 </td>
                 <td>{{ row.modified|date('M d, Y h:iA') }}</td>
                 <td class="text-center">
--- a/src/bundle/Resources/translations/dashboard.en.xliff
+++ b/src/bundle/Resources/translations/dashboard.en.xliff
@@ -56,6 +56,11 @@
         <target state="new">Contributor</target>
         <note>key: dashboard.table.contributor</note>
       </trans-unit>
+      <trans-unit id="e7790665f417ad7d9479f8b7c0920d212e4e24f6" resname="dashboard.table.contributor.not_found">
+        <source>Can’t fetch contributor</source>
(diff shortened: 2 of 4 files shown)


barbaragr:
I see "admin2" in Contributor cell instead of "Can not fetch contributor". Is it OK?

mikadamczyk:
For now, yes. It is a bug https://jira.ez.no/browse/EZP-28737. Your user is only partially deleted. I think we must wait.

src/bundle/Resources/views/dashboard/tab/all_media.html.twig
```diff
@@ -17,8 +17,12 @@
                 <td><a href="{{ url('_ez_content_view', { 'contentId': row.contentId }) }}">{{ row.name }}</a></td>
                 <td>{{ row.type }}</td>
                 <td>
-                    {{ ez_render_field(row.contributor, 'first_name') }}
-                    {{ ez_render_field(row.contributor, 'last_name') }}
+                    {% if row.contributor is instanceof('eZ\\Publish\\API\\Repository\\Values\\User\\User') %}
```
Nattfarinn: We don't need a plugin. `{% if row.contributor is defined %}` would be sufficient here.

src/bundle/Templating/Twig/InstanceOfExtension.php
```diff
@@ -0,0 +1,36 @@
...
+     * @param $var
+     * @param $instance
+     *
+     * @return bool
+     *
+     * @throws \ReflectionException
+     */
+    public function isInstanceOf($var, $instance): bool
+    {
+        return \is_object($var) && (new \ReflectionClass($instance))->isInstance($var);
```
Nattfarinn: wtf-meme-face-11

src/bundle/Resources/views/dashboard/tab/all_media.html.twig
```diff
@@ -17,8 +17,12 @@
                 <td><a href="{{ url('_ez_content_view', { 'contentId': row.contentId }) }}">{{ row.name }}</a></td>
                 <td>{{ row.type }}</td>
                 <td>
-                    {{ ez_render_field(row.contributor, 'first_name') }}
-                    {{ ez_render_field(row.contributor, 'last_name') }}
+                    {% if row.contributor is not null %}
+                        {{ ez_render_field(row.contributor, 'first_name') }}
```
webhdx: I actually don't like the idea of hardcoding `first_name` and `last_name` as as for now any content type can represent an user as long as a it has `ezuser` fieldtype. You are meant to use `getName()` method to get a proper name of the user. You can reference User Menu: https://github.com/ezsystems/ezplatform-admin-ui/blob/df1e4e7704b2d4749f1a7f4bed0a3d5f7205bf98/src/bundle/Resources/views/parts/navigation.html.twig#L21

PR #321, opened 2018-01-23
EZP-28488: A new version of a draft is created every time a content item is edited

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28488
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Tests pass?   |yes
| Doc needed?   | no

When close out the edit mode by clicking on the 'cross' symbol without saving the changes then a new Draft version should NOT be created.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_content-edit.scss
+++ b/src/bundle/Resources/public/scss/_content-edit.scss
@@ -9,6 +9,7 @@
         color: $ez-black;
         width: 2rem;
         height: 2rem;
+        cursor: pointer;
 
         &:before,
         &:after {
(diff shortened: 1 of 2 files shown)


micszo:
@barbaragr proceed with caution, this probably requires further discussion.

mikadamczyk:
Ok, so probably this need a new action

webhdx:
Nope. This ticket needs reworking content edit so that the draft is not created on entering edit mode at all.

andrerom:
@webhdx Just be somewhat aware of autosave feature, PM will be asking about the return of this in some form, and it will affect this. Or at least if the change you propose here is big in terms of repo forms and all, if so it should probably be thought about a bit during the change here, so you don't need to revert the whole thing later on.

Before it was basically done by having a internal draft created when you enter edit, and only transform it to a real draft once user clicks _save draft_  or _save and exit_ buttons or similar.

PR #322, opened 2018-01-23
EZP-28766: Notification gets covered by radio-button in content editor

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28766
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

Notification gets covered by radio-button in content editor.

screen shot 2018-01-23 at 10 30 59 am

screen shot 2018-01-23 at 11 03 33 am

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_notifications.scss
+++ b/src/bundle/Resources/public/scss/_notifications.scss
@@ -3,6 +3,7 @@
     bottom: 0;
     width: 100%;
     line-height: 2.25;
+    z-index: 50000;
 
     .alert {
         padding: 0.75rem 5rem;


src/bundle/Resources/public/scss/_notifications.scss
```diff
@@ -3,6 +3,7 @@
     bottom: 0;
     width: 100%;
     line-height: 2.25;
+    z-index: 1;
```
micszo: @dew326 is the value big enough? 🙂

sunpietro: @micszo it's big enough.

dew326: It will work, but for me, it should be something bigger like 50000 (UDW has 99999), just for safety.

micszo:
@barbaragr this one should be quick.

mikadamczyk:
As @sunpietro mentioned it is also cover the case when the rich text field type menu bar is appearing below the notification bar

src/bundle/Resources/public/scss/_notifications.scss
dew326: And what if someone wants to create an element that is before all other but behind the notifications. With z-index 1 he has no space for adding rule to own element

mikadamczyk: @dew326 ok, safety first. I will change to 50000

micszo:
@sunpietro on 2.0 I can confirm that rich text menu and `+` button are also affected.

lserwatka:
But, let's fix it as separate PR. I'm merging this one.

micszo:
The way I understood @mikadamczyk this fix should cover both cases. 🙂

lserwatka:
OK, good 👍

PR #327, opened 2018-01-24
[EZP-28683] Removes asterisk sign from labels when content editing

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28676
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

At first wanted to edit in _base-fields.scss but then it turns out that asterisk is added also in _forms.scss. 
Dont know if it should be somehow excluded from .ez-data-source__label instead.

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_mixins.scss
+++ b/src/bundle/Resources/public/scss/_mixins.scss
@@ -172,15 +172,9 @@
         content: ': ';
     }
 
-    &.required:not(.ez-data-source__label) {
+    &.required {
         &:after {
             content: '*: ';
-        }
-    }
-
-    &.required.ez-data-source__label {
-        &:after {
-            content: ': ';
         }
     }
 }
--- a/src/bundle/Resources/public/scss/_mixins.scss
+++ b/src/bundle/Resources/public/scss/_mixins.scss
@@ -172,9 +172,15 @@
         content: ': ';
     }
 
-    &.required {
+    &.required:not(.ez-data-source__label) {
         &:after {
             content: '*: ';
+        }
+    }
+
+    &.required.ez-data-source__label {
+        &:after {
+            content: ': ';
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/public/scss/_mixins.scss
```diff
@@ -172,9 +172,15 @@
         content: ': ';
     }
 
-    &.required {
+    &.required:not(.ez-data-source__label) {
```
sunpietro: I think we can avoid adding `:not(.ez-data-source__label)` here, as the next selector is a stronger than this one.

src/bundle/Resources/public/scss/_mixins.scss
```diff
@@ -172,9 +172,15 @@
...
         &:after {
             content: '*: ';
         }
     }
+
+    &.required.ez-data-source__label {
```
dew326: Cannot we just drop using this mixin for `.ez-data-source__label`? https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/bundle/Resources/public/scss/fieldType/edit/_base-field.scss#L3

And just add `&:after {content: ": "}` https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/bundle/Resources/public/scss/fieldType/edit/_base-field.scss#L51 or in a twig?

Because for me it seems like generic mixin gets information about where is used and it's no longer generic. ;)

ViniTou: Thought about this, but then it needs to be somehow duplicated in 
https://github.com/ezsystems/ezplatform-admin-ui/blob/dbffc62c46e8b6137895402184b286fae43f2f97/src/bundle/Resources/public/scss/_forms.scss#L2

I am right?

ViniTou:
@sunpietro 
@dew326 
I channeled inner frontend dev and tried to improve this. Can you take a look at that now? :)

PR #330, opened 2018-01-27
adapt targetPath for login form if previous request was on an ez_urla…

noodle69:
adapt targetPath for login form if previous request was on an ez_urla…

| Question      | Answer
| ------------- | ---
| Tickets       | N/A
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/Security/login.html.twig
+++ b/src/bundle/Resources/views/Security/login.html.twig
@@ -35,7 +35,12 @@
                                 </div>
 
                                 <input type="hidden" name="_csrf_token" value="{{ csrf_token("authenticate") }}" />
-                                <input type="hidden" name="_target_path" value="{{ url('ezplatform.dashboard') }}" />
+                                {% set targetPath = url('ezplatform.dashboard') %}
+                                {% set sfRedirect = app.session.get('sf_redirect') %}
+                                {% if sfRedirect and sfRedirect.status_text == 'Found' and sfRedirect.route == 'ez_urlalias' %}
+                                    {% set targetPath = app.session.get('_security.ezpublish_front.target_path') %}
+                                {% endif %}
+                                <input type="hidden" name="_target_path" value="{{ targetPath }}" />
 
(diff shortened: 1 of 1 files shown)


webhdx:
Hey @noodle69 I'm not a fan of keeping target path resolving code in Twig, it should not belong there. `targetPath` should be resolved in the controller (f.ex. similar to [this](https://symfony.com/doc/current/security/form_login.html#redirecting-to-the-last-accessed-page-with-targetpathtrait)).

noodle69:
Hi @webhdx, you are right
i found no service or controller were to catch previous request to inject in hidden input login form
if you found one it would be great for the #OMOB epsiode 10 :P

src/bundle/Resources/views/Security/login.html.twig
```diff
@@ -35,7 +35,12 @@
                                 </div>
 
                                 <input type="hidden" name="_csrf_token" value="{{ csrf_token("authenticate") }}" />
-                                <input type="hidden" name="_target_path" value="{{ url('ezplatform.dashboard') }}" />
+                                {% set targetPath = url('ezplatform.dashboard') %}
+                                {% set sfRedirect = app.session.get('sf_redirect') %}
```
adamwojs: Q: Why `sf_redirect` is taken into consideration in resolving target path? 

PS. In SF 3.4.4 `sf_redirect` was moved from session to cookies (see https://github.com/symfony/symfony/pull/25719).

noodle69: A: if i'm not conneced to admin siteaccess and try to go to http://<...>/admin/Media/Files (from my bookmark), i'm redirected to form login.
After connection, i'm redirected to the dashboard, never to page i've previously requested

I've noticed this "issue" when i've dvelop https://packagist.org/packages/edgar/ez-webpush-bundle

i've noticed quote from @webhdx and your info about deprecation, it was definitely not a good idea to manage this redirection in the login twig :P

PR #333, opened 2018-02-01
EZP-28796: Fieldtype 'ezselection' is not handling empty value properly

dew326:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28796
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/fieldtypes/edit/ezselection.html.twig
+++ b/src/bundle/Resources/views/fieldtypes/edit/ezselection.html.twig
@@ -12,7 +12,7 @@
     </ul>
     <ul class="ez-data-source__options ez-data-source__options--hidden">
         {%- if placeholder is not none -%}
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/views/fieldtypes/edit/ezselection.html.twig
```diff
@@ -11,6 +11,9 @@
         {% endfor %}
     </ul>
     <ul class="ez-data-source__options ez-data-source__options--hidden">
+        {%- if placeholder is not none -%}
```
franck-grenier: Here, you do not seem to manage if the field is required or not.

Will it be possible not to set any value ?

webhdx: It will work for required field - it will select first value by default as required field cannot be empty.

It won't be possible to not set any value as the validation will kick in.

PR #334, opened 2018-02-01
EZP-27776: Version conflict window when I edit an existing draft base…

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-27776
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

As a Editor, I want to see a draft conflict window when I edit a content item which has existing drafts

screen shot 2018-02-02 at 2 19 41 pm

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/Version/VersionConflictController.php
+++ b/src/bundle/Controller/Version/VersionConflictController.php
@@ -1,0 +1,61 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformAdminUiBundle\Controller\Version;
+
+use eZ\Publish\API\Repository\ContentService;
+use EzSystems\EzPlatformAdminUi\Specification\Version\VersionHasConflict;
+use EzSystems\EzPlatformAdminUiBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Response;
+use eZ\Publish\Core\Base\Exceptions\BadStateException;
+
+class VersionConflictController extends Controller
+{
+    /**
+     * @var ContentService
+     */
+    private $contentService;
+
+    /**
+     * @param ContentService $contentService
+     */
+    public function __construct(ContentService $contentService)
+    {
+        $this->contentService = $contentService;
+    }
+
+    /**
+     * Checks if Version has conflict with another published Version.
+     *
+     * If Version has no conflict, return empty Response. If it has conflict return HTML with content of modal.
+     *
+     * @param int $contentId
+     * @param int $versionNo
+     *
+     * @return Response
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     * @throws \InvalidArgumentException
+     * @throws \eZ\Publish\Core\Base\Exceptions\BadStateException
+     */
+    public function versionHasNoConflictAction(int $contentId, int $versionNo): Response
+    {
+        $versionInfo = $this->contentService->loadVersionInfoById($contentId, $versionNo);
+
+        if (!$versionInfo->isDraft()) {
+            throw new BadStateException('Version status', 'status is not draft');
+        }
+
+        if ((new VersionHasConflict($this->contentService))->isSatisfiedBy($versionInfo)) {
+            return new Response('', Response::HTTP_CONFLICT);
+        }
+
+        return new Response();
+    }
+}
--- a/src/bundle/Resources/config/routing.yml
+++ b/src/bundle/Resources/config/routing.yml
@@ -418,6 +418,13 @@ ezplatform.version.remove:
     defaults:
         _controller: 'EzPlatformAdminUiBundle:Version:remove'
 
+ezplatform.version.has_no_conflict:
+    path: /version/has-no-conflict/{contentId}/{versionNo}/{languageCode}
+    options:
+    defaults:
+        _controller: 'EzPlatformAdminUiBundle:Version\VersionConflict:versionHasNoConflict'
+        languageCode: ~
+
 # LocationView / Locations tab
 
 ezplatform.location.add:
--- a/src/bundle/Resources/views/content/tab/versions/tab.html.twig
+++ b/src/bundle/Resources/views/content/tab/versions/tab.html.twig
@@ -12,18 +12,12 @@
             'attr': { 'class': 'ez-toggle-btn-state', 'data-toggle-button-id': '#delete-translations-' ~ form_version_remove_draft.remove.vars.id }
         }) }}
         {% include '@EzPlatformAdminUi/parts/table_header.html.twig' with { headerText: 'tab.versions.draft_under_edit'|trans()|desc('Draft under edit'), tools: tab.table_header_tools(form_version_remove_draft) } %}
-        {% if draft_pager.currentPageResults is not empty %}
-            {{ include('@EzPlatformAdminUi/content/tab/versions/table.html.twig', {
-                'versions': draft_pager.currentPageResults,
-                'is_draft': true,
-                'form': form_version_remove_draft,
-                'haveToPaginate': draft_pager.haveToPaginate
-            }) }}
-        {% else %}
-            <p class="ez-relations-box-list-no-content">
-                {{ 'tab.versions.no_permission'|trans()|desc('You don\'t have access to view the content item\'s versions') }}
-            </p>
-        {% endif %}
+        {{ include('@EzPlatformAdminUi/content/tab/versions/table.html.twig', {
+            'versions': draft_pager.currentPageResults,
+            'is_draft': true,
+            'form': form_version_remove_draft,
+            'haveToPaginate': draft_pager.haveToPaginate
+        }) }}
         {{ form_end(form_version_remove_draft) }}
 
         {% if draft_pager.haveToPaginate %}
@@ -47,6 +41,7 @@
             </div>
         {% endif %}
     </section>
+    {% include '@EzPlatformAdminUi/content/modal_version_conflict.html.twig' %}
 {% endif %}
 
 {% if published_versions is not empty %}
@@ -106,6 +101,7 @@
 {% block javascripts %}
     {% javascripts
     '@EzPlatformAdminUiBundle/Resources/public/js/scripts/button.state.toggle.js'
+    '@EzPlatformAdminUiBundle/Resources/public/js/scripts/admin.version.edit.conflict.js'
     %}
     <script type="text/javascript" src="{{ asset_url }}"></script>
     {% endjavascripts %}
--- a/src/bundle/Resources/views/content/tab/versions/table.html.twig
+++ b/src/bundle/Resources/views/content/tab/versions/table.html.twig
@@ -48,12 +48,23 @@
             </td>
             {% if is_draft %}
                 <td>
-                    <a href="{{ path('ez_content_draft_edit', { 'contentId': version.contentInfo.id, 'versionNo': version.versionNo, 'language': version.translations[0].languageCode }) }}"
-                       class="btn btn-icon" title="{{ 'tab.versions.table.action.draft.edit'|trans|desc('Edit Draft') }}">
+                    <button
+                        data-content-draft-edit-url="{{ path('ez_content_draft_edit', {
+                            'contentId': version.contentInfo.id,
+                            'versionNo': version.versionNo,
+                            'language': version.translations[0].languageCode
+                        }) }}"
+                        data-version-has-conflict-url="{{ path('ezplatform.version.has_no_conflict', {
+                            'contentId': version.contentInfo.id,
+                            'versionNo': version.versionNo,
+                            'languageCode': version.translations[0].languageCode
+                        }) }}"
+                        class="btn btn-icon ez-btn--content-draft-edit"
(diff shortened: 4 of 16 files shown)


lserwatka:
I changed it for master branch as it is a new feature.

src/bundle/Resources/public/js/scripts/admin.content.draft.edit.js
```diff
@@ -0,0 +1,25 @@
+(function (global, doc) {
+    const editVersion = (event) => {
+        const contentId = event.currentTarget.dataset.contentId;
+        const versionNo = event.currentTarget.dataset.versionNo;
+        const languageCode = event.currentTarget.dataset.languageCode;
+        const contentDraftEditUrl = event.currentTarget.dataset.href;
+        const checkLink = window.Routing.generate('ezplatform.version.check', { contentId, versionNo, languageCode });
```
dew326: Please use `global` instead of `window`, it applies to all `window`

src/bundle/Resources/public/js/scripts/admin.content.draft.edit.js
```diff
@@ -0,0 +1,25 @@
...
+
+        event.preventDefault();
+
+        let request = $.get(checkLink);
+
+        request.fail(function(xhr) {
+            if(xhr.status === 409){
```
dew326: CS `if (xhr.status === 409) {`

src/lib/Tests/Specification/VersionHasConflictTest.php
```diff
@@ -0,0 +1,124 @@
...
+             * @param array $properties
+             */
+            public function __construct(array $properties)
+            {
+                parent::__construct($properties);
+                $this->contentInfo = $properties['contentInfo'];
+            }
+
+            public function getNames()
+            {
```
webhdx: There is no point in implementing an anonymous class which takes 20+ lines of code and half of the code are invalid methods (against the interface, I'd expect `getNames()` to at least return an empty array). Please stick to mocks, they looks 100 times better and doesn't require you to write invalid code.

src/lib/Specification/VersionHasConflict.php
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @return bool
+     *
+     * @throws UnauthorizedException
+     * @throws BadStateException
+     */
+    public function isSatisfiedBy(VersionInfo $versionInfo): bool
+    {
+        if (!$versionInfo->isDraft()) {
+            throw new BadStateException('Version status', 'status is not draft');
```
webhdx: I don't think `Specification` should throw any exceptions. You are definitely checking too many things here. Please implement separate class for checking if the version is draft (so that we can reuse it in the future) and chain those specification objects together.

src/bundle/Resources/public/js/scripts/admin.content.draft.edit.js
```diff
@@ -0,0 +1,25 @@
...
+                $('.modal-content').html(xhr.responseText);
+                $('#draft-has-conflict-modal').modal('show');
+            }
+            if(xhr.status === 301){
```
webhdx: I discussed it with @dew326 and it's kind of awkward that the endpoint won't ever return 2xx status. Also from the endpoint's POV 301 would mean that the endpoint is moved permanently which is not true. I'd opt for `204 No Content` status here just to confirm that there is no version conflict.

Also there are code style infringements.

src/bundle/Resources/config/routing.yml
```diff
@@ -418,6 +418,14 @@ ezplatform.version.remove:
     defaults:
         _controller: 'EzPlatformAdminUiBundle:Version:remove'
 
+ezplatform.version.check:
+    path: /version/check/{contentId}/{versionNo}/{languageCode}
+    options:
+        expose: true
+    defaults:
+        _controller: 'EzPlatformAdminUiBundle:Version\VersionCheck:versionCheck'
+        languageCode: null
```
webhdx: Use `~` instead of `null`

src/bundle/Resources/views/content/tab/versions/tab.html.twig
```diff
@@ -106,6 +113,7 @@
 {% block javascripts %}
     {% javascripts
     '@EzPlatformAdminUiBundle/Resources/public/js/scripts/button.state.toggle.js'
+    '@EzPlatformAdminUiBundle/Resources/public/js/scripts/admin.content.draft.edit.js'
```
dew326: I'm not sure if `content draft edit` is a proper name. I'm thinking about next feature `edit conflict` and we will have to edit conflicts. This should be something like `version edit conflict`.

src/bundle/Resources/views/content/tab/versions/table.html.twig
```diff
@@ -48,8 +48,16 @@
             </td>
             {% if is_draft %}
                 <td>
-                    <a href="{{ path('ez_content_draft_edit', { 'contentId': version.contentInfo.id, 'versionNo': version.versionNo, 'language': version.translations[0].languageCode }) }}"
-                       class="btn btn-icon" title="{{ 'tab.versions.table.action.draft.edit'|trans|desc('Edit Draft') }}">
+                    <a data-href="{{ path('ez_content_draft_edit', {
```
dew326: This should be just a `button` not `a`.

src/bundle/Resources/public/js/scripts/admin.content.draft.edit.js
```diff
@@ -0,0 +1,25 @@
...
+    const editVersion = (event) => {
+        const contentId = event.currentTarget.dataset.contentId;
+        const versionNo = event.currentTarget.dataset.versionNo;
+        const languageCode = event.currentTarget.dataset.languageCode;
+        const contentDraftEditUrl = event.currentTarget.dataset.href;
+        const checkLink = window.Routing.generate('ezplatform.version.check', { contentId, versionNo, languageCode });
+
+        event.preventDefault();
+
+        let request = $.get(checkLink);
```
dew326: We want to avoid using jQuery so `fetch` is your friend ;)

src/bundle/Resources/public/js/scripts/admin.content.draft.edit.js
```diff
@@ -0,0 +1,25 @@
...
+
+        request.fail(function(xhr) {
+            if(xhr.status === 409){
+                $('.modal-content').html(xhr.responseText);
+                $('#draft-has-conflict-modal').modal('show');
+            }
+            if(xhr.status === 301){
+                window.location.href = contentDraftEditUrl;
```
dew326: As @webhdx  pointed this should be on success. And second thing for me this endpoint should return an `url` on success where redirect user.

src/bundle/Resources/views/content/modal_content_draft_edit.html.twig
```diff
@@ -0,0 +1,39 @@
+{% trans_default_domain 'content_edit_draft' %}
+
+<div class="modal-header">
+    <h5>{{ 'conflict.edit.header'|trans|desc('Version conflict when editing') }}</h5>
+    <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+        <span aria-hidden="true">&times;</span>
+    </button>
+</div>
+<div class="modal-body">
+    <p class="ez-modal-body__main">
```
dew326: Why do we need this empty paragraph?

src/bundle/Resources/views/content/modal_content_draft_edit.html.twig
```diff
@@ -0,0 +1,39 @@
...
+
+    </p>
+    <p class="ez-modal-body__explanation">
+        <strong>
```
dew326: Not sure if this should be in the `strong`

src/bundle/Resources/views/content/modal_content_draft_edit.html.twig
```diff
@@ -0,0 +1,39 @@
...
+</div>
+<div class="modal-footer">
+    <button type="button" class="btn btn-secondary btn--no" data-dismiss="modal">
+        {{ 'delete.form.cancel'|trans|desc('Cancel') }}
+    </button>
+    <a class="btn btn-primary"
+        href="{{ contentEditUrl }}">
+        {{ 'delete.form.continue'|trans|desc('Continue') }}
+    </a>
+</div>
```
dew326: newline at the end

src/bundle/Resources/views/content/modal_content_draft_edit.html.twig
```diff
@@ -0,0 +1,39 @@
...
+        ) }}
+    </p>
+    <p class="ez-modal-body__explanation">
+        {{ 'conflict.edit.choices'|trans({'%languageName%': languageName})|desc(
+        'To resolve the conflict, click on Cancel and delete the draft. Then edit the published content item in %languageName%'
+        ) }}
+    </p>
+</div>
+<div class="modal-footer">
+    <button type="button" class="btn btn-secondary btn--no" data-dismiss="modal">
```
dew326: `btn--no`?

lserwatka:
@micszo we can start round of QA here

lserwatka:
@webhdx @dew326 please have a final review round

src/bundle/Controller/Version/VersionConflictController.php
```diff
@@ -0,0 +1,67 @@
...
+     * @param string $languageCode
+     *
+     * @return Response
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     * @throws \InvalidArgumentException
+     * @throws \eZ\Publish\Core\Base\Exceptions\BadStateException
+     */
+    public function versionConflictAction(int $contentId, int $versionNo, string $languageCode): Response
```
webhdx: `$languageCode` is not used anywhere.

src/bundle/Controller/Version/VersionConflictController.php
```diff
@@ -0,0 +1,67 @@
...
+class VersionConflictController extends Controller
+{
+    /**
+     * @var ContentService
+     */
+    private $contentService;
+    /**
+     * @var LanguageService
+     */
+    private $languageService;
```
webhdx: `$languageService` is not used anymore.

src/bundle/Resources/config/routing.yml
```diff
@@ -418,6 +418,13 @@ ezplatform.version.remove:
     defaults:
         _controller: 'EzPlatformAdminUiBundle:Version:remove'
 
+ezplatform.version.has_conflict:
+    path: /version/has-conflict/{contentId}/{versionNo}/{languageCode}
+    options:
+    defaults:
+        _controller: 'EzPlatformAdminUiBundle:Version\VersionConflict:versionConflict'
```
webhdx: No quotes here

src/bundle/Resources/views/content/tab/versions/tab.html.twig
```diff
@@ -47,6 +47,47 @@
             </div>
         {% endif %}
     </section>
+    <div class="modal fade ez-modal ez-modal--version-conflict" id="version-conflict-modal" tabindex="-1" role="dialog">
```
dew326: Maybe it's good to have this modal in own twig and do an include?

src/bundle/Resources/config/routing.yml
```diff
@@ -418,6 +418,13 @@ ezplatform.version.remove:
     defaults:
         _controller: 'EzPlatformAdminUiBundle:Version:remove'
 
+ezplatform.version.has_conflict:
+    path: /version/has-conflict/{contentId}/{versionNo}/{languageCode}
```
webhdx: I'm still not convinced with the naming (sorry for bringing this one more time). *Has conflict* with 200 status is kinda misleading as I'd assume 200 is a confirmation. cc @lserwatka

src/bundle/Resources/views/content/tab/versions/tab.html.twig
```diff
@@ -47,6 +47,47 @@
...
+                        'To resolve the conflict, click on Cancel and delete the draft. Then edit the published content item.'
+                        ) }}
+                    </p>
+                </div>
+                <div class="modal-footer">
+                    <button type="button" class="btn btn-secondary" data-dismiss="modal">
+                        {{ 'version.conflict.form.cancel'|trans|desc('Cancel') }}
+                    </button>
+                    <a id="edit-conflicted-draft" class="btn btn-primary"
+                       href="">
```
webhdx: Not an HTML expert but I don't think empty `href` attribute creates valid HTML element. How about using `#` as a default value so at least it has a meaning?

src/lib/Specification/NotSpecification.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
```
webhdx: Missing `declare(strict_types=1)`

src/bundle/Resources/config/routing.yml
mikadamczyk: As we agreed, quotes stay in place, to keep consistency in this file

src/bundle/Resources/config/routing.yml
lserwatka: Yes, @webhdx is right. 200 would indicate there is a conflict. So to retain this name, we would have to change a logic in the code. Or change a route name.

mikadamczyk: Ok, I changed route and action name to "hasNoConflict"

PR #338, opened 2018-02-07
EZEE-1836: Fix broken tests

mnocon:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZEE-1836
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | Yes 🐧  
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review


src/lib/Behat/PageElement/Element.php
```diff
@@ -14,14 +14,17 @@
...
+
     protected $fields;
 
     public function __construct(UtilityContext $context)
     {
         $this->context = $context;
     }
 
-    abstract public function verifyVisibility(): void;
+    public function verifyVisibility(): void
```
mnocon: I've reverted that change, it would require too much changes in other repositories at this moment.

src/lib/Behat/BusinessContext/ContentViewContext.php
```diff
@@ -14,11 +18,56 @@ class ContentViewContext extends BusinessContext
...
+        $contentStructurePage = PageObjectFactory::createPage($this->utilityContext, ContentStructurePage::PAGE_NAME);
+        $updatePage = $contentStructurePage->startCreatingContent('Article');
+
+        $updatePage->updateForm->fillFIeldWithValue('Title', $name);
+        $updatePage->updateForm->fillRichtextWithValue('Test desc');
+    }
+
+    /**
+     * @Given I start editing the content in :language language
+     */
```
m-tyrala: We should decide also if we comment parameters or not. We have both cases in our code.

src/lib/Behat/Helper/UtilityContext.php
```diff
@@ -196,6 +196,18 @@ function () use ($selector, $baseElement) {
...
+     * Filters an array of elements and returns the visible ones.
+     *
+     * @param NodeElement[] $elements
+     *
+     * @return NodeElement[]
+     */
+    public function getVisibleElements(array $elements): array
+    {
+        return array_filter($elements, function ($element) { return $element->isVisible(); });
+    }
```
m-tyrala: Nice!

src/lib/Behat/PageElement/LanguagePicker.php
```diff
@@ -0,0 +1,15 @@
...
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformAdminUi\Behat\PageElement;
+
+class LanguagePicker extends Element
+{
+    public function chooseLanguage($language)
+    {
+        $this->context->getElementByText($language, '#content_edit_language .form-check-label')->click();
```
m-tyrala: Please move selector to variable.

src/lib/Behat/PageElement/LeftMenu.php
```diff
@@ -0,0 +1,30 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformAdminUi\Behat\PageElement;
+
+use PHPUnit\Framework\Assert;
+
+class LeftMenu extends Element
+{
+    private $buttonSelector = '.ez-sticky-container .btn';
+    private $menuSelector = 'ez-side-menu';
```
m-tyrala: Please move it to `$fields` array like in other Elements.
Also, is the lack of '.' in `$menuSelector` on purpose?

src/lib/Behat/PageElement/UniversalDiscoveryWidget.php
```diff
@@ -0,0 +1,62 @@
...
+{
+    public function __construct(UtilityContext $context)
+    {
+        parent::__construct($context);
+        $this->fields = [
+            'tabSelector' => '.c-tab-nav-item',
+            'confirmButton' => '.m-ud__action--confirm',
+            'selectContentButton' => '.c-meta-preview__btn--select',
+            'elementSelector' => '.c-finder-tree-branch:nth-of-type(%d) .c-finder-tree-leaf',
+        ];
```
m-tyrala: Actually it's a good point - we should define fields in this way rather than through defining new `$fields` variable. Please change it also in other places.

src/lib/Behat/PageObject/ContentStructurePage.php
```diff
@@ -26,22 +26,17 @@ public function __construct(UtilityContext $context)
...
+     * @param string $contentType
      */
-    public function startCreatingContent(string $contentType): void
+    public function startCreatingContent(string $contentType): UpdateItemPage
     {
         $this->rightMenu->clickButton('Create');
         $this->context->getElementByText($contentType, '.form-check-label')->click();
+
+        return PageObjectFactory::createPage($this->context, UpdateItemPage::PAGE_NAME);
     }
```
m-tyrala: I was thinking about whether we should write functions in Pages, or just treat it like set of elements and write functions only in Contexts (by calling elements through pages).

src/lib/Behat/BusinessContext/ContentViewContext.php
mnocon: Ok, what's your opinion? I actually feel like they're not needed most of the time (and we should add them only when a description is really needed).

src/lib/Behat/PageElement/LeftMenu.php
mnocon: 1) 👍 
2) Nope, I've missed it - added together with a missing `verifyVisibility` call

src/lib/Behat/BusinessContext/ContentViewContext.php
m-tyrala: I think the same, so maybe we just remove it where parameter meanings are trivial.

src/lib/Behat/PageObject/ContentStructurePage.php
mnocon: For me Pages should definitely contain functions: for example editing content (which requires RightMenu and LanguagePicker) should be a function `editContent($language)` which will perform required actions.

mnocon:
I will resolve the conflicts

mnocon:
@lserwatka this one can be merged: I can merge it upstream later (it requires changing the metarepo branch in composer.json from 2.0 to master)

PR #349, opened 2018-02-16
EZP-28170 Behat Covering of Roles

m-tyrala:
| Question      | Answer
| ------------- | ---
| Tickets       | 
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/features/Roles.feature
+++ b/features/Roles.feature
@@ -1,0 +1,168 @@
+Feature: Roles management
+  As an administrator
+  In order to customize my eZ installation
+  I want to manage users Roles.
+
+  Background:
+    Given I am logged as "admin"
+      And I go to "Roles" in "Admin" tab
+
+  @javascript @common
+  Scenario: Changes can be discarded while creating Role
+    When I start creating new "Role"
+      And I set "Name" to "Test RL"
+      And I click on the edit action bar button "Discard changes"
+    Then I should be on "Roles" page
+      And there's no "Test RL" on "Roles" list
+
+  @javascript @common
+  Scenario: New Role can be created
+    When I start creating new "Role"
+      And I set "Name" to "Test RL"
+      And I click on the edit action bar button "Create"
+    Then I should be on "Role" "Test RL" page
+      And "Policies" list in "Role" "Test RL" is empty
+      And "Assignments" list in "Role" "Test RL" is empty
+
+  @javascript @common
+  Scenario: I can navigate to Roles through breadcrumb
+    Given I go to "Test RL" "Role" page
+    When I click on "Roles" on breadcrumb
+    Then I should be on "Roles" page
+
+  @javascript @common
+  Scenario: Changes can be discarded while editing Role
+    Given there's "Test RL" on "Roles" list
+    When I start editing "Role" "Test RL"
+      And I set "Name" to "Test RL edited"
+      And I click on the edit action bar button "Discard changes"
+    Then I should be on "Roles" page
+      And there's "Test RL" on "Roles" list
+      And there's no "Test RL edited" on "Roles" list
+
+  @javascript @common
+  Scenario: Role can be edited
+    Given there's "Test RL" on "Roles" list
+    When I start editing "Role" "Test RL"
+      And I set "Name" to "Test RL edited"
+      And I click on the edit action bar button "Save"
+    Then I should be on "Role" "Test RL edited" page
+      And "Policies" list in "Role" "Test RL" is empty
+      And "Assignments" list in "Role" "Test RL" is empty
+
+  @javascript @common
+  Scenario: User assignation can be discarded
+    Given there's "Test RL edited" on "Roles" list
+    When I start assigning to "Test RL edited" "Role"
+      And I select "Administrator User" from "User"
+      And I select "Editors" from "Group"
+      And I additionally select "Users" from "Group"
+      And I set "Sections" to "true"
+      And I select "Media" from "role_assignment_create_sections"
+      And I click on the edit action bar button "Discard changes"
+    Then I should be on "Role" "Test RL edited" page
+      And "Policies" list in "Role" "Test RL edited" is empty
+      And "Assignments" list in "Role" "Test RL edited" is empty
+
+  @javascript @common
+  Scenario: User can be assigned to role from the Roles list
+    Given there's "Test RL edited" on "Roles" list
+    When I start assigning to "Test RL edited" "Role"
+      And I select "Administrator User" from "User"
+      And I additionally select "Anonymous User" from "User"
+      And I select "Editors" from "Group"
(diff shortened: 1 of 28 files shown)


features/Roles.feature
```diff
@@ -0,0 +1,179 @@
...
+      | option             |
+      | Administrator User |
+      | Anonymous User     |
+      And I select option "Editor" from multiple selection "Group"
+      And I set field "Subtree" to "true"
+      And I click "Select Subtree" button
+      And I select "Media/Images"
+      And I click "Select content" button
+      And i click "Confirm" button
+      And I click on the edit action bar button "Save"
```
micszo: There are seven ands in a row, I would consider going for more declarative steps. But if you insist on going pure imperative style then that's fine by me too.

features/Roles.feature
```diff
@@ -0,0 +1,179 @@
...
+      And I select options from multiple selection "Class"
+      | option  |
+      | Article |
+      | Folder  |
+      And I click "Select location" button for "Node"
+      And I select "Users/Anonymous Users"
+      And I click "Select content" button
+      And i click "Confirm" button
+      And I select options "Lock:Locked" from multiple selection "State"
+      And I click on the edit action bar button "Update"
```
micszo: Same as above.

features/Roles.feature
```diff
@@ -0,0 +1,179 @@
...
+      And I select options from multiple selection "User"
+      | option             |
+      | Administrator User |
+      | Anonymous User     |
+      And I select option "Editor" from multiple selection "Group"
+      And I set field "Subtree" to "true"
+      And I click "Select Subtree" button
+      And I select "Media/Images"
+      And I click "Select content" button
+      And i click "Confirm" button
```
micszo: Small case. ;)

features/Roles.feature
m-tyrala: Actually it's just a temporary notation of action that we must take, at the end it'll be aggregated for sure.

src/lib/Behat/PageElement/UpdateForm.php
```diff
@@ -72,6 +73,11 @@ public function fillFieldWithValue(string $fieldName, string $value, ?string $co
             case 'checkbox':
                 $fieldNode->setValue(filter_var($value, FILTER_VALIDATE_BOOLEAN));
                 break;
+            case 'radio':
+                if ($fieldNode->isChecked() xor ($value === 'true')) {
```
mnocon: I'd change XOR to not equal, it's clearer for me 🙂

features/Roles.feature
```diff
@@ -0,0 +1,168 @@
...
+    Given there's "Test RL edited" on "Roles" list
+      And I go to "Test RL edited" "Role" page
+    When I delete assignment from "Test RL edited" role
+      | item                |
+      | Administrator User  |
+      | Editors             |
+      | Users	            |
+    Then I should be on "Role" "Test RL edited" page
+      And "Policies" list in "Role" "Test RL edited" is empty
+      And There's a assignment "Subtree of Location: /Media/Images" for "Anonymous User" on the "Test RL edited" assignments list
```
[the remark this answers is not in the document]
m-tyrala: Actually it's the same case as e.g. in Content Types scenarios:
> And Content Type "Test CT" has proper fields
>      | fieldName      | fieldType |
>      | CountryField   | ezcountry |
>      | DateField      | ezdate    |
and 
> And Content Type "Test CT" has field "CountryField" of type "ezcountry"
There's steps for checking multiple cheracteristics of given type, and only one. From code side, the function for aggregated actions uses the function for single one.

src/lib/Behat/PageElement/Dialog.php
```diff
@@ -14,12 +14,15 @@ class Dialog extends Element
     /** @var string Name by which Element is recognised */
     public const ELEMENT_NAME = 'Dialog';
 
-    public function __construct(UtilityContext $context)
+    public $dialogContainer;
+
+    public function __construct(UtilityContext $context, ?string $containerLocator = 'body')
```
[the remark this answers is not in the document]
m-tyrala: I'll check it for Dialog and in this case you are probably right, but for Admin list it's not the only reason - there are pages that has two admin lists (not in tabs like here), like content type details or section details, and in that case we need to specify on which we want to perform actions. 
Actually, it was working till now for ContentTypes because we had generic locators, and we checking specific records from tables, but in fact, in most cases we searched through both lists instead of the one that we specified.

src/lib/Behat/BusinessContext/RolesContext.php
```diff
@@ -0,0 +1,135 @@
...
+{
+    private $tabMapping = [
+        'policy' => 'Policies',
+        'assignment' => 'Assignments',
+    ];
+
+    /**
+     * @When I start assigning users and groups to :roleName from :pageName page
+     */
+    public function iStartAssigningTo(string $roleName, string $pageName): void
```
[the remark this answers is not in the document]
m-tyrala: No, we can assign from RolePage, as well as from RolesPage.

src/lib/Behat/PageElement/Dialog.php
```diff
@@ -14,7 +14,7 @@ class Dialog extends Element
     /** @var string Name by which Element is recognised */
     public const ELEMENT_NAME = 'Dialog';
 
-    public $dialogContainer;
+    //public $dialogContainer;
```
mnocon: 1) Please remove
2) Did the approach with `.show` failed in case of Dialog? I think we should be using it wherever we can, because it's easier to use in development (one selector less to think about when creating Dialog object).

src/lib/Behat/PageElement/UpdateForm.php
```diff
@@ -142,8 +142,14 @@ public function verifyNewFieldDefinitionFormExists(string $fieldName): void
      *
      * @param string $label
      */
-    public function clickButton(string $label): void
+    public function clickButton(string $label, int $no = 0): void
     {
-        $this->context->getElementByTextFragment($label, $this->fields['button'])->click();
+        $formButtons = $this->context->findAllWithWait($this->fields['button'], $this->context->findElement($this->fields['mainFormSection']));
+        foreach ($formButtons as $key => $formButton) {
```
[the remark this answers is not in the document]
m-tyrala: I was not familiar with this - learned something new, thanks!

src/lib/Behat/PageElement/Dialog.php
m-tyrala: It appears that this change was lost in space, time and git commits - works fine with show and added it one more time.

mnocon:
@lserwatka this one can be merged

PR #357, opened 2018-02-26
EZP-28851: Added credentials to the request checking if selected item can be considered as an image

adamwojs:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28851
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Added credentials to the request which checks if selected item can be considered as an image. Actual change: https://github.com/ezsystems/ezplatform-admin-ui/pull/357/files#diff-5ed292618b1d024126a31fe3cdd78983R20

#### Checklist:
- [x] Ready for Code Review


adamwojs:
Q: Does regenerated *.js.map files should be included in PR?

src/bundle/Resources/public/js/alloyeditor/src/base/ez-embedimage.js
```diff
@@ -17,6 +17,7 @@ export default class EzEmbedImageButton extends EzEmbedDiscoverContentButton {
         const request = new Request(selection.item.ContentInfo.Content.ContentType._href, {
             method: 'GET',
             headers: {'Accept': 'application/vnd.ez.api.ContentType+json'},
```
adamwojs: Should we also include `X-Siteaccess` and `X-CSRF-Token` headers?

andrerom: > X-CSRF-Token 

not needed for GET afaik

> X-Siteaccess

Yes, it is used in the ez-embed so should be used here to.


@sunpietro Should we abstract fetch a bit to avoid these kind of issues happens?

andrerom: Also, should we rather set `mode: 'same-origin'` instead of having to set `credentials: 'same-origin'` all over?

In which cases will we send REST request to another domain in these cases?

micszo:
That's a huge patch. 😮

micszo:
Should I worry about Storm complaining?
screen shot 2018-02-28 at 12 14 36

sunpietro:
@micszo @adamwojs I believe it can be smaller. This PR should contain only an updated map related only to a changed JS file: `ez-embedimage.js` (I'm writing the filename from memory).
Other maps can be safely removed.

andrerom:
@adamwojs can you ommit the non relevant map updates?

adamwojs:
@andrerom @sunpietro Done.

lserwatka:
@adamwojs please merge this up to master

adamwojs:
@lserwatka done.

PR #360, opened 2018-02-27
EZP-28877: As a developer I would like to inject custom tabs configuration into the Universal Discovery module instance

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28877
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | yes

**IMPORTANT:**
All existing Universal Discovery module instances should be tested, just in case it suddenly breaks (but it shouldn't).

### DIFF
--- a/src/bundle/Resources/views/layout.html.twig
+++ b/src/bundle/Resources/views/layout.html.twig
@@ -92,12 +92,17 @@
             'bundles/ezplatformadminuiassets/vendors/bootstrap/dist/js/bootstrap.min.js'
             'bundles/ezplatformadminuiassets/vendors/create-react-class/create-react-class.min.js'
             'bundles/ezplatformadminuiassets/vendors/prop-types/prop-types.min.js'
+        %}
+            <script type="text/javascript" src="{{ asset_url }}"></script>
+        {% endjavascripts %}
+        {{ ezplatform_admin_ui_component_group('custom-admin-ui-modules') }}
+        {{ ezplatform_admin_ui_component_group('custom-admin-ui-config') }}
+        {%  javascripts
             'bundles/ezplatformadminuimodules/js/UniversalDiscovery.module.js'
             'bundles/ezplatformadminui/js/scripts/button.trigger.js'
             'bundles/ezplatformadminui/js/scripts/button.prevent.default.js'
             'bundles/ezplatformadminui/js/scripts/udw/browse.js'
             'bundles/ezplatformadminui/js/scripts/admin.user.menu.js'
-
         %}
             <script type="text/javascript" src="{{ asset_url }}"></script>
         {% endjavascripts %}


src/bundle/Resources/views/layout.html.twig
```diff
@@ -93,6 +93,14 @@
             'bundles/ezplatformadminuiassets/vendors/create-react-class/create-react-class.min.js'
             'bundles/ezplatformadminuiassets/vendors/prop-types/prop-types.min.js'
             'bundles/ezplatformadminuimodules/js/UniversalDiscovery.module.js'
+        %}
+            <script type="text/javascript" src="{{ asset_url }}"></script>
+        {% endjavascripts %}
```
dew326: why do we need to separate `javascripts` into two blocks?

src/bundle/Resources/views/layout.html.twig
```diff
@@ -93,6 +93,14 @@
...
+
+        {{ ezplatform_admin_ui_component_group('custom-admin-ui-modules') }}
```
dew326: This will include custom React module/panel?

src/bundle/Resources/views/layout.html.twig
```diff
@@ -93,6 +93,14 @@
...
+        {{ ezplatform_admin_ui_component_group('custom-admin-ui-config') }}
```
dew326: and this is for adding config to the `eZ.adminUiConfig.universalDiscoveryWidget.extraTabs`?

lserwatka:
@sunpietro do you need this feature in 2.0.x? I think this should go to master branch.

sunpietro:
@lserwatka without this feature we make UDW (and any other admin UI module) hardly extensible. There's no clean way to add custom tabs into any UDW instance from any external Symfony bundle. This PR fixes it, because I've prepared layouts hooks where custom JS from external bundles can be injected before the UI scripts, that instantiate the Universal Discovery modules. By the end these instances can take the advantage of provided custom tabs.
What I forgot to mention earlier is this PR: https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/51 that implements an extensibility point directly in the Universal Discovery module.
These issues were spotted while working on tutorial for @DominikaK about extending the UDW from third-party bundles.
The sample Symfony bundle that extends Universal Discovery module can be found here: https://github.com/sunpietro/extending-udw-tab

lserwatka:
@sunpietro I understand why do you need it, but this is a new feature, so it should go to master. We don't have intention to release 2.0.3. So target branch should be master here.

sunpietro:
@lserwatka I see your point. I'll update PR as soon as possible.

sunpietro:
The base branch has been changed to `master`

PR #365, opened 2018-03-02
[Enhancement][Not TESTED] Allow conditional display of a tab based on the actual content of it

Plopix:
According to a discussion on slack with @wizhippo. I think a good idea would be to render the tab here: https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/bundle/Resources/views/parts/tab/locationview.html.twig#L8
only when it is NOT empty.

@wizhippo use case: you want to display a tab only on some conditions. Today you can add the tab but you can not decide to do not display it.

**I did not test** and I am unsure of line 11, but I will try later unless someone to do it before me.

And at least it opens the discussion ;) maybe there is a better way.

- [ ] Tested
- [ ] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/parts/tab/locationview.html.twig
+++ b/src/bundle/Resources/views/parts/tab/locationview.html.twig
@@ -1,3 +1,4 @@
+{% set tabViews = {} %}
 <div>
     <div class="ez-header px-4">
         <div class="container">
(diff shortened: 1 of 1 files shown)


lserwatka:
@webhdx what do you think about this idea?

webhdx:
I'm not convinced. I believe it should be feasible to remove/hide the tab via Event Listener. Have you guys tried it first @wizhippo @Plopix?

wizhippo:
@webhdx I have and in the event listener there is no access to the parameters, and the use case I have would use the content parameter to check permissions and content type decide to show the tab or not.

wizhippo:
Here is a example of what I want to do:  https://gist.github.com/wizhippo/555d891dbe84a1e3196825b601908f06#file-locationviewtabsubscriber-php-L36

Nattfarinn:
Like @wizhippo already mentioned, the solution is to pass `$parameters` from `TabExtension` to `TabGroupEvent` (and `TabEvent`).

Closing this in favour of: https://github.com/ezsystems/ezplatform-admin-ui/pull/384

PR #366, opened 2018-03-02
EZP-28841: Content On the Fly v2

dew326:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28841
| Requires      | https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/55
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review
- [x] restrict non-container in trash restoring (https://jira.ez.no/browse/EZP-28027)
- [x] expose languageCode in meta
- [x] check permission endpoint
- [x] route names in `ezplatform-admin-ui-modules`: https://github.com/ezsystems/ezplatform-admin-ui/pull/366#issuecomment-371489491


src/bundle/Controller/ContentController.php
```diff
@@ -115,6 +141,93 @@ public function createAction(Request $request): Response
         return $this->redirect($this->generateUrl('ezplatform.dashboard'));
     }
 
+    /**
+     * @param Request $request
+     * @param string $languageCode
+     * @param ContentType $contentType
+     * @param Location $parentLocation
+     *
+     * @return ContentCreateView
```
ViniTou: ContentCreateView | Response
?

src/lib/RepositoryForms/Form/Processor/Content/ContentOnTheFlyProcessor.php
```diff
@@ -0,0 +1,70 @@
...
+use EzSystems\RepositoryForms\Event\FormActionEvent;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\Response;
+use Twig\Environment;
+
+class ContentOnTheFlyProcessor implements EventSubscriberInterface
+{
+    private $contentService;
+
+    private $twig;
```
ViniTou: phpdocs

src/lib/RepositoryForms/View/ViewTemplatesListener.php
```diff
@@ -0,0 +1,60 @@
...
+{
+    /**
+     * Hash of [View type FQN] => template.
+     * @var array
+     */
+    protected $viewTemplates;
+
+    /**
+     * @var string
+     */
```
ViniTou: /** @var string */

src/lib/UI/Config/Provider/Module/UniversalDiscoveryWidget.php
```diff
@@ -33,9 +47,48 @@ public function getConfig(): array
...
-            'startingLocationId' => $this->configResolver->getParameter(
-                'universal_discovery_widget_module.default_location_id'
-            ),
+            'startingLocationId' => $this->getStartingLocationId(),
+            'contentOnTheFly' => $this->getContentOnTheFlyConfiguration(),
+        ];
+    }
+
+    /**
+     * @return int
```
ViniTou: int|null according to typehint

mnocon:
@Nattfarinn have you seen the PHPUnit failure?

Nattfarinn:
@ViniTou Fixed.
@mnocon Fixed.

mnocon:
Two things so far:
1) we do not check if the user has `content/publish` permission, so it's possible to receive a 403 inside UDW
2) it looks like there are infinite permission checking requests in the Network tab

bdunogier:
Is there anything as for developer documentation ? I'm mostly thinking about how you integrate COTF in your custom code and how you may customize the locations/content types that are liste.

From the code's perspective, I think that a `ContentOnTheFlyController` would help keep `ContentController` thinner.

Nattfarinn:
@mnocon 

> we do not check if the user has content/publish permission, so it's possible to receive a 403 inside UDW

Fixed.

Nattfarinn:
@bdunogier 

> From the code's perspective, I think that a ContentOnTheFlyController would help keep ContentController thinner.

True. Now it has its own controller.

@dew326 Route names changed. I guess you have to update `ezplatform-admin-ui-modules` repository.

```diff
-ezplatform.content.create.on_the_fly:
+ezplatform.content_on_the_fly.create:
```
 ```diff
-ezplatform.content.create.on_the_fly.has_access:
+ezplatform.content_on_the_fly.has_access:
```

**Comment by lserwatka** [comment]
@dew326 could you address this change so we can unblock QA here?

**Comment by dew326** [comment]
@mnocon done, please update branch in ezplatform-admin-ui-modules.

**Comment by mnocon** [comment]
I cannot create anything now, the "Create content" button is disabled all the time. In Network tab, the request: `http://v2eedemo.prod/admin/content/create/onthefly/article/eng-GB/43/hasaccess` ends with 500 Internal Server Error 😞 : 
```
Class "EzPlatformAdminUiBundle" does not exist. [in vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Controller/ControllerResolver.php:180]
```

**Comment by Nattfarinn** [comment]
@mnocon Fixed

**Comment by lserwatka** [comment]
@Nattfarinn we need a rebase here too.

File: src/bundle/Resources/config/routing.yml [hunk 5]
```diff
@@ -540,7 +540,28 @@ ezplatform.user.delete:
...
+ezplatform.content_on_the_fly.create:
+    path: /content/create/onthefly/{contentTypeIdentifier}/{languageCode}/{locationId}
+    methods: ['GET', 'POST']
+    defaults:
+        _controller: 'EzPlatformAdminUiBundle:ContentOnTheFly:createContent'
+    options:
+        expose: true
+
+ezplatform.content_on_the_fly.has_access:
+    path: /content/create/onthefly/{contentTypeIdentifier}/{languageCode}/{locationId}/hasaccess
```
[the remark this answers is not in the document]
Nattfarinn: So... `onthefly` is fine, but `hasaccess` is not? :P

src/lib/RepositoryForms/Form/Processor/Content/ContentOnTheFlyProcessor.php
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * @param FormActionEvent $event
+     *
+     * @throws \Twig_Error_Loader
+     * @throws \Twig_Error_Runtime
+     * @throws \Twig_Error_Syntax
+     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
+     * @throws \eZ\Publish\API\Repository\Exceptions\ContentFieldValidationException
```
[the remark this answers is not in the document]
Nattfarinn: We agreed to keep FQCN

bdunogier:
I've been through the code, and I'd like to know how flexible we are on the starting location / content type / language options. This includes existing inclusions of COTF, like Relation or RichText. For those, I'd like to keep the system flexible enough so that:
- an integrator can hook into the business logic and adapt the choices to the customer
- we can develop configurable mechanisms, such as "Create images in the Media/Images directory when creating content for the `image` field of the `article` type".

dew326:
@bdunogier Starting location and content type are not developed. The only configuration option is forced language, this is an option per usage in button config.

bdunogier:
What about this ?

```
startingLocationId: window.eZ.adminUiConfig.universalDiscoveryWidget.startingLocationId
```

My concern is that we limit our options in terms of BC. Without thinking it very far, ideally, there would be some kind of event that can be hooked into and lets you decide of those options. Our own subscriber would set it to the globally configured value for the first iteration. The two options I have described above would still be open for implementation / integration.

lserwatka:
@Nattfarinn rebase is needed

Nattfarinn:
@bdunogier I agree, but as far as I know, it was more like "Nice to have." instead of requirement. The scope of this was discussed on preplanning and planning, and turned out that we won't be able to deliver everything within one sprint. We can make improvements for next version I guess and I don't think refining a way how the values for a new UDW options/flags is picked will be a BC break.

bdunogier:
To be honest, I'd like it to become a requirement.

As long as you keep it in mind, I'm fine with it. Do you think that a dedicated story, with detailed expectations, makes sense for this ?

PR #372, opened 2018-03-07
EZP-28884: Removing click when editing content in one single language

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28884
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

This probably could be done more on backend part, but I think that would involve modifications in menu build since we would have to inject content services there to build correct uri. And there are already events for language select form submission.

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/sidebar/btn/content.edit.js
+++ b/src/bundle/Resources/public/js/scripts/sidebar/btn/content.edit.js
@@ -1,0 +1,13 @@
+(function (global, doc) {
+    const editButton = doc.querySelector('.ez-btn--edit');
+    const languageRadioOption = doc.querySelector('.ez-extra-actions--edit.ez-extra-actions--prevent-show [type="radio"]');
+
+    if (!languageRadioOption) {
+        return;
+    }
+
+    editButton.addEventListener('click', () => {
+        languageRadioOption.checked = true;
+        languageRadioOption.dispatchEvent(new CustomEvent('change'))
+    }, false);
+})(window, document);
--- a/src/bundle/Resources/public/js/scripts/sidebar/extra.actions.js
+++ b/src/bundle/Resources/public/js/scripts/sidebar/extra.actions.js
@@ -1,13 +1,16 @@
 (function () {
     const CLASS_HIDDEN = 'ez-extra-actions--hidden';
+    const CLASS_PREVENT_SHOW = 'ez-extra-actions--prevent-show';
     const btns = [...document.querySelectorAll('.ez-btn--extra-actions')];
 
     btns.forEach(btn => {
         btn.addEventListener('click', () => {
             const actions = document.querySelector(`.ez-extra-actions[data-actions="${btn.dataset.actions}"]`);
-            const cssMethodName = actions.classList.contains(CLASS_HIDDEN) ? 'remove' : 'add';
+            const haveHiddenPart = (element) => {
+                return element.classList.contains(CLASS_HIDDEN) && !element.classList.contains(CLASS_PREVENT_SHOW)
+            };
+            const methodName = haveHiddenPart(actions) ? 'remove' : 'add';
             const clickOutsideMethodName = actions.classList.contains(CLASS_HIDDEN) ? 'addEventListener' : 'removeEventListener';
-            const btnRect = btn.getBoundingClientRect();
             const focusElement = actions.querySelector(btn.dataset.focusElement);
             const detectClickOutside = (event) => {
                 const isNotButton = !event.target.contains(btn);
@@ -21,7 +24,7 @@
             };
 
             actions.style.top = btn.offsetTop + 'px';
-            actions.classList[cssMethodName](CLASS_HIDDEN);
+            actions.classList[methodName](CLASS_HIDDEN);
             document.body[clickOutsideMethodName]('click', detectClickOutside, false);
 
             if (focusElement) {
--- a/src/bundle/Resources/views/content/widgets/content_edit.html.twig
+++ b/src/bundle/Resources/views/content/widgets/content_edit.html.twig
@@ -1,6 +1,6 @@
 {% trans_default_domain 'locationview' %}
 
-<div class="ez-extra-actions ez-extra-actions--edit ez-extra-actions--hidden bg-secondary" data-actions="edit">
+<div class="ez-extra-actions ez-extra-actions--edit ez-extra-actions--hidden bg-secondary {% if form.language.vars.choices|length == 1 %} ez-extra-actions--prevent-show {% endif %}" data-actions="edit">
     <div class="ez-extra-actions__header">{{ 'content.edit.select_language'|trans|desc('Select language') }}</div>
     <div class="ez-extra-actions__content">
         {{ form_start(form, { 'action': path('ezplatform.content.edit') }) }}
--- a/src/bundle/Resources/views/content/locationview.html.twig
+++ b/src/bundle/Resources/views/content/locationview.html.twig
@@ -122,6 +122,7 @@
         'bundles/ezplatformadminui/js/scripts/sidebar/instant.filter.js'
         '@EzPlatformAdminUiAssetsBundle/Resources/public/vendors/leaflet/dist/leaflet.js'
         'bundles/ezplatformadminui/js/scripts/admin.location.load.map.js'
+        'bundles/ezplatformadminui/js/scripts/sidebar/btn/content.edit.js'
     %}
         <script type="text/javascript" src="{{ asset_url }}"></script>
     {% endjavascripts %}


webhdx:
I'm wondering why not just rendering from the backend with special class to let know JS it has to do redirection? @dew326 @sunpietro

ViniTou:
I think that yes, additional class could be added to trigger underlying form submission, dunno if that wouldn't introduce sort of flickering of this language thingy.

ViniTou:
@webhdx Can you look at it now?

lserwatka:
@dew326 @sunpietro could you review front-end code here?

src/bundle/Resources/public/js/scripts/sidebar/btn/content.edit.js
```diff
@@ -0,0 +1,21 @@
+(function () {
+    const editButton = document.querySelector('.ez-btn--edit[data-actions="edit"]');
```
sunpietro: Do you need that specific selector? Won't `.ez-btn--edit` selector work?

src/bundle/Resources/public/js/scripts/sidebar/btn/content.edit.js
```diff
@@ -0,0 +1,21 @@
...
+
+    editButton.addEventListener('click', () => {
+        const languageButtons = document.querySelectorAll('.ez-extra-actions--edit[data-actions="edit"] .form-check [type="radio"]');
```
sunpietro: The selector looks very specific, can't it be simpler? The second thing is that `languageButtons` should be defined in the scope above. So it will be cached once and it will be always available. Otherwise, you're looking for it everytime you're clicking on a button.

src/bundle/Resources/public/js/scripts/sidebar/btn/content.edit.js
```diff
@@ -0,0 +1,21 @@
...
+
+        if (languageButtons.length !== 1) {
```
sunpietro: When there is a possibility that there are more than 1 buttons? Or maybe, you're checking if any button exists?

src/bundle/Resources/public/js/scripts/sidebar/btn/content.edit.js
```diff
@@ -0,0 +1,21 @@
...
+            return;
+        }
+
+        const triggerOnClickEventForLanguage = (button) => {
+            button.checked = true;
+            const onChangeEvent = document.createEvent("HTMLEvents");
+            onChangeEvent.initEvent("change", true, false);
+            button.dispatchEvent(onChangeEvent);
```
sunpietro: It can be simplified, to just `button.dispatchEvent(new CustomEvent('change'))`.
Another question, why do you need that event?

src/bundle/Resources/views/content/widgets/content_edit.html.twig
```diff
@@ -1,6 +1,6 @@
 {% trans_default_domain 'locationview' %}
 
-<div class="ez-extra-actions ez-extra-actions--edit ez-extra-actions--hidden bg-secondary" data-actions="edit">
+<div class="ez-extra-actions ez-extra-actions--edit ez-extra-actions--hidden bg-secondary" {% if form.language.vars.choices|length == 1 %} hidden="hidden" {% endif %} data-actions="edit">
```
sunpietro: Instead of `hidden="hidden"` put just `hidden`. It's enough.

src/bundle/Resources/public/js/scripts/sidebar/btn/content.edit.js
```diff
@@ -0,0 +1,21 @@
...
+    editButton.addEventListener('click', () => {
+        const languageButtons = document.querySelectorAll('.ez-extra-actions--edit[data-actions="edit"] .form-check [type="radio"]');
+
+        if (languageButtons.length !== 1) {
+            return;
+        }
+
+        const triggerOnClickEventForLanguage = (button) => {
+            button.checked = true;
+            const onChangeEvent = document.createEvent("HTMLEvents");
```
sunpietro: Please, use single quotes.

src/bundle/Resources/public/js/scripts/sidebar/btn/content.edit.js
ViniTou: When there is more than one language added for content, probably languagebuttons for it is not the best name then, gonna change it.

src/bundle/Resources/public/js/scripts/sidebar/btn/content.edit.js
ViniTou: There is an event listening for this form submission (checking for draft conflicts).

ViniTou:
@sunpietro 
updated

src/bundle/Resources/public/js/scripts/sidebar/btn/content.edit.js
```diff
@@ -0,0 +1,18 @@
+(function () {
+    const editButton = document.querySelector('.ez-btn--edit');
+    const languageRadioOptions = document.querySelectorAll('.ez-extra-actions--edit [type="radio"]');
```
dew326: I the prevent class would be added you can use selector: `doc.querySelector('.ez-extra-actions--edit.ez-extra-actions--prevent-show [type="radio"]')`

src/bundle/Resources/public/js/scripts/sidebar/btn/content.edit.js
```diff
@@ -0,0 +1,18 @@
...
+        return;
+    }
+
+    editButton.addEventListener('click', () => {
+        const triggerOnClickEventForLanguage = (radioOption) => {
+            radioOption.checked = true;
+            radioOption.dispatchEvent(new CustomEvent('change'))
+        };
+
+        triggerOnClickEventForLanguage(languageRadioOptions[0]);
```
dew326: With my previous comments you will have only one radioOption są this method is not needed and you can do just:
```js
radioOption.checked = true;
radioOption.dispatchEvent(new CustomEvent('change'));
```

ViniTou:
@dew326 all done

src/bundle/Resources/public/js/scripts/sidebar/btn/content.edit.js
```diff
@@ -0,0 +1,13 @@
+(function () {
+    const editButton = document.querySelector('.ez-btn--edit');
+    const languageRadioOption = document.querySelector('.ez-extra-actions--edit.ez-extra-actions--prevent-show [type="radio"]');
+
+    if (null === languageRadioOption) {
```
dew326: You can do just `if (!languageRadioOption) {`

src/bundle/Resources/public/js/scripts/sidebar/btn/content.edit.js
```diff
@@ -0,0 +1,13 @@
...
+        return;
+    }
+
+    editButton.addEventListener('click', () => {
+        languageRadioOption.checked = true;
+        languageRadioOption.dispatchEvent(new CustomEvent('change'))
+    }, false);
+})();
```
dew326: To be consistent with other js, we pass `window` and `document` to the method and then use `global` and `doc` inside

src/bundle/Resources/public/js/scripts/sidebar/btn/content.edit.js
sunpietro: Please, no yoda comparisons in JS.

src/bundle/Resources/public/js/scripts/sidebar/extra.actions.js
```diff
@@ -1,13 +1,13 @@
 (function () {
     const CLASS_HIDDEN = 'ez-extra-actions--hidden';
+    const CLASS_PREVENT_SHOW = 'ez-extra-actions--prevent-show';
     const btns = [...document.querySelectorAll('.ez-btn--extra-actions')];
 
     btns.forEach(btn => {
         btn.addEventListener('click', () => {
             const actions = document.querySelector(`.ez-extra-actions[data-actions="${btn.dataset.actions}"]`);
-            const cssMethodName = actions.classList.contains(CLASS_HIDDEN) ? 'remove' : 'add';
+            const cssMethodName = actions.classList.contains(CLASS_HIDDEN) && !actions.classList.contains(CLASS_PREVENT_SHOW)? 'remove' : 'add';
```
sunpietro: Two things: 
1. It's not cssMethodName, because CSS doesn't have methods;
2. To improve readability, I would put `actions.classList.contains(CLASS_HIDDEN) && !actions.classList.contains(CLASS_PREVENT_SHOW)` into some kind of a named variable.

src/bundle/Resources/views/content/widgets/content_edit.html.twig
```diff
@@ -1,6 +1,6 @@
 {% trans_default_domain 'locationview' %}
 
-<div class="ez-extra-actions ez-extra-actions--edit ez-extra-actions--hidden bg-secondary" data-actions="edit">
+<div class="ez-extra-actions ez-extra-actions--edit ez-extra-actions--hidden bg-secondary {% if form.language.vars.choices|length == 1 %} ez-extra-actions--prevent-show {% endif %}" data-actions="edit">
```
sunpietro: I'm wondering, when can be a situation that a `div` has a class of `ez-extra-actions--edit` but it's action is not to edit? Maybe you could drop `data-actions="edit"` attribute?

sunpietro: I'm also wondering, if you really need this class `ez-extra-actions--prevent-show`?

ViniTou: Don't know, but identical thing is in create button so prolly, there are some js that depends on it. Even in extra.action.js there is selector with that.

PR #377, opened 2018-03-08
EZP-28922: Design consistency for modals across the whole application

inakijv:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-28922](https://jira.ez.no/browse/EZP-28922)
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

Design consistency for modals across the whole application:
- Updated styling for modals (Translations, Send to Trash and Send to Trash from Trash manager), background-colors.
- In Translations modal aligned colors for language selection for translations to the existing one in content-edit-language widget.

screen shot 2018-03-08 at 1 28 06 pm
screen shot 2018-03-08 at 1 40 34 pm
screen shot 2018-03-08 at 1 41 56 pm

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_add-translation.scss
+++ b/src/bundle/Resources/public/scss/_add-translation.scss
@@ -7,15 +7,32 @@
             display: inline-block;
             left: 50%;
             transform: translateX(-50%);
+            border-radius: .25rem;
         }
     }
 
     .modal-header {
-        background: $ez-ground-base;
+        background: $ez-color-base-pale;
+        border-bottom: none;
+
+        .modal-title {
+            font-size: 1.5rem;
+        }
+
+        .close {
+            margin-top: -.5625rem;
+        }
     }
 
     .modal-body {
         padding: 2rem;
+        background-color: $ez-ground-base-medium;
+        border-radius: 0 0 .25rem .25rem;
+    }
+
+    .modal-footer {
+        border-top: none;
+        padding: 1rem 1rem 0 1rem;
     }
 
     &__title {
@@ -34,7 +51,7 @@
 
     &__input-wrapper {
         width: 15rem;
-        background: $ez-ground-primary;
+        background: $ez-white;
         margin: .5rem;
         border-radius: 5px;
         line-height: 1;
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/scss/_add-translation.scss
```diff
@@ -7,15 +7,28 @@
...
         }
     }
 
     .modal-header {
-        background: $ez-ground-base;
+        background: $ez-color-base-pale;
+        border-bottom: none;
+
+        .close {
+            margin-top: -.55rem;
```
sunpietro: If the root font size is `16px` then `.55rem` is `8.8px`. Maybe it should be `.5625rem` (`9px`) to make it rounded to the nearest number.

src/bundle/Resources/public/scss/_modals.scss
```diff
@@ -1,9 +1,13 @@
 .ez-modal--send-to-trash {
     .modal-dialog {
-        background-color: $ez-ground-base-medium;
```
sunpietro: I would keep it here.

src/bundle/Resources/views/content/tab/translations/modal_add_translation.html.twig
```diff
@@ -6,7 +6,7 @@
         {{ form_start(form, {'action': path('ezplatform.translation.add')}) }}
         <div class="modal-content">
             <div class="modal-header">
-                <h5 class="modal-title">{{ 'tab.translations.add.title'|trans|desc('Create new translation') }}</h5>
+                <h2 class="modal-title">{{ 'tab.translations.add.title'|trans|desc('Create new translation') }}</h2>
```
sunpietro: We need to think about semantics here. If you set it to `h2` then it semantically means it is a second important title in the HTML markup. If you would like to resize the title. Then maybe you should create a CSS selector that provides a desired font size definition.

inakijv:
Thanks for your review @sunpietro :-)
Updated PR based on it.

PR #380, opened 2018-03-12
EZP-28930: Labels are one line below checkbox/radiobutton

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28930
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

# TODO

### ADMIN/CONTENT TYPES/CONTENT/CREATING A NEW CONTENT TYPE:

Global Properties:
- [x] Contaner
- [x] Default content availability

Fields:
- [x] New field definition
- [x] Required/Searchable/Translatable

ezboolean:
- [ ] Default value

ezcountry:
- [ ] multiple choice

ezdate:
- [ ] default value: empty/current date

ezdatetime:
- [ ] Use seconds
- [ ] Default value: empty/Current date/Adjusted current dateime

ezselection:
- [ ] multiple choice

### CREATING/EDITING A NEW LANGUAGE
- [ ] Enabled

### ADDING LIMITATONS
- [ ] no limitations
- [ ] sections
- [ ] subtree

### DIFF
--- a/src/bundle/Resources/views/admin/content_type/edit.html.twig
+++ b/src/bundle/Resources/views/admin/content_type/edit.html.twig
@@ -33,10 +33,10 @@
                 {{ form_row(form.description) }}
                 {{ form_row(form.nameSchema) }}
                 {{ form_row(form.urlAliasSchema) }}
-                {{ form_row(form.isContainer) }}
+                {{ form_row(form.isContainer, {'label_attr': {'class': 'checkbox-inline'}}) }}
                 {{ form_row(form.defaultSortField) }}
                 {{ form_row(form.defaultSortOrder) }}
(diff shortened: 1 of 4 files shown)


src/bundle/Resources/views/admin/content_type/edit.html.twig
```diff
@@ -113,8 +126,8 @@
         </div>
     </section>
 
-    {{ form_widget(form.publishContentType, { 'attr': { 'hidden': 'hidden' }}) }}
```
webhdx: Please do not remove quotes from the keys, this is the standard we are following in Twig templates.

PR #386, opened 2018-03-13
EZP-28874: Main Menu is not responsive under 992 pixels

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28874
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

**THINGS FIXED:**
- [x] top nav menu not responsive under 992px,
- [x] the right sidebar breaking the content edit layout at specific width dimensions.

**HOW TO TEST IT?**
Just resize the browser window and watch if navbar and the sidebar are not breaking the layout. Test on all modern browsers.

### DIFF
--- a/src/bundle/Resources/views/content/content_edit/content_create_draft.html.twig
+++ b/src/bundle/Resources/views/content/content_edit/content_create_draft.html.twig
@@ -6,13 +6,15 @@
 {% block content %}
     <div class="row align-items-stretch ez-main-row">
         {% block left_sidebar %}{% endblock left_sidebar %}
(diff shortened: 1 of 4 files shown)


dew326:
@m-tyrala The `'.navbar-expand-sm` is not the best selector for behat, it is generic one from bootstrap and there might be plenty of this on the page. I would rather use specific class for sub-nav `.ez-main-sub-nav`.

PR #390, opened 2018-03-13
EZP-28922: Modal Draft Conflict design consistency update

inakijv:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-28922](https://jira.ez.no/browse/EZP-28922)
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

Most important aspects:
- Updated Draft Conflict modal based on design consistency requirements for modals throughout the application.

ez platform 32

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_modals.scss
+++ b/src/bundle/Resources/public/scss/_modals.scss
@@ -95,3 +95,37 @@
         background-color: $ez-ground-base;
     }
 }
+
+.ez-modal--version-draft-conflict {
+    .modal-header {
+        background: $ez-color-base-pale;
+        border-bottom: none;
+
+        .modal-title {
+            font-size: 1.5rem;
+        }
+
+        .close {
+            margin-top: -.5625rem;
+        }
+    }
+
+    .modal-body {
+        padding: 2rem;
+        background-color: $ez-ground-base-medium;
+        border-radius: 0 0 .25rem .25rem;
+
+        .ez-modal-body__main {
+            margin-bottom: 1.5rem;
+
+            .ez-modal-body__main-content {
+                margin-bottom: 0;
+            }
+        }
+
+        .ez-table-header,
+        .ez-table__draft-conflict {
+            margin-bottom: 0;
+        }
+    }
+}
--- a/src/bundle/Resources/views/content/modal_draft_conflict.html.twig
+++ b/src/bundle/Resources/views/content/modal_draft_conflict.html.twig
@@ -6,21 +6,23 @@
     <div class="modal-dialog modal-lg" role="document">
         <div class="modal-content">
             <div class="modal-header">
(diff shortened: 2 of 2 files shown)


src/bundle/Resources/public/scss/_modals.scss
```diff
@@ -95,3 +95,37 @@
         background-color: $ez-ground-base;
     }
 }
+
+.ez-modal--version-draft-conflict {
+    .modal-header {
+        background: $ez-color-base-pale;
+        border-bottom: none;
+
+        h5 {
```
sunpietro: We should avoid styling by tag selectors. Please, implement the style update in the context of the `modal-title` class selector.

src/bundle/Resources/views/content/modal_draft_conflict.html.twig
```diff
@@ -6,7 +6,7 @@
     <div class="modal-dialog modal-lg" role="document">
         <div class="modal-content">
             <div class="modal-header">
-                <h5>{{ 'draft.conflict.header'|trans|desc('Draft conflict when editing') }}</h5>
+                <h5 class="modal-title">{{ 'draft.conflict.header'|trans|desc('Draft conflict when editing') }}</h5>
                 <button type="button" class="close" data-dismiss="modal" aria-label="Close">
                     <span aria-hidden="true">&times;</span>
```
sunpietro: Shouldn't we use SVG icon instead of HTML entitty for the close button symbol, the `x` sign?

inakijv: @sunpietro 👍 that's going to be covered in a different PR (Update closing buttons for modals)

inakijv:
Updated based on review

src/bundle/Resources/public/scss/_modals.scss
```diff
@@ -95,3 +95,37 @@
...
+    .modal-body {
+        padding: 2rem;
+        background-color: $ez-ground-base-medium;
+        border-radius: 0 0 .25rem .25rem;
+
+        .ez-modal-body__main {
+            margin-bottom: 0;
+        }
+
+        .ez-modal-body__main:last-of-type {
```
sunpietro: Is there a possibility that there are many elements with `.ez-modal-body__main` inside `.modal-body`?

inakijv: Right now it's two sentences only, but it might be changed when translated or adapted when extending it. How do you see it?

sunpietro: This class should be added to only one element inside the modal. Because it's meant to be used with the main content of the modal. I can't imagine someone is using it twice, because it will be structuraly incorrect.

inakijv:
Updated based on review. Ping @sunpietro

PR #391, opened 2018-03-14
EZP-28334: Fix for restoring from trash after removing structure of items.

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28334
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   |no

When a Parent item is sent to Trash - then all the corresponding child items will no longer have a parent location ID and each of these child content item's should display a status/message 'Item's ancestors are in Trash' under the path column.

And whenever a user tries to restore the child item then the option to restore is disabled and only the option to restore under a new parent is enabled.

screen shot 2018-03-14 at 9 07 19 am

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/Form/Data/TrashItemData.php
+++ b/src/lib/Form/Data/TrashItemData.php
@@ -91,6 +91,8 @@ class TrashItemData
 
     public function isParentInTrash(): bool
     {
-        return !end($this->ancestors) instanceof Location;
+        $lastAncestor = end($this->ancestors);
+
+        return $this->location->path !== array_merge($lastAncestor->path, [(string)$this->location->id]);
     }
 }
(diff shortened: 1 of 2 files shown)


lserwatka:
@mikadamczyk could you re-base to check if Unit Tests are passing?

mikadamczyk:
ping @Nattfarinn @webhdx

src/lib/Form/Data/TrashItemData.php
```diff
@@ -91,6 +91,8 @@ public function setAncestors(array $ancestors)
 
     public function isParentInTrash(): bool
     {
-        return !end($this->ancestors) instanceof Location;
+        $lastAncestor = end($this->ancestors);
+
+        return $this->location->pathString !== $lastAncestor->pathString . $this->location->id . '/';
```
webhdx: I don't think resolving this problem by string comparison is suggested resolution here. `$this->location->path` provides array of consecutive locations.  Comparing arrays is more reliable in general. cc @Nattfarinn

PR #395, opened 2018-03-15
EZP-28932: Broken empty option design after selecting it from ezselection fieldtype

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28932
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

### DIFF
--- a/src/bundle/Resources/public/scss/fieldType/edit/_ezselection.scss
+++ b/src/bundle/Resources/public/scss/fieldType/edit/_ezselection.scss
@@ -16,6 +16,9 @@
             position: relative;
             min-height: 2.8rem;
             background-color: $ez-white;
+            display: flex;
+            align-items: center;
+            flex-wrap: wrap;
 
             &:after {
                 content: '';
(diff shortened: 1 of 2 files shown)


sunpietro:
ping @dew326 @barbaragr 
Please, test with long text examples in the options labels.

barbaragr:
If I had two the same options, for example: `abc`, only one of them is show in creation form. Should we allow user to add identical options then?

sunpietro:
@barbaragr  I believe it's a question to PMs. Currently, it's out of scope of this PR.
cc @SylvainGuittard

SylvainGuittard:
Regarding identical options, I agree with @sunpietro: it's out of scope of this PR.
I think not allowing the user to use the identical labels for option is a good idea. I also don't think this will happen in a real use case. So, no need for a follow up story.

PR #392, opened 2018-03-14
EZP-28859: Bad request after filling one coordinate from ezmaplocation

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28859
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no
| License       | [GPL-2.0](https://github.com/ezsystems/ezplatform-admin-ui/blob/master/LICENSE)

In fact, this PR doesn't fix the backend issue. It improves the frontend validation before publishing the content.

PR #396, opened 2018-03-15
EZP-28859: Bad request after filling one coordinate from ezmaplocation - obsolete code removal

sunpietro:
The followup of #392 
Removed the dead code.

| Question      | Answer
| ------------- | ---
| Tickets       | N/A
| Bug fix?      | ?
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezgmaplocation.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezgmaplocation.js
@@ -71,12 +71,7 @@
             const invalidLatitude = allEmptyOrFilledResult.invalidInputType === POSITION_TYPE_LATITUDE;
 
             if (latResult.isError || (!isNativeEvent && invalidLatitude)) {
-                if (invalidLatitude && !latResult.isError) {
-                    latResult.isError = true;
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/scripts/fieldType/ezgmaplocation.js
```diff
@@ -71,12 +71,7 @@
             const invalidLatitude = allEmptyOrFilledResult.invalidInputType === POSITION_TYPE_LATITUDE;
 
             if (latResult.isError || (!isNativeEvent && invalidLatitude)) {
-                if (invalidLatitude && !latResult.isError) {
-                    latResult.isError = true;
-                    latResult.errorMessage = allEmptyOrFilledResult.errorMessage;
-                }
-
-                latInput.dispatchEvent(new Event('showLatitudeError'));
+                return false;
```
dew326: This should return the object with `isError = false`

sunpietro: Nope

PR #398, opened 2018-03-16
EZP-28957: Cannot open draft conflict modal when two jQuery are loaded

dew326:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28957
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.location.tab.js
+++ b/src/bundle/Resources/public/js/scripts/admin.location.tab.js
@@ -1,11 +1,8 @@
-(function () {
-    var url = document.location.toString();
-    if (url.match('#')) {
-        $('.ez-tabs a[href="#' + url.split('#')[1] + '"]').tab('show');
-    }
+(function (global, doc, $) {
+    $('.ez-tabs a[href="#' + global.location.hash.split('#')[1] + '"]').tab('show');
 
     // Change hash for page-reload
     $('.ez-tabs a').on('shown.bs.tab', function (e) {
-        window.location.hash = e.target.hash + '#tab';
+        global.location.hash = e.target.hash + '#tab';
     })
-})();
+})(window, document, window.jQuery);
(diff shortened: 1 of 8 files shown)


src/bundle/Resources/public/js/scripts/admin.location.tab.js
```diff
@@ -1,11 +1,12 @@
-(function () {
-    var url = document.location.toString();
+(function (global, doc, jQuery) {
+    const $ = jQuery;
+    var url = doc.location.toString();
```
sunpietro: `var`?

sunpietro: I don't think it's needed anyway. See comment below

src/bundle/Resources/public/js/scripts/admin.location.tab.js
```diff
@@ -1,11 +1,12 @@
...
     if (url.match('#')) {
```
sunpietro: `window.location.hash` returns `''` (empty string) if there's no hash and a string like `'#4'`if there's any.
So no need to stringify `window.location` and no need to do the match check.

src/bundle/Resources/public/js/scripts/admin.location.view.js
```diff
@@ -1,4 +1,5 @@
-(function (global, doc) {
+(function (global, doc, jQuery) {
```
sunpietro: Why not: `(function (global, doc, $) {`?

dew326:
ping @sunpietro

dew326:
ping @sunpietro please check the last commit

PR #402, opened 2018-03-19
EZP-28856: Toolbar from Richtext disappears

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28856
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no


sunpietro:
Looks like fixing one bug uncovered another bug, that is a consequence of the previous one. This one is way more tricky to fix.

sunpietro:
@barbaragr I think this PR can be merged. The bug you found is a different bug and it should have an individual ticket in JIRA.

cc @lserwatka @dew326

PR #409, opened 2018-03-20
EZP-28982: Modal Send to trash content update

inakijv:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-28982](https://jira.ez.no/browse/EZP-28982)
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

Most important aspects:
- Update content from Send to trash URL for consistency purposes;
- Update content  and color from Send to trash field type when in Content type editing view.

screen shot 2018-03-20 at 1 57 33 pm
screen shot 2018-03-20 at 2 01 52 pm

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_card.scss
+++ b/src/bundle/Resources/public/scss/_card.scss
@@ -17,7 +17,6 @@
     &__header {
         background-color: $ez-color-base-pale;
         color: $ez-black;
-        font-weight: 700;
         border-radius: .25rem .25rem 0 0;
         padding: .75rem 1.25rem .5rem 1.25rem;
 
--- a/src/bundle/Resources/views/admin/content_type/delete_confirmation_modal.html.twig
+++ b/src/bundle/Resources/views/admin/content_type/delete_confirmation_modal.html.twig
@@ -4,7 +4,6 @@
     <div class="modal-dialog" role="document">
         <div class="modal-content">
             <div class="modal-header">
-                <h5 class="modal-title">{{ 'content_type.modal.title'|trans|desc('Please confirm') }}</h5>
                 <button type="button" class="close" data-dismiss="modal" aria-label="Close">
                     <svg class="ez-icon ez-icon--medium" aria-hidden="true">
                         <use xlink:href="{{ asset('bundles/ezplatformadminui/img/ez-icons.svg') }}#discard"></use>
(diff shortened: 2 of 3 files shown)


inakijv:
Ping @sunpietro @dew326

dew326:
It not related to your changes but is it correct that the order in one modal is `delete` `cancel` and in the other `cancel` `delete`?

src/bundle/Resources/views/admin/content_type/edit.html.twig
```diff
@@ -44,7 +44,7 @@
     <section class="container mt-4 px-5">
         <div class="card ez-card">
             <div class="ez-card__header ez-card__header--secondary d-flex justify-content-between">
-                <div class="p-2">
+                <div class="title p-2">
```
dew326: I think `.title` is not present in bootstrap so you should use our class convention. This could something like: `.ez-card__title`

inakijv:
@dew326 I did not change it the order of the buttons because I was waiting for a new PR for buttons in modals. But I agree 100% with you, it's not the order and I can change it right away.

dew326:
@inakijv I was just asking. If there is incoming PR with buttons in modals I'm fine with doing it there. Up to you.

inakijv:
Updated based on review (including order of buttons)

screen shot 2018-03-22 at 9 49 18 am

src/bundle/Resources/public/scss/_card.scss
```diff
@@ -17,11 +17,10 @@
     &__header {
         background-color: $ez-color-base-pale;
         color: $ez-black;
-        font-weight: 700;
         border-radius: .25rem .25rem 0 0;
         padding: .75rem 1.25rem .5rem 1.25rem;
 
-        &--secondary {
```
dew326: This will have an impact on content type create, content type edit, role assignment create. Do we want it in all places?
What I mean is class `ez-card__header--secondary` is used in few places. Do we want to not style it anymore?

dew326: And I don't understand why we need this class `ez-card__title` and we cannot use `ez-card__header--secondary`? This should work the same as the previous solution, in content type edit.

inakijv: Updating it. I see what you mean. I am ok keeping it.

src/bundle/Resources/views/admin/content_type/edit.html.twig
```diff
@@ -44,7 +44,7 @@
     <section class="container mt-4 px-5">
         <div class="card ez-card">
             <div class="ez-card__header ez-card__header--secondary d-flex justify-content-between">
-                <div class="p-2">
+                <div class="ez-card__title p-2">
```
dew326: Still, why do we need this extra class? It has the same styling as `ez-card__header--secondary` (color and font-weight), this styling will be inherited from the parent.

inakijv: Update it based on review 👍

PR #423, opened 2018-03-23
EZP-27966: Modified Language should only display one language

webhdx:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-27966
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [ ] Ready for Code Review


ViniTou:
Aside from code, how is that to this:
https://jira.ez.no/browse/EZP-28671

webhdx:
We are supposed to only display one language as the column's label reads "Modified language". This is the known issue and I'm sure @SylvainGuittard https://jira.ez.no/browse/EZP-27966 already wanted to fix that.

src/lib/UI/Value/ValueFactory.php
```diff
@@ -105,6 +105,9 @@ public function createVersionInfo(VersionInfo $versionInfo): UIValue\Content\Ver
         return new UIValue\Content\VersionInfo($versionInfo, [
             'author' => $author,
             'translations' => $translationsDataset->getTranslations(),
+            'modifiedLanguage' => $versionInfo->initialLanguageCode
```
ViniTou: I think we decided to use `admin_ui_config.languages.map` as in previous PRs?

src/bundle/Resources/views/content/tab/versions/table.html.twig
```diff
@@ -32,9 +32,7 @@
                 {{ version.versionNo }}
             </td>
             <td>
-                {% for translation in version.translations %}
-                    {{ translation.name }}<br>
-                {% endfor %}
+                {{ version.modifiedLanguage.name }}
```
mikadamczyk: We can simply do `admin_ui_config.languages.map[version.initialLanguageCode]`

SylvainGuittard:
> This is the known issue and I'm sure @SylvainGuittard https://jira.ez.no/browse/EZP-27966 already wanted to fix that.
Thanks @webhdx for finding the related ticket :) 

> We are supposed to only display one language as the column's label reads "Modified language".
@ViniTou @webhdx Exactly, `Modified language` should only display one language. This is really important for the editors to quickly see which draft they want edit.

barbaragr:
According to this ticket: https://jira.ez.no/browse/EZP-28671 I've mentioned also that the problem occurs also in Dashboard.

webhdx:
@barbaragr can you briefly recheck with https://github.com/ezsystems/repository-forms/pull/228 ?

PR #429, opened 2018-03-28
EZP-28169 - Behat Covering of Sections

m-tyrala:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-28169](https://jira.ez.no/browse/EZP-28169)
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes/no
| Doc needed?   | no

13 tests that covers Admin -> Sections 

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/features/Sections.feature
+++ b/features/Sections.feature
@@ -1,0 +1,138 @@
+Feature: Sections management
+  As an administrator
+  In order to customize my eZ installation
+  I want to manage my content Sections.
+
+  Background:
+    Given I am logged as "admin"
+    And I go to "Sections" in "Admin" tab
+
+  @javascript @common
+  Scenario: Changes can be discarded while creating new Section
+    When I start creating new "Section"
+    And I set fields
+      | label      | value                  |
+      | Name       | Test Section           |
+      | Identifier | TestSectionIdentifier  |
+    And I click on the edit action bar button "Discard changes"
+    Then I should be on "Sections" page
+    And there's no "Test Section" on "Sections" list
+
+  @javascript @common
+  Scenario: New Section can be added
+    When I start creating new "Section"
+    And I set fields
+      | label      | value                  |
+      | Name       | Test Section           |
+      | Identifier | TestSectionIdentifier  |
+    And I click on the edit action bar button "Create"
+    Then I should be on "Section" "Test Section" page
+    And "Content items" list in "Section" "Test Section" is empty
+    And "Section" "Test Section" has proper attributes
+      | label      | value                 |
+      | Name       | Test Section          |
+      | Identifier | TestSectionIdentifier |
+
+  @javascript @common
+  Scenario: I can navigate to Admin / Sections through breadcrumb
+    Given I go to "Test Section" "Section" page
+    When I click on "Sections" on breadcrumb
+    Then I should be on "Sections" page
+
+  @javascript @common
+  Scenario: Content item assignation can be discarded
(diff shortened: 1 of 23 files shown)


src/lib/Behat/PageObject/SectionPage.php
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /**
+     * Verifies if list of Content items is empty.
+     *
+     * @param string $tabName
+     */
+    public function verifyListIsEmpty(): void
+    {
+        $firstRowValue = $this->adminLists['Content items']->table->getCellValue(1,1);
```
mnocon: I'd change it into two separate assertions:
Assert item count is equal to 1
Assert the message

src/lib/Behat/BusinessContext/AdministrationContext.php
```diff
@@ -176,18 +177,37 @@ public function iGoToListItem(string $itemName, string $itemType, string $itemCo
      */
     public function iStartEditingItem(string $itemType, string $itemName, ?string $containerName = null): void
```
mnocon: I think this one should be split into two separate steps:
One:
```
* @When I start editing :itemType :itemName		     
* @When I start editing :itemType :itemName from :containerName
```
and second
```
 
* @When I start editing :itemType :itemName from details page
```

src/lib/Behat/BusinessContext/AdministrationContext.php
```diff
@@ -176,18 +177,37 @@ public function iGoToListItem(string $itemName, string $itemType, string $itemCo
...
+            ->startEditingItem($itemName);
     }
 
     /**
-     * @When I delete :itemType :itemName
-     * @When I delete :itemType :itemName from :itemContainer
+     * @When I delete :itemType
+     * @When I delete :itemType from :containerName
      */
-    public function iDeleteItem(string $itemType, string $itemName, ?string $itemContainer = null): void
```
mnocon: Same here: I'd split it into 3 steps (I delete, I delete from container, I delete from details page). I don't think we need to have such complicated steps.

features/Sections.feature
```diff
@@ -0,0 +1,137 @@
...
+  @javascript @common
+  Scenario: Content item can be reassigned to section from the Sections details
+    Given I go to "Media" "Section" page
+    When I start assigning to "Media" from "Section" page
+      And I select content "Media/Images" through UDW
+      And I confirm UDW the selection
+    Then I should be on "Section" "Media" page
+      And content items list in section "Media" contains items
+      | Name   | Content Type | Path  |
+      | Images | Folder       | Media |
```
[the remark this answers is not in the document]
m-tyrala: Next scenario `Empty section can be deleted` serves as an assertion here. I don't really want to create `Then` steps that contains navigation actions.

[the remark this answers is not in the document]
mnocon: You're right that it's not pretty (I don't have any better idea). But still, for me it's better to have a clunky assertion in the same test which does the action than to have a hidden assertion in the next test (I didn't know that Sections behave like that and was really puzzled what happened to the assigned item 🤔 )

One potential reason: one day we may decide that the one of these tests is less important and we will run it less often - meaning that this assertion will also be executed less often.

m-tyrala:
@lserwatka this one can be merged

PR #432, opened 2018-03-29
EZP-28168 - Behat Covering of SystemInfo

m-tyrala:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-28168](https://jira.ez.no/browse/EZP-28168)
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes/no
| Doc needed?   | no

5 tests that covers Admin -> System Information

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/lib/Behat/PageObject/SystemInfoPage.php
+++ b/src/lib/Behat/PageObject/SystemInfoPage.php
@@ -1,0 +1,76 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformAdminUi\Behat\PageObject;
+
+
+use EzSystems\EzPlatformAdminUi\Behat\Helper\UtilityContext;
+use EzSystems\EzPlatformAdminUi\Behat\PageElement\AdminList;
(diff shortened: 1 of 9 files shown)


src/lib/Behat/PageObject/SystemInfoPage.php
```diff
@@ -0,0 +1,76 @@
...
+    public function verifySystemInfoRecords(string $tableName, array $records): void
+    {
+        $this->adminLists[$tableName]->verifyVisibility();
+        $tableHash = $this->adminLists[$tableName]->table->getTableHash();
+
+        foreach($records as $desiredRecord) {
+            $found = false;
+            foreach ($tableHash as $actualRecord) {
+                if($desiredRecord['Name'] === $actualRecord['Name']) {
+                    $found = true;
```
mnocon: You can `break` once `found` is true.
Or it can be done in a slightly different manner:
intersect the $desired array with $actual (so only the ones that are present in both are left) and then check if the result is the same as $desired (it should be).
See https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/lib/Behat/Helper/EnvironmentRestore.php#L90

m-tyrala:
Yeah, after Section merge this PR should be rebased.

m-tyrala:
@lserwatka this one can be merged.

PR #433, opened 2018-03-29
EZP-29046: As an Editor, I want to have "Content field definitions" header sticky to the top

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29046
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

screen shot 2018-03-29 at 2 10 49 pm

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_card.scss
+++ b/src/bundle/Resources/public/scss/_card.scss
@@ -24,6 +24,12 @@
             color: $ez-secondary-ground-medium;
             font-weight: 400;
         }
+
+        &--sticky-top {
+            position: sticky;
+            top: 0;
+            z-index: 9999;
+        }
     }
 
     &__body {
--- a/src/bundle/Resources/views/admin/content_type/edit.html.twig
+++ b/src/bundle/Resources/views/admin/content_type/edit.html.twig
@@ -43,7 +43,7 @@
 
     <section class="container mt-4 px-5">
         <div class="card ez-card">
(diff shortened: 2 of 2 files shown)


inakijv:
Thank you @mikadamczyk for this improvement. This way the most important action buttons will be always available for the user.

src/bundle/Resources/views/admin/content_type/edit.html.twig
```diff
@@ -43,7 +43,7 @@
 
     <section class="container mt-4 px-5">
         <div class="card ez-card">
-            <div class="ez-card__header ez-card__header--secondary d-flex justify-content-between">
+            <div class="ez-card__header ez-card__header--sticky-top ez-card__header--secondary d-flex justify-content-between" style="position: sticky; top: 0; z-index: 9999;">
```
dew326: The `style="...` must go. We do not use inline styling.

src/bundle/Resources/public/scss/_card.scss
```diff
@@ -28,7 +29,19 @@
...
             top: 0;
+            padding: .25rem 1.25rem;
+            justify-content: space-between;
             z-index: 9999;
+
+            .form-inline {
+                .btn-danger {
+                    .ez-icon {
+                        width: 1.5rem;
+                        height: 1.5rem;
```
dew326: You can use class `ez-icon--medium` for the `width` and `height`

PR #434, opened 2018-03-30
EZEE-1610: Ignore Language Picker for Content with one Language

mnocon:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZEE-1610
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

This change is for 1.1 - since 2.1 release, the Language Picker is not visible if Content has only one language.

This is required by: https://github.com/ezsystems/flex-workflow/pull/97/files

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/Behat/BusinessContext/ContentViewContext.php
+++ b/src/lib/Behat/BusinessContext/ContentViewContext.php
@@ -13,6 +13,7 @@ use EzSystems\EzPlatformAdminUi\Behat\PageElement\RightMenu;
 use EzSystems\EzPlatformAdminUi\Behat\PageElement\UniversalDiscoveryWidget;
 use EzSystems\EzPlatformAdminUi\Behat\PageObject\ContentStructurePage;
 use EzSystems\EzPlatformAdminUi\Behat\PageObject\PageObjectFactory;
+use EzSystems\EzPlatformAdminUi\UI\Value\Content\Language;
 use PHPUnit\Framework\Assert;
 
 class ContentViewContext extends BusinessContext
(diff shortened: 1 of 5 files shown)


src/lib/Behat/BusinessContext/ContentViewContext.php
```diff
@@ -42,15 +42,19 @@ public function startCreatingArticle(string $name): void
...
     {
         $rightMenu = new RightMenu($this->utilityContext);
         $rightMenu->clickButton('Edit');
 
         $languagePicker = new LanguagePicker($this->utilityContext);
-        $languagePicker->chooseLanguage($language);
+
+        if ($languagePicker->isVisible()) {
+            $languagePicker->chooseLanguage($language);
+        }
```
m-tyrala: Have you checked, how it's going to behave if there'll be a bug and language picker appears when it shouldn't? 
`chooseLanguage` function will throw an error, or will choose first available language? If it's second case we probably wont detect such bug.

mnocon: Good question, I can add an additional assertion here, something like:
if a picker is visible then check if there is more than one language to pick from, what do you think?

m-tyrala: I think that is a good solution.

mnocon:
@lserwatka this one can be merged too

PR #437, opened 2018-04-03
EZP-29049: As a user, I want to delete a single item in the trash.

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29049
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

There is no way to delete a single item (or multiple items) in the trash. There is only one button: Empty trash.

screen shot 2018-04-03 at 3 39 35 pm

screen shot 2018-04-03 at 3 36 36 pm

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/TrashController.php
+++ b/src/bundle/Controller/TrashController.php
@@ -4,6 +4,8 @@
  * @copyright Copyright (C) eZ Systems AS. All rights reserved.
  * @license For full copyright and license information view LICENSE file distributed with this source code.
  */
+declare(strict_types=1);
+
 namespace EzSystems\EzPlatformAdminUiBundle\Controller;
 
 use eZ\Publish\API\Repository\ContentService;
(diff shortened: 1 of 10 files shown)


src/bundle/Resources/public/js/scripts/admin.trash.list.js
```diff
@@ -31,15 +31,20 @@
...
     const buttonRestore = document.querySelector('#trash_item_restore_restore');
     const buttonRestoreUnderNewParent = document.querySelector('#trash_item_restore_location_select_content');
+    const buttonDelete = document.querySelector('#delete-trash-items');
 
     const enableButtons = (event) => {
+        const deleteCheckbox = document.querySelector('form[name="trash_item_delete"] input[type="checkbox"][value="' + event.target.value + '"]');
         const isNonEmptySelection = checkboxes.some(el => el.checked);
         const isMissingParent = checkboxes.some(el => el.checked && parseInt(el.dataset.isParentInTrash, 10) === 1);
 
+        if (deleteCheckbox !== null) {
```
dew326: `if (deleteCheckbox) {` is enough.

src/bundle/Resources/views/admin/trash/list.html.twig
```diff
@@ -83,6 +97,18 @@
                     {{ form_widget(form_trash_item_restore._token) }}
                     {{ form_end(form_trash_item_restore, { 'render_rest': false }) }}
                 </div>
+
+                {{ form_start(form_trash_item_delete, {
+                    'action': path('ezplatform.trash.delete')
+                }) }}
+                {% for key, form in form_trash_item_delete.trash_items %}
+                {% set trash_item = trash_items[key] %}
+                        {{ form_widget(form) }}
```
dew326: indentation

src/bundle/Controller/TrashController.php
```diff
@@ -240,4 +266,48 @@ public function restoreAction(Request $request): Response
...
+     *
+     * @throws \Symfony\Component\OptionsResolver\Exception\InvalidOptionsException
+     * @throws \Symfony\Component\Translation\Exception\InvalidArgumentException
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
+     * @throws \LogicException
+     * @throws \InvalidArgumentException
+     */
+    public function deleteAction(Request $request): Response
+    {
+        if ($this->isGranted(new Attribute('content', 'remove'))) {
```
webhdx: Use early return/redirection in this case.

webhdx: Use early return/redirection in this case.

PR #439, opened 2018-04-05
EZP-29065: As an system administrator I would to collapse fieldtype f…

sunpietro:
…orms in the content type edit/create views

| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29065
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no


sunpietro:
cc @inakijv @SylvainGuittard

SylvainGuittard:
Thanks @sunpietro for this improvement!
Regarding the design, could you please change the color of the button and the orientation of the icon when collapsed?
image

Is it possible after adding a field to open this specific field?

sunpietro:
@SylvainGuittard regarding the colors and icons: I'll do it
But in terms of displaying the fieldtype form visible after adding it, then it will require more wokr on the backend side. This fix is purely frontend fix. Maybe we should create a followup for someone from the backend devs.

SylvainGuittard:
@sunpietro I will create this follow up story :)
Thanks

wizhippo:
Nice:)  How about drag and drop ordering?

SylvainGuittard:
@sunpietro Sorry for bothering you again about design :)
Why do you have a different background color when the field is visible? Is it because the button status is `active`?

lserwatka:
I think the bigger arrow like @SylvainGuittard made it in the mockup looks better.

sunpietro:
@SylvainGuittard yes, it's because it's in focus state. I can remove it.

lserwatka:
@m-tyrala could you update Behat test here, according to this change?

SylvainGuittard:
@sunpietro You can keep it. We'll see with @inakijv later.
@lserwatka is right. It looks better when the icon is a little bit bigger :)

sunpietro:
The size of an icon has been increased.

src/bundle/Resources/views/admin/content_type/create.html.twig
```diff
@@ -70,12 +70,20 @@
...
 
-                    <div class="card ez-card ez-card--fieldtype-container">
+                    <div class="card ez-card ez-card--fieldtype-container mb-3 ez-card--collapsed">
                         <div id="field-definition-{{ value.identifier }}" class="ez-card__header ez-card__header--secondary">
                             {{ form_widget(field_definition.selected, {
                                 'label': '%s (%s)' | format(value.names[language_code], value.fieldTypeIdentifier),
                                 'label_attr': {class: 'checkbox-inline'}
                             }) }}
+                            <button class="ez-card__body-display-toggler btn">
+                                <svg class="ez-icon ez-icon--medium ez-icon--caret-down" aria-hidden="true">
```
dew326: I think the `aria-hidden` is not correctly set. Both are hidden? If you want to use `aria-hidden` you should also change the state of it when user clicks on the button and change the state.

sunpietro: Well, I disagree. By using any `aria-*` attributes you're informing screen readers about the element purpose. In this case, when using `aria-hidden` it doesn't mean it's visually hidden. It means that screen readers can omit this piece of HTML code and process the next HTML nodes.

dew326: According to the W3C recommendations: 
```
aria-hidden (state)§

Indicates whether the element is exposed to an accessibility API. See related aria-disabled.

User agents determine an element's hidden status based on whether it is rendered, and the rendering is usually controlled by CSS. For example, an element whose display property is set to none is not rendered. An element is considered hidden if it, or any of its ancestors are not rendered or have their aria-hidden attribute value set to true.
```
ref: https://www.w3.org/TR/wai-aria-1.1/#aria-hidden

So as I understand this if you set `aria-hidden=true` the screen reader will omit this element.

sunpietro: Yes, it hidden for screen readers only.

dew326: So on a screen reader, you will not be able to expand the card?

sunpietro: It's a screen reader, it won't try to read the code of the icon. The button will still work. If we try to increase the accessibility of our interface maybe we should start applying correct `aria` attributes here and there in our codebase.
In this case, it just hides an icon from reading it.

src/lib/Behat/PageElement/Notification.php
```diff
@@ -21,6 +22,7 @@ public function __construct(UtilityContext $context)
             'alert' => '.alert',
             'successAlert' => '.alert-success',
             'failureAlert' => '.alert-danger',
+            'closeAlert' => '.alert-success.show button',
```
dew326: I have some suggestions to make it more bulletproof. 
`closeAlert` do not match to the class `.alert-success` - this should be `closeSuccessAlert` or use selector '.alert'.
The `button` is to generic, there might be at some point more buttons in alert, I would rather use `button.close` selector.

barbaragr:
@lserwatka I'm working on this :)

PR #443, opened 2018-04-13
EZP-29066: As a website administrator, I want to see the roles assigned to a user or user group

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29066
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

As a website administrator, I want to see the roles assigned to a user or user group

screen shot 2018-04-13 at 4 04 09 pm

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/Tab/ConditionalTabInterface.php
+++ b/src/lib/Tab/ConditionalTabInterface.php
@@ -1,0 +1,25 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformAdminUi\Tab;
+
+/**
+ * Conditional Tab interface needs to be implemented by tabs,
+ * which needs to be evaluate depends on context.
+ */
+interface ConditionalTabInterface
+{
+    /**
+     * Get information about tab presence.
+     *
+     * @param array $parameters
+     *
+     * @return bool
+     */
+    public function evaluate(array $parameters): bool;
+}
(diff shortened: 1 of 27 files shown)


SylvainGuittard:
Good job @mikadamczyk !
I like it 🎉

src/lib/Tab/ConditionalTabInterface.php
```diff
@@ -0,0 +1,25 @@
...
+interface ConditionalTabInterface
+{
+    /**
+     * Get information about tab presence.
+     *
+     * @param array $parameters
+     *
+     * @return bool
+     */
+    public function getPresence(array $parameters): bool;
```
webhdx: Proposition: `isVisible()`?

Nattfarinn: To hide or display is not responsibility of this interface. `evaluate` would be more proper name here.

src/lib/Tab/Event/Subscriber/ConditionalTabSubscriber.php
```diff
@@ -0,0 +1,77 @@
...
+     * @param TabInterface[] $tabs
+     * @param array $parameters
+     *
+     * @return array
+     */
+    private function processTabs($tabs, $parameters): array
+    {
+        foreach ($tabs as $tab) {
+            if ($tab instanceof ConditionalTabInterface && !$tab->evaluate($parameters)) {
+                unset($tabs[$tab->getIdentifier()]);
```
Nattfarinn: As we have `TAB_GROUP_INITIALIZE` now, I would not go this way. I would reverse this logic as it make much more sense to add tabs if they match condition instead of removing tabs which don't.

I would remove this `processTabs` method and move foreach to `onTabGroupInitialize` with something like this:

```
foreach ($tabs as $tab) {
    if ($tab instanceof ConditionalTabInterface && $tab->evaluate($parameters)) {
        $tabGroup->addTab($tab);
    }
}
```

barbaragr:
follow-up https://jira.ez.no/browse/EZP-29113

PR #447, opened 2018-04-18
EZP-29067: As a website administrator, I want to see the policies assigned to a user or user group

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29067
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

screen shot 2018-04-18 at 2 34 48 pm

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/content/tab/policies/limitation_values.html.twig
+++ b/src/bundle/Resources/views/content/tab/policies/limitation_values.html.twig
@@ -1,5 +1,17 @@
 {% extends '@EzSystemsRepositoryForms/limitation_values.html.twig' %}
 
+{% block ez_limitation_node_value %}
+    {% spaceless %}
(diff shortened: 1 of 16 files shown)


src/bundle/Resources/views/content/tab/policies/limitation_values.html.twig
```diff
@@ -0,0 +1,124 @@
+{# fallback block for limitations without defined value mapper #}
+{% block ez_limitation_value_fallback %}
```
webhdx: Do you add anything particular to this template? Extending the original file from repository-forms would make more sense because it will make it safe after something was added there.

PR #449, opened 2018-04-19
EZP-28806: As a developer I want to define the SiteAccess name

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28806
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Tests pass?   | yes
| Doc needed?   | yes

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/content/content_preview.html.twig
+++ b/src/bundle/Resources/views/content/content_preview.html.twig
@@ -42,7 +42,7 @@
                 <div class="ez-preview__item ez-preview__item--siteaccess">
                     <select class="form-control">
                         {% for previewSiteaccess in siteaccesses %}
-                            <option value="{{ path('_ezpublishPreviewContent', {'contentId': content.id, 'versionNo': versionNo, 'language': language_code, 'siteAccessName': previewSiteaccess}) }}">{{ previewSiteaccess }}</option>
+                            <option value="{{ path('_ezpublishPreviewContent', {'contentId': content.id, 'versionNo': versionNo, 'language': language_code, 'siteAccessName': previewSiteaccess}) }}">{{ previewSiteaccess|trans({}, 'ezplatform_siteaccess') ~ ' (' ~ previewSiteaccess ~ ')' }}</option>
                         {% endfor %}
                     </select>
                 </div>


src/bundle/Templating/Twig/SiteAccessNameExtension.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        return [
+            new Twig_SimpleFunction(
+                'ezplatform_admin_ui_site_access_name',
+                [$this, 'getSiteAccessName']
+            ),
+        ];
+    }
+
+    public function getSiteAccessName(
```
mikadamczyk: Missing docblock

src/bundle/Resources/views/content/content_preview.html.twig
```diff
@@ -42,7 +42,7 @@
                 <div class="ez-preview__item ez-preview__item--siteaccess">
                     <select class="form-control">
                         {% for previewSiteaccess in siteaccesses %}
-                            <option value="{{ path('_ezpublishPreviewContent', {'contentId': content.id, 'versionNo': versionNo, 'language': language_code, 'siteAccessName': previewSiteaccess}) }}">{{ previewSiteaccess }}</option>
+                            <option value="{{ path('_ezpublishPreviewContent', {'contentId': content.id, 'versionNo': versionNo, 'language': language_code, 'siteAccessName': previewSiteaccess}) }}">{{ ezplatform_admin_ui_site_access_name(previewSiteaccess) ~ ' (' ~ previewSiteaccess ~ ')' }}</option>
```
webhdx: I don't think the label should have format: `Readable name (identifier)` but simply `Readable name`. Can you double check that with wireframes?

src/bundle/Templating/Twig/SiteAccessNameExtension.php
```diff
@@ -0,0 +1,49 @@
...
+    public function __construct(TranslatorInterface $translator)
+    {
+        $this->translator = $translator;
+    }
+
+    public function getFunctions()
+    {
+        return [
+            new Twig_SimpleFunction(
+                'ezplatform_admin_ui_site_access_name',
```
webhdx: You don't have to prefix twig function with `ezplatform_admin_ui`, simple `siteaccess_label(identifier)` is enough. 

Also please note it's called `siteaccess` not `site access`. This also applies to the method names where you are using `siteAccess` term.

src/bundle/Resources/views/content/content_preview.html.twig
ViniTou: Based on:
https://jira.ez.no/browse/EZP-28806?focusedCommentId=222740&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-222740

src/bundle/Resources/views/content/content_preview.html.twig
```diff
@@ -42,7 +42,7 @@
                 <div class="ez-preview__item ez-preview__item--siteaccess">
                     <select class="form-control">
                         {% for previewSiteaccess in siteaccesses %}
-                            <option value="{{ path('_ezpublishPreviewContent', {'contentId': content.id, 'versionNo': versionNo, 'language': language_code, 'siteAccessName': previewSiteaccess}) }}">{{ previewSiteaccess }}</option>
+                            <option value="{{ path('_ezpublishPreviewContent', {'contentId': content.id, 'versionNo': versionNo, 'language': language_code, 'siteAccessName': previewSiteaccess}) }}">{{ siteaccess_label(previewSiteaccess) ~ ' (' ~ previewSiteaccess ~ ')' }}</option>
```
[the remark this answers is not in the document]
Nattfarinn: @lserwatka @webhdx I would use just: `previewSiteaccess|trans({}, 'ezplatform_siteaccess')` instead of TWIG function for now in that case. Better to implement twig helper directly in new Core bundle to avoid BC break instead of having two or one of them deprecated or something.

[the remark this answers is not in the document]
SylvainGuittard: Indeed a TWIG helper will be very useful for frontend developer. Definitely something to add to core bundle.
ping @bdunogier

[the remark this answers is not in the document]
ViniTou: Well, as I told during demo this will go into new separate package.
@lserwatka

DominikaK:
@bdunogier doc is not done yet, we usually do it when the JIRA ticket falls into our lap and link there, but we can also try adding links here if that makes things easier.

bdunogier:
No problem @DominikaK. I think that it helps, yes. As a bonus, if in the doc PR you reference the PR that introduced the documented feature, the link will show up both ways (doc <-> code).

PR #451, opened 2018-04-24
EZP-28479: In the content item edit view, the application does not display the help text when I hover over the input field

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28479
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/content/form_fields.html.twig
+++ b/src/bundle/Resources/views/content/form_fields.html.twig
@@ -82,7 +82,12 @@
     {% set attr = attr|merge({'class': (attr.class|default('') ~ ' ez-data-source__input')|trim}) %}
     {% set wrapper_attr = wrapper_attr|default({})|merge({'class': (wrapper_attr.class|default('') ~ ' ' ~ wrapper_class)|trim}) %}
 
-    <div{% with { attr: wrapper_attr } %}{{ block('attributes') }}{% endwith %}>
+    {% set field_type_descriptions = fieldtype.vars.value.fieldDefinition.descriptions %}
+    {% if field_type_descriptions[fieldtype.vars.languageCode] is defined %}
+        {% set wrapper_attr = wrapper_attr|merge({'title': field_type_descriptions[fieldtype.vars.languageCode]}) %}
+    {% endif %}
+
+    <div {% with { attr: wrapper_attr } %}{{ block('attributes') }}{% endwith %}>
         <div{% with { attr: label_wrapper_attr } %}{{ block('attributes') }}{% endwith %}>
             {{ block('form_label') }}
             {{ block('form_errors') }}


src/bundle/Resources/views/content/form_fields.html.twig
```diff
@@ -82,7 +82,12 @@
     {% set attr = attr|merge({'class': (attr.class|default('') ~ ' ez-data-source__input')|trim}) %}
     {% set wrapper_attr = wrapper_attr|default({})|merge({'class': (wrapper_attr.class|default('') ~ ' ' ~ wrapper_class)|trim}) %}
 
-    <div{% with { attr: wrapper_attr } %}{{ block('attributes') }}{% endwith %}>
+    {% if fieldtype.vars.value.fieldDefinition.descriptions[fieldtype.vars.languageCode] is defined %}
```
adamwojs: For me it looks more readable after extracting `field type.vars.value.field Definition.descriptions` to the local variable:

```twig
{% set field_type_descriptions = fieldtype.vars.value.fieldDefinition.descriptions %}
{% if field_type_descriptions[fieldtype.vars.languageCode] is defined %}
    {% set wrapper_attr = wrapper_attr|merge({'title': field_type_descriptions[fieldtype.vars.languageCode]}) %}
{% endif %}
```

lserwatka: @ViniTou could you incorporate this feedback?

ViniTou: @lserwatka 
done

micszo:
Using Article as an example everything's fine except for the RichText fields (Summary & Body).
Instead of the set descriptions tooltip is displayed for both as on screenshot below.
screen shot 2018-05-02 at 10 55 48 2

micszo:
@ViniTou wdyt?

ViniTou:
@micszo 
I think it is rendered in separate view, gonna check it out. It would probably require separate PR to new richTextBundle.

ViniTou:
Ok, that comes directly from alloy editor, is there any way we can override this? @sunpietro

sunpietro:
@ViniTou do you mean the tooltip? It might be. But I'm not sure how we can improve the styling. I'll have to investigate it, but as far as I know it might require some effort to style it.
@dew326 might have some opinions based on his actual experience with AlloyEditor.

dew326:
@sunpietro @ViniTou @micszo  The AlloyEditor takes the `id` of the editable element: https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/bundle/Resources/views/fieldtypes/edit/ezrichtext.html.twig#L4

One way would be to change the `id`, but ids usually are not human readable. The other option is to check AlloyEditor/CKEditor and replace the method which is adding the title. 

For me in order to make this right, we need the following ticket and dig more into AlloyEditor/CKEditor.

webhdx:
How does it relate to the wireframes?
According to those: 
image
The description should be displayed below.

micszo:
@webhdx The way I understood the ticket the first paragraph related to the bug and the second one to a future Improvement/Story.
And also in the CS ticket this came from a custom template was provided to the customer which displays it differently.

lserwatka:
@ViniTou could you re-base here to trigger Behat?

PR #459, opened 2018-04-30
EZP-29023: Setting limitations while creating policies

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29023
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

Currently, after creating/adding a new policy to a role it is instantly active with no limitations. They could be added by editing this policy. The idea is to specify limitations while creating a new policy by setting this as next step to follow when policy can have limitations.

screen shot 2018-04-30 at 1 37 40 pm

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/features/Roles.feature
+++ b/features/Roles.feature
@@ -78,7 +78,7 @@ Feature: Roles management
       And I click on the edit action bar button "Save"
     Then I should be on "Role" "Test Role edited" page
       And "Policies" list in "Role" "Test Role edited" is empty
-      And There are assignments on the "Test Role edited" assignments list
+      And there are assignments on the "Test Role edited" assignments list
       | user/group          | limitation                         |
       | Administrator User  | Subtree of Location: /Media/Images |
       | Anonymous User      | Subtree of Location: /Media/Images |
@@ -93,7 +93,7 @@ Feature: Roles management
       And I click on the edit action bar button "Save"
     Then I should be on "Role" "Test Role edited" page
       And "Policies" list in "Role" "Test Role edited" is empty
-      And There are assignments on the "Test Role edited" assignments list
+      And there are assignments on the "Test Role edited" assignments list
       | user/group          | limitation                         |
       | Administrator User  | Subtree of Location: /Media/Images |
       | Editors             | Subtree of Location: /Media/Images |
@@ -111,7 +111,7 @@ Feature: Roles management
       | Users	            |
     Then I should be on "Role" "Test Role edited" page
       And "Policies" list in "Role" "Test Role edited" is empty
-      And There's an assignment "Subtree of Location: /Media/Images" for "Anonymous User" on the "Test Role edited" assignments list
+      And there is an assignment "Subtree of Location: /Media/Images" for "Anonymous User" on the "Test Role edited" assignments list
 
   @javascript @common
   Scenario: Adding policy can be discarded
@@ -122,7 +122,7 @@ Feature: Roles management
       And I click on the edit action bar button "Discard changes"
     Then I should be on "Role" "Test Role edited" page
       And "Policies" list in "Role" "Test Role edited" is empty
-      And There's an assignment "Subtree of Location: /Media/Images" for "Anonymous User" on the "Test Role edited" assignments list
+      And there is an assignment "Subtree of Location: /Media/Images" for "Anonymous User" on the "Test Role edited" assignments list
 
   @javascript @common
   Scenario: Policies can be added to role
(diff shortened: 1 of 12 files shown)


mikadamczyk:
@m-tyrala could you take a look on Behat tests?

src/lib/Behat/BusinessContext/RolesContext.php
```diff
@@ -106,12 +106,39 @@ public function thereIsAPolicy(string $moduleAndFunction, string $limitation, st
...
+        $adminList = $rolePage->adminLists[$this->tabMapping['policy']];
+        $actualPoliciesList = $adminList->table->getTableHash();
+        $policyExists = false;
+        $expectedModule = explode('/', $moduleAndFunction)[0];
+        $expectedFunction = explode('/', $moduleAndFunction)[1];
+        foreach ($actualPoliciesList as $policy) {
+            if (
+                $policy['Module'] === $expectedModule &&
+                $policy['Function'] === $expectedFunction &&
+                strpos($policy['Limitations'], $limitation) !== false
```
ViniTou: yoda conditional

src/lib/Behat/BusinessContext/RolesContext.php
```diff
@@ -99,19 +99,46 @@ public function thereIsAPolicy(string $moduleAndFunction, string $limitation, st
...
 
     /**
-     * @Then There's an assignment :limitation for :userOrGroup on the :roleName assignments list
+     * @Then there is no policy :moduleAndFunction with :limitation limitation on the :roleName policies list
+     */
+    public function thereIsNoPolicy(string $moduleAndFunction, string $limitation, string $roleName): void
+    {
+        $rolePage = PageObjectFactory::createPage($this->utilityContext, RolePage::PAGE_NAME, $roleName);
+        $rolePage->navLinkTabs->goToTab($this->tabMapping['policy']);
+        $adminList = $rolePage->adminLists[$this->tabMapping['policy']];
```
mnocon: Can we extract it to a separate method? Something like `isRoleWithLimitationPresent` in RolePage? Same thing is used in Step above.

features/Roles.feature
```diff
@@ -122,7 +122,7 @@ Feature: Roles management
       And I click on the edit action bar button "Discard changes"
     Then I should be on "Role" "Test Role edited" page
       And "Policies" list in "Role" "Test Role edited" is empty
-      And There's an assignment "Subtree of Location: /Media/Images" for "Anonymous User" on the "Test Role edited" assignments list
+      And there is an assignment "Subtree of Location: /Media/Images" for "Anonymous User" on the "Test Role edited" assignments list
 
   @javascript @common
   Scenario: Policies can be added to role
```
mnocon: I'd change the name to something like "Limitations can be specified when creating a Role"

m-tyrala: I agree, but maybe more like "Policies with Limitations can be added to role", to be more specific?

mikadamczyk:
@andrerom for now this is a two step process. In the first step we only add Policy to role and publish it. In step two (if policy can have limitation) we fetch limitations by module and function, create a form and allow to save policy. I think it was designed that way to avoid possibility to have unpublished draft role, nested try-catch blocks like in v1, race conditions or locks on role draft. But maybe we must redesign it.

andrerom:
@mikadamczyk Well, if user intends to add limitations, and if system does not allow him to set those before first publishing w/o limitations => then we have a security issue.

lserwatka:
@mnocon we will need a re-test after last code changes.

mikadamczyk:
ping @webhdx @Nattfarinn

src/lib/Behat/PageObject/RolePage.php
```diff
@@ -104,4 +104,37 @@ public function startAssigningToItem(string $roleName): void
...
+    public function isRoleWithLimitationPresent(string $listName, string $moduleAndFunction, string $limitation): bool
+    {
+        $policyExists = false;
+
+        $this->navLinkTabs->goToTab($listName);
+        $adminList = $this->adminLists[$listName];
+        $actualPoliciesList = $adminList->table->getTableHash();
+
+        $expectedModule = explode('/', $moduleAndFunction)[0];
+        $expectedFunction = explode('/', $moduleAndFunction)[1];
```
adamwojs: This could be simplified to: 

```php
list($expectedModule, $expectedFunction) = explode('/', $moduleAndFunction);
```

but both versions make an assumption that '/' is always present in the `$moduleAndFunction` string 😉

src/lib/Behat/PageObject/RolePage.php
```diff
@@ -104,4 +104,37 @@ public function startAssigningToItem(string $roleName): void
...
+
+        foreach ($actualPoliciesList as $policy) {
+            if (
+                $policy['Module'] === $expectedModule &&
+                $policy['Function'] === $expectedFunction &&
+                false !== strpos($policy['Limitations'], $limitation)
+            ) {
+                $policyExists = true;
```
adamwojs: We don't need to check next elements of`$actualPoliciesList`

src/lib/Behat/PageObject/RolePage.php
webhdx: I'm against `list()` - explicit variable declaration is more readable and it easier to understand for ~stupid~ less intelligent IDEs.

mikadamczyk:
ping @Nattfarinn

lserwatka:
@micszo we need a round of QA here?

PR #468, opened 2018-05-14
Fix rendering custom fieldgroup names

slaci:
| Question      | Answer
| ------------- | ---
| Tickets       | N/A
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

If I specify a random `fieldGroup` for a content type's field, then the admin can only display the group if I define the group in the symfony config: `ezpublish.repositories.default.fields_groups`.

Otherwise I get a fatal error when viewing a content in the admin:
ezfatal

 In 1.x I didn't have to specify the group in the config, so there was a BC break somewhere I think. With my change, the `fieldGroup` string (that I randomly defined at content type creation) is the fallback for the name, if translation or definition is missing (I would call this the old behavior).

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/Tab/LocationView/ContentTab.php
+++ b/src/lib/Tab/LocationView/ContentTab.php
@@ -110,6 +110,7 @@ class ContentTab extends AbstractTab implements OrderedTabInterface
             }
 
             $fieldDefinitionsByGroup[$groupId]['fieldDefinitions'][] = $fieldDefinition;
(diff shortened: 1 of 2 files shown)


slaci:
@lserwatka I wrote a test for this feature, I will try to rebase and create jira issue a bit later.

lserwatka:
@slaci this is great, thank you.

slaci:
@lserwatka I found a jira issue for this bug so I referenced it, but I had some problems with rebasing so I just created a new PR instead: https://github.com/ezsystems/ezplatform-admin-ui/pull/474

PR #469, opened 2018-05-14
EZP-29103: Implement permissions for Object State in Admin Panel

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29103
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | yes/no
| Tests pass?   | yes/no
| Doc needed?   | yes/no

For now state/assign policy is only checked by `$this->permissionResolver->hasAccess('state', 'assign')` method. When user has some limitation, button will be displayed.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/admin/object_state/list.html.twig
+++ b/src/bundle/Resources/views/admin/object_state/list.html.twig
@@ -7,27 +7,35 @@
         <div class="ez-table-header">
             <div class="ez-table-header__headline">{{ 'object_state.view.list.title'|trans({'%count%': object_states|length})|desc('Object States (%count%)') }}</div>
             <div>
-                <a title="{{ 'object_state.new'|trans|desc('Create a new Object State ') }}"
-                   href="{{ path('ezplatform.object_state.state.add', {'objectStateGroupId': object_state_group.id}) }}"
-                   class="btn btn-primary" {% if not can_administrate %} disabled="disabled"{% endif %}>
-                    <svg class="ez-icon ez-icon-create">
-                        <use xmlns:xlink="http://www.w3.org/1999/xlink"
-                             xlink:href="{{ asset('bundles/ezplatformadminui/img/ez-icons.svg') }}#create"></use>
-                    </svg>
-                </a>
-                {% set modal_data_target = 'delete-object-state-modal' %}
-                <button id="delete-object-state" type="button" class="btn btn-danger" disabled data-toggle="modal"
-                        data-target="#{{ modal_data_target }}" title="{{ 'object_state.delete.bulk_delete.submit'|trans|desc('Delete Object State') }}">
-                    <svg class="ez-icon ez-icon-trash">
-                        <use xmlns:xlink="http://www.w3.org/1999/xlink"
-                             xlink:href="{{ asset('bundles/ezplatformadminui/img/ez-icons.svg') }}#trash"></use>
-                    </svg>
-                </button>
-                {% include '@EzPlatformAdminUi/admin/bulk_delete_confirmation_modal.html.twig' with {
-                    'id': modal_data_target,
-                    'message': 'object_state.modal.message'|trans|desc('Do you want to delete Object State?'),
-                    'data_click': '#object_states_delete_delete',
-                }%}
+                {% if can_administrate %}
+                    <a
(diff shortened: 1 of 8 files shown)


src/bundle/Resources/views/admin/object_state/list.html.twig
```diff
@@ -7,27 +7,38 @@
...
-                    'message': 'object_state.modal.message'|trans|desc('Do you want to delete Object State?'),
-                    'data_click': '#object_states_delete_delete',
-                }%}
+                {% if can_administrate %}
+                    <a
+                        title="{{ 'object_state.new'|trans|desc('Create a new Object State ') }}"
+                        href="{{ path('ezplatform.object_state.state.add', {'objectStateGroupId': object_state_group.id}) }}"
+                        class="btn btn-primary">
+                        <svg class="ez-icon ez-icon-create">
+                            <use xmlns:xlink="http://www.w3.org/1999/xlink"
```
sunpietro: You can remove `xmlns:xlink` attribute from here. It's not needed.

src/bundle/Resources/views/admin/object_state/list.html.twig
```diff
@@ -7,27 +7,38 @@
...
+                        id="delete-object-state"
+                        type="button"
+                        class="btn btn-danger"
+                        disabled
+                        data-toggle="modal"
+                        data-target="#{{ modal_data_target }}"
+                        title="{{ 'object_state.delete.bulk_delete.submit'|trans|desc('Delete Object State') }}">
+                        <svg class="ez-icon ez-icon-trash">
+                            <use
+                                xmlns:xlink="http://www.w3.org/1999/xlink"
```
sunpietro: You can remove it

mikadamczyk:
@sunpietro fixed

src/lib/Tab/LocationView/DetailsTab.php
```diff
@@ -152,6 +151,7 @@ public function renderView(array $parameters): string
             'sort_field_clause_map' => $this->getSortFieldClauseMap(),
             'form_state_update' => $contentObjectStateUpdateTypeByGroupId,
             'can_see_section' => $can_see_section,
```
webhdx: Ugh, not your fault but... `$can_see_section` :( Not sure how we overlooked that. Can you fix this?

src/bundle/Resources/views/admin/object_state/list.html.twig
```diff
@@ -56,16 +64,21 @@
                 {% for object_state in object_states %}
                     <tr>
                         <td class="ez-checkbox-cell">
-                            {{ form_widget(form_states_delete.objectStates[object_state.id], {"disabled": not can_administrate}) }}
+                            {% if can_administrate %}
+                                {{ form_widget(form_states_delete.objectStates[object_state.id]) }}
+                            {% else %}
+                                {% do form_states_delete.objectStates.setRendered %}
```
webhdx: It raises my concerns - it is only marking a field as rendered which means you can prepare a form with any data you want. If `setup:administrate` policy is not validated later by the kernel or the controller handling the for it will allow you to remove any object state without having `setup:administrate` policy.

SylvainGuittard:
Thanks @barbaragr for testing
> Now I see the dropdown list with two options, not one. So should it be like scenario says or should I see the complete list?

It should be like it's described in the scenario. The object is already on `Locked` state. So there is no point of displaying `Locked` in the list. This list is only for changing the state of the object.s
> In addition, when there are two states in application and I have the policy ‘State/Assign' with limitation 'State/Lock:Locked' assigned to my User Group I will see dropdown lists with all items for both States. Shouldn't I see dropdown list only for State:Lock?

Does those two states belong to the same object state group?

barbaragr:
Thank you @SylvainGuittard 
Ping: @mikadamczyk we've got two issues here.

mikadamczyk:
First issue, with option witch is already set, is fixed. Second is not doable. For now we use hasAccess() method which does not check limitations. We can not use canUser() check, because issue with 'targets' cc @lserwatka

barbaragr:
Fix works fine :) I will approve it when it will be said what about the second issue.

PR #471, opened 2018-05-15
EZP-29102: Implement permissions for Languages in Admin Panel

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29102
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

I assumed that we will have (as in content, class and section) separate policies to edit, create and delete Language.

screen shot 2018-05-15 at 11 39 37 am
screen shot 2018-05-15 at 11 40 17 am
screen shot 2018-05-15 at 11 40 08 am
screen shot 2018-05-15 at 11 39 48 am

#### Checklist:
- [x] Ready for Code Review


src/bundle/Controller/LanguageController.php
```diff
@@ -107,8 +107,9 @@ public function listAction(Request $request): Response
         return $this->render('@EzPlatformAdminUi/admin/language/list.html.twig', [
             'pager' => $pagerfanta,
             'form_languages_delete' => $deleteLanguagesForm->createView(),
-            'canEdit' => $this->isGranted(new Attribute('content', 'translations')),
-            'canAssign' => $this->isGranted(new Attribute('content', 'translations')),
+            'can_edit' => $this->isGranted(new Attribute('content', 'translations')),
+            'can_delete' => $this->isGranted(new Attribute('content', 'translations')),
+            'can_create' => $this->isGranted(new Attribute('content', 'translations')),
```
dew326: Why do we need 3 variables containing the same thing?

src/bundle/Controller/LanguageController.php
```diff
@@ -107,8 +107,9 @@ public function listAction(Request $request): Response
         return $this->render('@EzPlatformAdminUi/admin/language/list.html.twig', [
             'pager' => $pagerfanta,
             'form_languages_delete' => $deleteLanguagesForm->createView(),
-            'canEdit' => $this->isGranted(new Attribute('content', 'translations')),
-            'canAssign' => $this->isGranted(new Attribute('content', 'translations')),
```
dew326: Isn't this a BC break?

webhdx: Yes. We should not freely rename Twig variables because it will break custom templates. I opt for adding both variables + camelCase can take a deprecation comment. What do you think?

src/bundle/Controller/LanguageController.php
webhdx: Indeed something is wrong here. I'm also not sure how `$this->isGranted()` is working, I'd rather use `PermissionResolver::hasAccess` for that. But `isGranted()` might be okay, just never looked what it does internally.

mikadamczyk:
@webhdx, @dew326 I changed to `can_administrate`. 'content/translations policy ' has no limitation, so IMHO it is fine to use `IsGranted()` cc @alongosz

src/bundle/Controller/LanguageController.php
lserwatka: @mikadamczyk but please add this deprecation comment, so we can remove this variable in 2.3 release 😸

src/bundle/Controller/LanguageController.php
```diff
@@ -107,8 +107,11 @@ public function listAction(Request $request): Response
         return $this->render('@EzPlatformAdminUi/admin/language/list.html.twig', [
             'pager' => $pagerfanta,
             'form_languages_delete' => $deleteLanguagesForm->createView(),
+            /* @deprecated since version 2.2, to be removed in 3.0. Use 'can_administrate' instead. */
             'canEdit' => $this->isGranted(new Attribute('content', 'translations')),
+            /* @deprecated since version 2.2, to be removed in 3.0. Use 'can_administrate' instead. */
             'canAssign' => $this->isGranted(new Attribute('content', 'translations')),
+            'can_administrate' => $this->isGranted(new Attribute('content', 'translations')),
```
ViniTou: Are we sure about changing capitalization here?

PR #477, opened 2018-05-21
EZP-29216: Implement permissions for Role management in Admin Panel

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29216
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?  |no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/admin/policy/list.html.twig
+++ b/src/bundle/Resources/views/admin/policy/list.html.twig
@@ -6,27 +6,34 @@
     <div class="ez-table-header">
         <div class="ez-table-header__headline">{{ 'policy.view.list.title.count'|trans({'%count%': role.policies|length})|desc('Policies (%count%)') }}</div>
         <div>
-            <a title="{{ 'policy.view.list.action.add'|trans|desc('Add a new Policy') }}"
-               href="{{ path('ezplatform.policy.create', {roleId: role.id}) }}"
-               class="btn btn-primary">
-                <svg class="ez-icon ez-icon-create">
-                    <use xmlns:xlink="http://www.w3.org/1999/xlink"
-                         xlink:href="{{ asset('bundles/ezplatformadminui/img/ez-icons.svg') }}#create"></use>
-                </svg>
-            </a>
-            {% set modal_data_target = 'delete-policies-modal' %}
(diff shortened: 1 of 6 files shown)


lserwatka:
@mikadamczyk could you make a JIRA ticket for this PR?

src/bundle/Resources/views/admin/policy/list.html.twig
```diff
@@ -6,27 +6,34 @@
...
+                    disabled
+                    data-toggle="modal"
+                    data-target="#{{ modal_data_target }}"
+                    title="{{ 'policy.view.list.action.delete'|trans|desc('Delete Policy') }}">
+                    <svg class="ez-icon ez-icon-trash">
+                        <use xlink:href="{{ asset('bundles/ezplatformadminui/img/ez-icons.svg') }}#trash"></use>
+                    </svg>
+                </button>
+                {% include '@EzPlatformAdminUi/admin/bulk_delete_confirmation_modal.html.twig' with {
+                'id': modal_data_target,
```
sunpietro: Indentation issue?

barbaragr:
`2.2.1.1 Create new roles` 
Test passed, because when I add Role/Create policy I can see [+] button, but I move step forward and I tried to add a new Role. I was redirected to new role form, filled it and click Create and then I was inform by error message that I don’t have access to Role/Update. Is it OK that Role/Create shows only [+] button, but doesn’t allow to create a new role?

`2.2.2.1 Update role name` 
I can’t update role name only with Role/Update policy. I’ve got message that I need Role/Create policy. The same with `2.2.2.2 Update role policies` - I need both: create and update to edit or remove policy. Scenarios passed, because they say that I should see edit buttons/ checkboxes, trash button - it is ok, but I can’t do it.

I’ve got error 500 with missing Section/View policy when I try to update role adding a new policy. Policies of my user:
- Content/*
- User/ Login
- User/Password
- Setup/Administrate
- Role/Read
- Role/Update
I go to a role, click on it, and try to add new policy - I chose Content/Remove.

SylvainGuittard:
For the record: we agreed to improve the policies when we will rework the interfaces.

PR #494, opened 2018-05-29
[Behat] Wait until notification is closed

mnocon:
| Question      | Answer
| ------------- | ---
| Tickets       | -
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/lib/Behat/PageElement/Notification.php
+++ b/src/lib/Behat/PageElement/Notification.php
@@ -14,6 +14,8 @@ class Notification extends Element
     /** @var string Name by which Element is recognised */
     public const ELEMENT_NAME = 'Notification';
 
+    private $checkVisibilityTimeout;
+
     public function __construct(UtilityContext $context)
     {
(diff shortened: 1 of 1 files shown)


src/lib/Behat/PageElement/Notification.php
```diff
@@ -48,6 +48,9 @@ public function getMessage(): string
     public function closeAlert(): void
     {
         $this->context->findElement($this->fields['closeAlert'])->click();
+        $this->context->waitUntil($this->defaultTimeout, function () {
+            return !$this->isVisible();
```
m-tyrala: `checkVisibilityByClass` has default timeout inside, equal to 5 seconds, so from the moment when Selenium noticed that element is not visible, we always have to wait 5s to get the `false` output.

PR #497, opened 2018-05-30
EZEE-2040: Expose in config map of content type identifiers to names

Nattfarinn:
| Question      | Answer
| ------------- | ---
| Tickets       | http://jira.ez.no/browse/EZEE-2040
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

We need to expose additional (flap) map of ContentType identifiers to names, without grouping etc.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/UI/Config/Provider/ContentTypeNames.php
+++ b/src/lib/UI/Config/Provider/ContentTypeNames.php
@@ -1,0 +1,40 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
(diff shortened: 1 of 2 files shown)


src/lib/UI/Config/Provider/ContentTypeNames.php
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * @param \eZ\Publish\API\Repository\ContentTypeService $contentTypeService
+     */
+    public function __construct(ContentTypeService $contentTypeService)
+    {
+        $this->contentTypeService = $contentTypeService;
+    }
+
+    /**
+     * @return mixed Anything that is serializable via json_encode()
```
mikadamczyk: I think we simply return array here

PR #499, opened 2018-06-04
EZP-28890: Added crowdin in-context translation JS code

bdunogier:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-28890](http://jira.ez.no/browse/EZP-28890)
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | yes

Adds support for in-context translation, as introduced in https://github.com/ezsystems/PlatformUIBundle/pull/773.

#### Checklist:
- [x] Consider having a variable that sets the crowdin project, instead of hardcoding `ezplatform`. It would allow usage of in-context for any 3rd party bundle.
- [ ] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/layout.html.twig
+++ b/src/bundle/Resources/views/layout.html.twig
@@ -34,6 +34,13 @@
         </script>
         <script src="/bundles/fosjsrouting/js/router.js"></script>
         <script src="{{ path('fos_js_routing_js', { callback: 'fos.Router.setData' }) }}"></script>
+        {% if app.request.locale == 'ach_UG' %}
+            <script type="text/javascript">
+                var _jipt = [];
+                _jipt.push(['project', 'ezplatform']);
+            </script>
+            <script type="text/javascript" src="//cdn.crowdin.com/jipt/jipt.js"></script>
+        {% endif %}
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/views/layout.html.twig
```diff
@@ -34,6 +34,13 @@
         </script>
         <script src="/bundles/fosjsrouting/js/router.js"></script>
         <script src="{{ path('fos_js_routing_js', { callback: 'fos.Router.setData' }) }}"></script>
+        {% if app.request.locale == 'ach_UG' %}
```
bdunogier: This could be abstracted so that the way to detect in-context isn't bound to the implementation (`if app.in_context_translation_enabled`).

src/bundle/Resources/views/layout.html.twig
```diff
@@ -34,6 +34,13 @@
...
+            <script type="text/javascript">
```
bdunogier: This could/should be moved to the bottom of the file. It could also be added by an external file, as it is supported by the template. Pointers appreciated.

bdunogier:
Ping @sunpietro @dew326. This needs to move forward guys :)

Without it, no in-context.

PR #502, opened 2018-06-11
EZEE-2059: Cannot embed content when editing a content in Page Builder

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZEE-2059
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

### DIFF
--- a/src/bundle/Resources/views/layout.html.twig
+++ b/src/bundle/Resources/views/layout.html.twig
@@ -4,6 +4,7 @@
         <meta charset="UTF-8" />
         <meta name="CSRF-Token" content="{{ csrf_token('authenticate') }}" />
         <meta name="SiteAccess" content="{{ app.request.get('siteaccess').name }}" />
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/views/layout.html.twig
```diff
@@ -4,6 +4,7 @@
         <meta charset="UTF-8" />
         <meta name="CSRF-Token" content="{{ csrf_token('authenticate') }}" />
         <meta name="SiteAccess" content="{{ app.request.get('siteaccess').name }}" />
+        <meta name="LanguageCode" content="{{ language.languageCode }}" />
```
dew326: This supposed to be a languageCode of editing content. I don't think this variable is available across whole admin system. It should be scope only for Page Builder edit.

PR #504, opened 2018-06-13
EZP-29295: Second level of main menu has no context passed

webhdx:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29295
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/parts/navigation.html.twig
+++ b/src/bundle/Resources/views/parts/navigation.html.twig
@@ -1,3 +1,9 @@
+{% set main_menu_context = {
+    'location': location ?? null,
+    'content': content ?? null
+} %}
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/views/parts/navigation.html.twig
```diff
@@ -1,3 +1,9 @@
+{% set main_menu_context = {
+    'location': location is defined ? location : null,
```
mikadamczyk: `'location': location ?? null` ?

mnocon:
Looks good, Behat failure is unrelated - it's something @m-tyrala is investigating right now.

PR #509, opened 2018-06-15
EZP-29266: Content item seems to be always bookmarked

tischsoic:
Fixes the bug concerning a bookmark star on a location page. The content was always bookmarked, even if it wasn't.

| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29266
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | -
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
+++ b/src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
@@ -3,7 +3,6 @@
     const SELECTOR_BOOKMARK_CHECKBOX = '#location_update_bookmark_bookmarked';
     const SELECTOR_BOOKMARK_LOCATION_INPUT = '#location_update_bookmark_location';
     const SELECTOR_BOOKMARK_WRAPPER = '.ez-add-to-bookmarks';
-
     const CLASS_BOOKMARK_CHECKED = 'ez-add-to-bookmarks--checked';
 
     const updateBookmarkLocationInput = doc.querySelector(SELECTOR_BOOKMARK_LOCATION_INPUT);
--- a/src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
+++ b/src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
@@ -2,6 +2,7 @@
     const SELECTOR_FORM = 'form[name="location_update_bookmark"]';
     const SELECTOR_BOOKMARK_CHECKBOX = '#location_update_bookmark_bookmarked';
     const SELECTOR_BOOKMARK_LOCATION_INPUT = '#location_update_bookmark_location';
+    const SELECTOR_BOOKMARK_WRAPPER = '.ez-add-to-bookmarks';
 
     const updateBookmarkLocationInput = doc.querySelector(SELECTOR_BOOKMARK_LOCATION_INPUT);
     const currentLocationId = parseInt(updateBookmarkLocationInput.value, 10);
@@ -15,14 +16,31 @@
     const isCurrentLocation = (locationId) => {
         return parseInt(locationId, 10) === currentLocationId;
     };
+    const setBookmarkWrapperCheckedClass = (bookmarked) => {
+        const wrapperClassList = document.querySelector(SELECTOR_BOOKMARK_WRAPPER).classList;
+        const bookmarkCheckedClass = 'ez-add-to-bookmarks--checked';
+
+        if (bookmarked) {
+            wrapperClassList.add(bookmarkCheckedClass);
+        } else {
+            wrapperClassList.remove(bookmarkCheckedClass);
+        }
+    };
     const updateBookmarkForm = (event) => {
         const { bookmarked, locationId } = event.detail;
 
         if (isCurrentLocation(locationId)) {
             updateBookmarkCheckbox(bookmarked);
+            setBookmarkWrapperCheckedClass(bookmarked);
         }
+    };
+    const updateBookmarkWrapperClass = (event) => {
+        const checked = event.target.checked;
+
+        setBookmarkWrapperCheckedClass(checked);
     };
 
     doc.body.addEventListener('ez-bookmark-change', updateBookmarkForm, false);
     doc.querySelector(SELECTOR_BOOKMARK_CHECKBOX).addEventListener('change', submitBookmarkForm, false);
+    doc.querySelector(SELECTOR_BOOKMARK_CHECKBOX).addEventListener('change', updateBookmarkWrapperClass, false);
 })(window, document);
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
```diff
@@ -15,14 +16,31 @@
     const isCurrentLocation = (locationId) => {
         return parseInt(locationId, 10) === currentLocationId;
     };
+    const setBookmarkWrapperCheckedClass = (bookmarked) => {
+        const wrapperClassList = document.querySelector(SELECTOR_BOOKMARK_WRAPPER).classList;
+        const bookmarkCheckedClass = 'ez-add-to-bookmarks--checked';
+
+        if (bookmarked) {
+            wrapperClassList.add(bookmarkCheckedClass);
```
ViniTou: Not an expert, but I believe there is toggle method
@sunpietro

tischsoic: This function also manages CSS class on event: `ez-bookmark-change`. 
I assume that can occur event `ez-bookmark-change` with `bookmarked` property set to `true` even if the content is already bookmarked.

src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
```diff
@@ -15,14 +16,31 @@
     const isCurrentLocation = (locationId) => {
         return parseInt(locationId, 10) === currentLocationId;
     };
+    const setBookmarkWrapperCheckedClass = (bookmarked) => {
+        const wrapperClassList = document.querySelector(SELECTOR_BOOKMARK_WRAPPER).classList;
```
sunpietro: Should be `const wrapper = document.querySelector(SELECTOR_BOOKMARK_WRAPPER);`

src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
```diff
@@ -15,14 +16,31 @@
     const isCurrentLocation = (locationId) => {
         return parseInt(locationId, 10) === currentLocationId;
     };
+    const setBookmarkWrapperCheckedClass = (bookmarked) => {
```
sunpietro: `isBookmarked`

src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
sunpietro: You can do it like this `wrapper.classList.toggle('ez-add-to-bookmarks--checked', isBookmarked);`. See the reference: https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
```diff
@@ -15,14 +16,31 @@
...
+        } else {
+            wrapperClassList.remove(bookmarkCheckedClass);
+        }
+    };
     const updateBookmarkForm = (event) => {
         const { bookmarked, locationId } = event.detail;
 
         if (isCurrentLocation(locationId)) {
             updateBookmarkCheckbox(bookmarked);
+            setBookmarkWrapperCheckedClass(bookmarked);
```
sunpietro: I don't like the function name. I see it more like this: `toggleBookmarkCheckedState`

src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
```diff
@@ -15,14 +16,31 @@
...
         }
     };
+    const updateBookmarkWrapperClass = (event) => {
```
sunpietro: `updateBookmarkCheckedState`

src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
```diff
@@ -15,14 +16,31 @@
     const isCurrentLocation = (locationId) => {
         return parseInt(locationId, 10) === currentLocationId;
     };
+    const setBookmarkWrapperCheckedClass = (bookmarked) => {
+        const wrapperClassList = document.querySelector(SELECTOR_BOOKMARK_WRAPPER).classList;
+        const bookmarkCheckedClass = 'ez-add-to-bookmarks--checked';
```
dew326: This should be at the top `CLASS_BOOKMARK_CHECKED`.

src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
dew326: `doc`

src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
tischsoic: I think that `updateBookmarkCheckedState ` suggests that the input state is being updated, when it is not.

tischsoic: Maybe, `updateBookmarkIconState`?

src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
tischsoic: Maybe, `toggleBookmarkIconState`?

dew326:
I don't where the markup was fixed to use bootstrap one?

src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
```diff
@@ -4,6 +4,8 @@
     const SELECTOR_BOOKMARK_LOCATION_INPUT = '#location_update_bookmark_location';
     const SELECTOR_BOOKMARK_WRAPPER = '.ez-add-to-bookmarks';
 
+    const CLASS_BOOKMARK_CHECKED = 'ez-add-to-bookmarks--checked';
```
sunpietro: Please, no extra lines before this `const`

lserwatka:
@dew326 @sunpietro are you OK with last changes?

src/bundle/Resources/public/js/scripts/admin.location.bookmark.js
sunpietro: You're using this CSS class once. There's no need to store it in a variable.

dew326: It was when you were using it twice. There is no need to cache string when you use it once but it's also not a mistake, so it can stay. :)

PR #532, opened 2018-06-22
EZP-29322: As an editor I want to use translated Admin panel

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29322
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

Refers to https://github.com/ezsystems/ezpublish-kernel/pull/2361

screen shot 2018-06-22 at 2 58 13 pm

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/config/services.yml
+++ b/src/bundle/Resources/config/services.yml
@@ -60,6 +60,13 @@ services:
         tags:
             - { name: kernel.event_subscriber }
 
+    EzSystems\EzPlatformAdminUi\EventListener\RequestLocaleListener:
+        arguments:
+            $siteAccessGroups: '%ezpublish.siteaccess.groups%'
+            $availableTranslations: '%available_translations%'
+        tags:
+            - { name: kernel.event_subscriber }
+
     EzSystems\EzPlatformAdminUi\EventListener\RequestAttributesListener:
         arguments:
             - '%ezpublish.siteaccess.groups%'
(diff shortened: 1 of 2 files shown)


lserwatka:
@bdunogier ping

mikadamczyk:
@bdunogier Is any chance to rename translation file name to the same as in symfony, for example from bookmark.de_DE.xlf to bookmark.de.xlf. It is a problem when browser send accept language as ‘no’ for ‘bn_no’ or ‘el’ for ‘el_GR’. So far this is the only solution I have.

bdunogier:
It will require some work & research, @mikadamczyk, but I think that we need to do this, indeed.

bdunogier:
The request listener looks good.

Maybe I would remove the `preg_match` part. Instead, we could modify the kernel's translation loader so that it registers translations using the language code instead of the language + region. This part belongs in the kernel rather than in the admin, as the frontend uses translation as well.

What do you think ?

bdunogier:
I agree about changing files names. But I also think it is too late so close from release. I'm only suggesting that instead of doing that in `AvailableTranslationPass`, we do it in the `TranslationCollectorPass`, that's all.

Also, what would we change them to ? The issue you mention about `fr_ca` is very valid. But it applies to renaming files & treating locales the way you did here. As far as crowdin.com is concerned, French is `fr`, and the other variants are regional (`fr_ca`, `fr_be`...).

For now, unless we decide to remove the translation files AND packages before friday, we should do the renaming when collecting the files, using a pass of some sort.

mikadamczyk:
First we must merge https://github.com/ezsystems/ezpublish-kernel/pull/2361

PR #538, opened 2018-06-26
EZP-29179: Implement permissions for "Content/Hide"

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29179
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

In addition we added ajax request to change location visibility

screen shot 2018-06-26 at 1 18 11 pm

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/LocationController.php
+++ b/src/bundle/Controller/LocationController.php
@@ -13,7 +13,6 @@ use eZ\Publish\API\Repository\ContentTypeService;
 use eZ\Publish\API\Repository\LocationService;
 use eZ\Publish\API\Repository\PermissionResolver;
 use eZ\Publish\API\Repository\TrashService;
-use eZ\Publish\API\Repository\Values\Content\Location;
 use eZ\Publish\API\Repository\Values\Content\LocationUpdateStruct;
 use eZ\Publish\Core\Base\Exceptions\InvalidArgumentException;
 use EzSystems\EzPlatformAdminUi\Form\Data\Content\Location\ContentLocationAddData;
@@ -476,14 +475,30 @@ class LocationController extends Controller
         if ($form->isSubmitted() && $form->isValid()) {
             $data = $form->getData();
 
+            /** @var \eZ\Publish\API\Repository\Values\Content\Location $location */
             $location = $data->getLocation();
             $hidden = $data->getHidden();
 
             try {
+                /** @var \eZ\Publish\API\Repository\Values\Content\ContentInfo $contentInfo */
+                $contentInfo = $location->getContentInfo();
+
                 if ($hidden) {
                     $this->locationService->hideLocation($location);
+                    $message = $this->translator->trans(
+                        /** @Desc("Location '%name%' hidden.") */
+                        'location.update_success.success.hidden',
+                        ['%name%' => $contentInfo->name],
+                        'location'
+                    );
                 } else {
                     $this->locationService->unhideLocation($location);
+                    $message = $this->translator->trans(
+                        /** @Desc("Location '%name%' unhidden.") */
+                        'location.update_success.success.unhidden',
+                        ['%name%' => $contentInfo->name],
+                        'location'
+                    );
                 }
             } catch (APIRepositoryUnauthorizedException $e) {
                 return new JsonResponse(['errors' => [$e->getMessage()]], Response::HTTP_UNAUTHORIZED);
@@ -497,7 +512,7 @@ class LocationController extends Controller
             return new JsonResponse(['errors' => $errors], Response::HTTP_UNPROCESSABLE_ENTITY);
         }
 
-        return new JsonResponse();
+        return new JsonResponse(['message' => $message]);
     }
 
     /**
--- a/src/bundle/Controller/LocationController.php
+++ b/src/bundle/Controller/LocationController.php
@@ -11,8 +11,9 @@ namespace EzSystems\EzPlatformAdminUiBundle\Controller;
 use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\ContentTypeService;
 use eZ\Publish\API\Repository\LocationService;
+use eZ\Publish\API\Repository\PermissionResolver;
 use eZ\Publish\API\Repository\TrashService;
-use eZ\Publish\API\Repository\Values\Content\ContentInfo;
+use eZ\Publish\API\Repository\Values\Content\Location;
 use eZ\Publish\API\Repository\Values\Content\LocationUpdateStruct;
 use eZ\Publish\Core\Base\Exceptions\InvalidArgumentException;
 use EzSystems\EzPlatformAdminUi\Form\Data\Content\Location\ContentLocationAddData;
@@ -21,7 +22,6 @@ use EzSystems\EzPlatformAdminUi\Form\Data\Location\LocationCopyData;
 use EzSystems\EzPlatformAdminUi\Form\Data\Location\LocationCopySubtreeData;
 use EzSystems\EzPlatformAdminUi\Form\Data\Location\LocationMoveData;
 use EzSystems\EzPlatformAdminUi\Form\Data\Location\LocationSwapData;
-use EzSystems\EzPlatformAdminUi\Form\Data\Location\LocationUpdateVisibilityData;
 use EzSystems\EzPlatformAdminUi\Form\Data\Location\LocationTrashData;
 use EzSystems\EzPlatformAdminUi\Form\Data\Location\LocationUpdateData;
 use EzSystems\EzPlatformAdminUi\Form\Factory\FormFactory;
@@ -29,13 +29,12 @@ use EzSystems\EzPlatformAdminUi\Form\SubmitHandler;
 use EzSystems\EzPlatformAdminUi\Notification\NotificationHandlerInterface;
 use EzSystems\EzPlatformAdminUi\Tab\LocationView\DetailsTab;
 use EzSystems\EzPlatformAdminUi\Tab\LocationView\LocationsTab;
+use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\OptionsResolver\Exception\InvalidOptionsException;
 use Symfony\Component\Translation\TranslatorInterface;
 use eZ\Publish\API\Repository\Exceptions\UnauthorizedException as APIRepositoryUnauthorizedException;
-use Symfony\Component\Translation\Exception\InvalidArgumentException as TranslationInvalidArgumentException;
 
 class LocationController extends Controller
 {
@@ -63,6 +62,9 @@ class LocationController extends Controller
     /** @var \EzSystems\EzPlatformAdminUi\Form\SubmitHandler */
     private $submitHandler;
 
+    /** @var \eZ\Publish\API\Repository\PermissionResolver */
+    private $permissionResolver;
+
     /**
      * @param \EzSystems\EzPlatformAdminUi\Notification\NotificationHandlerInterface $notificationHandler
      * @param \Symfony\Component\Translation\TranslatorInterface $translator
@@ -72,6 +74,7 @@ class LocationController extends Controller
      * @param \eZ\Publish\API\Repository\TrashService $trashService
      * @param \EzSystems\EzPlatformAdminUi\Form\Factory\FormFactory $formFactory
      * @param \EzSystems\EzPlatformAdminUi\Form\SubmitHandler $submitHandler
+     * @param \eZ\Publish\API\Repository\PermissionResolver $permissionResolver
      */
     public function __construct(
         NotificationHandlerInterface $notificationHandler,
@@ -81,7 +84,8 @@ class LocationController extends Controller
         ContentService $contentService,
         TrashService $trashService,
         FormFactory $formFactory,
-        SubmitHandler $submitHandler
+        SubmitHandler $submitHandler,
+        PermissionResolver $permissionResolver
     ) {
         $this->notificationHandler = $notificationHandler;
         $this->translator = $translator;
@@ -91,14 +95,13 @@ class LocationController extends Controller
         $this->trashService = $trashService;
         $this->formFactory = $formFactory;
         $this->submitHandler = $submitHandler;
+        $this->permissionResolver = $permissionResolver;
     }
 
     /**
-     * @param Request $request
+     * @param \Symfony\Component\HttpFoundation\Request $request
      *
-     * @return Response
-     *
-     * @throws InvalidArgumentException
+     * @return \Symfony\Component\HttpFoundation\Response
      */
     public function moveAction(Request $request): Response
     {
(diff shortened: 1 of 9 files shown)


lserwatka:
@sunpietro @dew326 we need frontend code review here.

src/bundle/Resources/public/js/scripts/services/notification.service.js
```diff
@@ -0,0 +1,93 @@
...
+     * Dispatches danger notification event
+     *
+     * @function showErrorNotification
+     * @param {Error} error
+     */
+    const showErrorNotification = (error) => {
+        showDangerNotification(error.message);
+    };
+
+    eZ.services = eZ.services || {};
```
dew326: I don't know if `services` is an accurate name. Maybe `helpers`?

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
```diff
@@ -1,12 +1,36 @@
...
+        const isHiddenNumberValue = !isVisible ? 1 : 0;
+        const changeVisibilityLink = window.Routing.generate('ezplatform.location.hide', { locationId, hidden: isHiddenNumberValue });
 
-    [...doc.querySelectorAll('.ez-checkbox-icon__checkbox')].forEach(checkbox => checkbox.addEventListener('change', updateVisibility, false));
+        fetch(changeVisibilityLink, fetchInit)
+            .then(handleUpdateResponse(event))
+            .catch(handleUpdateError);
+    };
+
+    for (const checkbox of visibilityCheckboxes) {
```
dew326: Why not old fashion `forEach`? :)

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
```diff
@@ -1,12 +1,36 @@
...
+        throw new Error(response.statusText);
+    };
+    const updateVisibility = (event) => {
+        const { value: locationId, checked: isVisible } = event.target;
+        const isHiddenNumberValue = !isVisible ? 1 : 0;
+        const changeVisibilityLink = window.Routing.generate('ezplatform.location.hide', { locationId, hidden: isHiddenNumberValue });
 
-    [...doc.querySelectorAll('.ez-checkbox-icon__checkbox')].forEach(checkbox => checkbox.addEventListener('change', updateVisibility, false));
+        fetch(changeVisibilityLink, fetchInit)
+            .then(handleUpdateResponse(event))
```
dew326: Why not just `handleUpdateResponse.bind(null, event)`?

src/bundle/Resources/config/routing.yml
```diff
@@ -479,6 +479,16 @@ ezplatform.location.update_visibility:
     defaults:
         _controller: 'EzPlatformAdminUiBundle:Location:updateVisibility'
 
+ezplatform.location.hide:
```
dew326: I think `hide` is not an accurate name.

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
```diff
@@ -1,12 +1,36 @@
...
+    const fetchInit = {
+        credentials: 'same-origin',
+    };
+    const visibilityCheckboxes = [...doc.querySelectorAll(SELECTOR_VISIBILITY_CHECKBOXES)];
 
-        doc.querySelector('#location_update_visibility_data_location').value = event.target.value;
-        doc.querySelector('#location_update_visibility_data_hidden').checked = !event.target.checked;
+    const onVisibilityUpdated = (event) => {
+        const { target } = event;
+        const { checked: isHidden } = target;
```
sunpietro: Do you want set a default value if none exists?

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
```diff
@@ -1,12 +1,36 @@
...
+(function(global, doc) {
+    const SELECTOR_VISIBILITY_CHECKBOXES = '#ez-tab-location-view-locations .ez-checkbox-icon__checkbox';
+    const fetchInit = {
+        credentials: 'same-origin',
+    };
+    const visibilityCheckboxes = [...doc.querySelectorAll(SELECTOR_VISIBILITY_CHECKBOXES)];
 
-        doc.querySelector('#location_update_visibility_data_location').value = event.target.value;
-        doc.querySelector('#location_update_visibility_data_hidden').checked = !event.target.checked;
+    const onVisibilityUpdated = (event) => {
```
sunpietro: `({ target }) => {`

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
```diff
@@ -1,12 +1,36 @@
...
+        const { target } = event;
+        const { checked: isHidden } = target;
 
-        doc.querySelector('form[name="location_update_visibility_data"]').submit();
-    }
+        target.closest('.ez-checkbox-icon').classList.toggle('is-checked', isHidden);
+    };
+    const handleUpdateError = global.eZ.services.notification.showErrorNotification;
+    const handleUpdateResponse = (event) => (response) => {
```
sunpietro: Is it intentional? You created a higher order function. Function that returns function.

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
```diff
@@ -1,12 +1,36 @@
...
+        if (response.status === 200) {
```
dew326: I would do it in opposite way in the `if` checked if there is an error and did `return`.

src/bundle/Controller/LocationController.php
```diff
@@ -485,6 +483,9 @@ public function updateVisibilityAction(Request $request): Response
         if ($form->isSubmitted() && $form->isValid()) {
             $result = $this->submitHandler->handle($form, function (LocationUpdateVisibilityData $data) use ($contentInfo) {
                 $location = $data->getLocation();
+
+                $this->canUserHideLocation($location);
```
Nattfarinn: Don't use "breakpoint"-like methods (no return value with exception in case of failure). There is need to wrap this simple `if` here.

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
tischsoic: Yes, it's intentional.

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
tischsoic: I think that this: `handleUpdateResponse(event)` is more clear than this: `handleUpdateResponse.bind(null, event)`.
In my opinion `bind` don't express well intention of applying one argument to the function. Higher order functions are better IMO.

src/bundle/Resources/public/js/scripts/services/notification.service.js
tischsoic: I was thinking about this and I actually searched for an answer on the Internet and I encountered this thread:
https://softwareengineering.stackexchange.com/questions/132067/difference-between-a-service-class-and-a-helper-class

The top answer says: 
> A Service class/interface provides a way of a client to interact with some functionality in the application.
and for me this code, which is sending notifications does that, "provides a way of a client to interact with some functionality in the application".
If you have better definition etc. I would gladly see it because I really want to name it correctly. :)

A helper for me would be e.g. a set of functions generating `CustomEvent`s but not dispatching it to the DOM.

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
sunpietro: Please, remove empty line above the `const onVisibilityUpdated`

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
sunpietro: Please, use binding instead.

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
```diff
@@ -1,12 +1,36 @@
...
+            onVisibilityUpdated(event);
+            return;
+        }
+
+        throw new Error(response.statusText);
+    };
+    const updateVisibility = (event) => {
```
sunpietro: `const updateVisibility = ({ value: locationId, checked: isVisible }) => {`

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
sunpietro: It's not a common approach in our codebase to use HoF. I'd stick with current approach and use `.bind()` function instead.

src/bundle/Resources/public/js/scripts/services/notification.service.js
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Dispatches notification event
+     *
+     * @function showNotification
+     * @param {Object} detail
+     */
+    const showNotification = (detail) => {
+        const event = new CustomEvent('ez-notify', {
+            detail,
```
sunpietro: It can be in the same line  with the `event` variable declaration

sunpietro: What's more, you don't even need this variable and make the whole `const showNotification` function declaration in one line.

src/bundle/Resources/public/js/scripts/services/notification.service.js
sunpietro: IMO, a service is meant to manage data.

lserwatka:
@tischsoic @sunpietro please reach the conclusion on the implementation details in private. We MUST merge this PR today!

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
tischsoic: Why?

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
tischsoic: Should I? I think that there is no default value in this case.

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
tischsoic: The `event` is passed to `handleUpdateResponse` so it is rather impossible to write it like that.

src/bundle/Resources/public/js/scripts/services/notification.service.js
tischsoic: @sunpietro , can you provide some reference?

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
tischsoic: OK, I saw comment below.

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
tischsoic: OK, this is some argument. 
I accept it, consistency is important, but do you have any other arguments against using HoF in such cases? I am just asking from pure curiosity about good practices. :)

dew326: There is no point to create some workarounds if something is native in JS. 
This is about using `.bind()` vs using imitation of `bind`.

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
dew326: As discussed in private. There is no point to avoid side effects in this case and with `forEach` the code would be simpler.

src/bundle/Controller/LocationController.php
```diff
@@ -465,72 +464,48 @@ public function addAction(Request $request): Response
...
             }
+
+            return new JsonResponse(['errors' => $errors], Response::HTTP_UNPROCESSABLE_ENTITY);
         }
 
-        return $this->redirect($this->generateUrl('_ezpublishLocation', [
-            'locationId' => $contentInfo->mainLocationId,
-            '_fragment' => LocationsTab::URI_FRAGMENT,
-        ]));
+        return new JsonResponse();
```
dew326: I think this should return some `success` and ideally some text for the `success` notification.

src/bundle/Resources/views/content/locationview.html.twig
```diff
@@ -110,6 +110,7 @@
 
 {% block javascripts %}
     {%  javascripts
+        '@EzPlatformAdminUiBundle/Resources/public/js/scripts/services/notification.service.js'
```
dew326: I think this should be in `layout.html.twig` to make this global not only for `content structure`.

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
```diff
@@ -1,12 +1,35 @@
...
+
+        onVisibilityUpdated(event);
+    };
+    const updateVisibility = (event) => {
         doc.querySelector('#location_update_visibility_data_location').value = event.target.value;
         doc.querySelector('#location_update_visibility_data_hidden').checked = !event.target.checked;
 
-        doc.querySelector('form[name="location_update_visibility_data"]').submit();
-    }
+        $.ajax({
```
dew326: ?!

lserwatka:
@dew326 @sunpietro can we merge it now?

dew326:
Not with `jQuery` used for ajax request.

tischsoic:
I removed jQuery in last commit. @dew326

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
```diff
@@ -1,12 +1,42 @@
...
+    const handleUpdateSuccess = ({ message }) => {
+        global.eZ.helpers.notification.showSuccessNotification(message);
+    };
+    const handleUpdateResponse = (event, response) => {
+        if (response.status !== 200) {
+            throw new Error(response.statusText);
+        }
+
+        onVisibilityUpdated(event);
+        response.json().then(handleUpdateSuccess);
```
sunpietro: You should return `response.json()` and then in the next `then()` you should proccess this data.

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
```diff
@@ -1,12 +1,42 @@
...
+    const handleUpdateError = global.eZ.helpers.notification.showErrorNotification;
+    const handleUpdateSuccess = ({ message }) => {
+        global.eZ.helpers.notification.showSuccessNotification(message);
+    };
+    const handleUpdateResponse = (event, response) => {
+        if (response.status !== 200) {
+            throw new Error(response.statusText);
+        }
+
+        onVisibilityUpdated(event);
```
sunpietro: If you handle response here, then you should call this method in the next `then()`

src/bundle/Resources/public/js/scripts/admin.location.visibility.js
```diff
@@ -1,12 +1,42 @@
...
+        response.json().then(handleUpdateSuccess);
+    };
+    const updateVisibility = (event) => {
         doc.querySelector('#location_update_visibility_data_location').value = event.target.value;
         doc.querySelector('#location_update_visibility_data_hidden').checked = !event.target.checked;
```
dew326: This can be changed to not search in whole `doc` but only on `form`.

PR #539, opened 2018-06-26
EZP-29330: Long title covers (X) button in message bar

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29330
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | -
| Doc needed?   | no

Fixed long text in notification covering close button.

#### Checklist:
- [x] Ready for Code Review


dew326:
If we want less code we can add just more padding on the right side.

dew326:
If we want less code we can add just more padding on the right side.

PR #548, opened 2018-06-28
EZP-29114: Fix pagination parameter naming for draft, system and custom url

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29114
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Long time ago, there was this idea that there would be only one pagination parameter. Reality has shown something different ;)
As we two times already fall into trap of this defaultPaginationParam name be missguiding, we decided to rename it to be more specific.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/ContentViewController.php
+++ b/src/bundle/Controller/ContentViewController.php
@@ -52,7 +52,7 @@ class ContentViewController extends Controller
     private $bookmarkService;
 
     /** @var int */
-    private $defaultPaginationLimit;
+    private $defaultDraftPaginationLimit;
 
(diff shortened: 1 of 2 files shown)


micszo:
@ViniTou could you provide some more info either here or in ticket? thx

ViniTou:
@micszo 
done, there is nothing more here than renaming this unfortunately named parameter to avoid more confusion ;)

PR #552, opened 2018-07-01
[WIP] EZP-29363: switched delete-check to use versionremove

skonrad:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29363
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/Menu/ContentEditRightSidebarBuilder.php
+++ b/src/lib/Menu/ContentEditRightSidebarBuilder.php
@@ -99,7 +99,7 @@ class ContentEditRightSidebarBuilder extends AbstractBuilder implements Translat
 
         $canPublish = $this->permissionResolver->canUser('content', 'publish', $content);
         $canEdit = $this->permissionResolver->canUser('content', 'edit', $content);
-        $canDelete = $this->permissionResolver->canUser('content', 'remove', $content);
+        $canRemoveVersion = $this->permissionResolver->canUser('content', 'versionremove', $content->versionInfo);
 
         $publishAttributes = [
             'class' => self::BTN_TRIGGER_CLASS,
@@ -145,7 +145,7 @@ class ContentEditRightSidebarBuilder extends AbstractBuilder implements Translat
         $items[self::ITEM__CANCEL] = $this->createMenuItem(
             self::ITEM__CANCEL,
             [
-                'attributes' => $canDelete
+                'attributes' => $canRemoveVersion
                     ? $deleteAttributes
                     : array_merge($deleteAttributes, self::BTN_DISABLED_ATTR),
                 'extras' => ['icon' => 'circle-close'],


skonrad:
poke @andrerom @lserwatka

lserwatka:
@skonrad could you share some reasoning behind this change?

andrerom:
So this is for `ITEM__CANCEL` action during edit, so it's about deleting the version and not the whole content item right? _(unless first, unpublished draft)_

If so, makes sense, but then let's also rename the variable to make this more consistent. Something like ´$canDeleteDraft´ or `canDeleteVersion`.

skonrad:
The `ITEM__CANCEL` maybe should also be renamed to `ITEM__DISCARD` as cancel infers the actual cancel button in the view (which keeps the draft in system).

I will rename the `canDelete` to `canRemoveVersion` as this more likely matches the permission.

Regarding @lserwatka question: The condition and origin of this change is described in https://jira.ez.no/browse/EZP-29363 -> discarding the draft is disabled if the user is not allowed to delete the whole content from the system.

skonrad:
I reaffirmed that this check will also allow the user to discard the first draft of new content (basically a delete for unpublished content).

skonrad:
After further investigation I should change the variable definition of the check to the following code:
```
if ($content->versionInfo->versionNo > 1) {
    $canRemoveVersion = $this->permissionResolver->canUser('content', 'versionremove', $content->versionInfo);
} else {
    $canRemoveVersion = $this->permissionResolver->canUser('content', 'remove', $content);
}
```

The problem with this change is, that the editor can now create a draft of new content but not discard it. This would lead to open drafts which will probably never be published.

@andrerom maybe the permission system should catch that eventuality allowing the user to delete unpublished content if `versionremove` matches it? Another possibility would be to add `Status` policy to `content/delete` which checks the currently "active" (and usually published) version.

skonrad:
Created https://github.com/ezsystems/ezpublish-kernel/pull/2390 to adjust the permission check for removing content (checking for `versionremove` if unpublished version).

This PR should be on hold until the kernel change is merged (imho).

bdunogier:
> This PR should be on hold until the kernel change is merged (imho).

Will the current situation of admin be worse, equivalent or better if this is merged before the kernel change ? If it doesn't create new problems, we shouldn't block the merge.

As far as I get it, it is currently _impossible_ for an editor with `content/versionremove` but without `content/delete` to delete his draft. With this change, it becomes possible in many situations, does it not ?

skonrad:
@bdunogier it would be a shift of the actual issue (atm) and obscure the it. This could lead to users opening up new issues which are basically clones of this one.

sunpietro:
@skonrad what's the status here?

andrerom:
Closing

this was fixed in https://github.com/ezsystems/ezplatform-admin-ui/commit/93ec4cf63f9e7ce6c254ae0c24516d5b2913f3a5

PR #556, opened 2018-07-05
EZP-28890: Impl. extractor for *.js files

adamwojs:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28890
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Configuration for `willdurand/js-translation-bundle` will be separate PR. 

#### Checklist:
- [X] Ready for Code Review

### DIFF
--- a/src/lib/Translation/Extractor/JavaScriptFileVisitor.php
+++ b/src/lib/Translation/Extractor/JavaScriptFileVisitor.php
@@ -1,0 +1,231 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformAdminUi\Translation\Extractor;
+
+use Doctrine\Common\Annotations\DocParser;
+use JMS\TranslationBundle\Annotation\Desc;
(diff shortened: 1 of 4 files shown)


src/lib/Translation/Extractor/JavaScriptFileVisitor.php
```diff
@@ -0,0 +1,218 @@
...
+use Psr\Log\LoggerAwareTrait;
+use Psr\Log\NullLogger;
+use SplFileInfo;
+
+class JavaScriptFileVisitor implements FileVisitorInterface, LoggerAwareInterface
+{
+    use LoggerAwareTrait;
+
+    // Example match: /** @Desc("Default text") */
+    const DESC_PATTERN = '/\*?\s*@Desc\(["\'](.*)["\']\)\s*/';
```
Nattfarinn: I prefer full match instead of partial one. Right now it will match anything that contains `@Desc("Something")`, even if it is part of something like: `/* @todo: It would be good to add something like @Desc("Translated text") here when we get proper translation */`. It should look for exact match of a comment only with `@Desc` element. (Proper pattern is on the way)

adamwojs:
PR updated according to @ViniTou / @Nattfarinn suggestions.

adamwojs:
Switched from regular expression to the doctrine annotations parser for `@Desc()` comments parsing 😉

PR #567, opened 2018-07-17
EZP-29423: Lazy loading value of the admin_ui_config Twig global variable (CS)

adamwojs:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29423
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [X] Ready for Code Review

### DIFF
--- a/src/lib/UI/Config/ConfigWrapper.php
+++ b/src/lib/UI/Config/ConfigWrapper.php
@@ -16,8 +16,6 @@ class ConfigWrapper implements \ArrayAccess, \JsonSerializable
     private $config;
 
     /**
-     * Config constructor.
-     *
      * @param array $config
      */
     public function __construct(array $config)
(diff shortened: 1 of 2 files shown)


src/lib/UI/Config/ConfigWrapper.php
```diff
@@ -16,7 +16,7 @@ class ConfigWrapper implements \ArrayAccess, \JsonSerializable
     private $config;
 
     /**
-     * Config constructor.
+     * ConfigWrapper constructor.
```
webhdx: I know it's coming from PHPStorm but I suggest removing it as it doesn't provide any value.

adamwojs: Fixed.

PR #568, opened 2018-07-18
Fix EZP-29433. Unauthorized exception user has no permission to read the parent location

crevillo:
…the parent location

| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29433
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

When the situation explained in the ticket, the user has read permissions limited to '/1/2/' and '/1/43/' 

When viewing the location 2 in the admin, it tries to render the tabs with the info about related urls. The tab related code is trying to load the parent Location at some point and hence the error. 

This is not only a problem with the root node. if you have a role with content / read and limitations `/1/2/100/200` `/1/2/100/300`, you will have the same error if trying to view the location 200 or 300 in the admin interface. 

So, i suggest to hardcode the parent_name to something in case you can't read the parent location. the only use of this var later is to indicate you that you will be creating an url alias under `{{ parent_name }}`

Take this as a proof of concept. Maybe we can move that test to a constant and make it translatable too...

ping @andrerom @alongosz 

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/content/tab/url/modal_add_custom_url.html.twig
+++ b/src/bundle/Resources/views/content/tab/url/modal_add_custom_url.html.twig
@@ -28,7 +28,13 @@
                 <span class="ez-modal--custom-url-alias__label">{{ 'tab.urls.add.site_root'|trans|desc('Place alias at the site root') }}</span>
                 {{ form_widget(form.site_root) }}
                 <div class="ez-modal--custom-url-alias__info-text">{{ 'tab.urls.add.site_root.helper.checked'|trans|desc('Checked will create the alias at the site root.') }}</div>
-                <div class="ez-modal--custom-url-alias__info-text">{{ 'tab.urls.add.site_root.helper.unchecked'|trans({'%parent_name%': parent_name})|desc('Unchecked will create the new alias under %parent_name%') }}</div>
+                <div class="ez-modal--custom-url-alias__info-text">
+                    {% if parent_name is not null %}
+                        {{ 'tab.urls.add.site_root.helper.unchecked'|trans({'%parent_name%': parent_name})|desc('Unchecked will create the new alias under %parent_name%') }}
+                    {% else %}
+                        {{ 'tab.urls.add.site_root.helper.no_parent_name'|trans|desc('Unchecked will create the new alias under the parent of this location') }}
+                    {% endif %}
+                </div>
             </div>
             <div class="modal-footer justify-content-center">
                 <button type="button" class="btn btn-dark" data-dismiss="modal">
--- a/src/lib/Tab/LocationView/UrlsTab.php
+++ b/src/lib/Tab/LocationView/UrlsTab.php
@@ -8,6 +8,7 @@ declare(strict_types=1);
 
 namespace EzSystems\EzPlatformAdminUi\Tab\LocationView;
 
+use eZ\Publish\API\Repository\Exceptions\UnauthorizedException;
 use eZ\Publish\API\Repository\LocationService;
 use eZ\Publish\API\Repository\URLAliasService;
 use eZ\Publish\API\Repository\Values\Content\Location;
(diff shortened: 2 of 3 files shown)


andrerom:
Nice one! So either then something like this, or load it using sudo() then.

@webhdx / @mikadamczyk who should review this?

crevillo:
Sudo can be an option too, yes.
If you prefer it i can add it.

lserwatka:
@crevillo could you rebase with 1.2 branch? We can shipp it with 2.2.2.

crevillo:
@lserwatka ok, should i edit the pr to have it based on that branch too?

crevillo:
Anyway, and while this can fix the issue, i'm wondering. 
The editors group have the member role and the editor role assigned
Member Role is the one adding the subtreelimitations `/1/2/` `/1/43/`
But Editor role adds no limitation at all... Shouldn't this other rule have "priority"? i mean, if in one role i give full read permissions, why i have limitations coming from the other?

ping @andrerom

lserwatka:
Yes, I meant target branch for PR to be set to 1.2 brach na rebased

crevillo:
done.

src/lib/Tab/LocationView/UrlsTab.php
```diff
@@ -130,21 +131,26 @@ public function renderView(array $parameters): string
...
             'system_urls_pager' => $systemUrlPagerfanta,
             'system_urls_pagination_params' => $systemUrlsPaginationParams,
             'can_edit_custom_url' => $canEditCustomUrl,
         ];
 
+        try {
+            $parentLocation = $this->locationService->loadLocation($location->parentLocationId);
+            $viewParameters['parent_name'] = $parentLocation->contentInfo->name;
+        } catch (UnauthorizedException $exception) {
+            $viewParameters['parent_name'] = 'the parent of this location';
```
mikadamczyk: 'the parent of this location' should be translatable. We can set `$viewParameters['parent_name'] = null` and in `src/bundle/Resources/views/content/tab/url/modal_add_custom_url.html.twig:31`, if parent_name is null, we can display proper message.

crevillo:
@mikadamczyk done that, can you explain which is the process to add that literal to a xliff file?

crevillo:
@mikadamczyk done, thanks. I squashed everything in one commit too.

ViniTou:
@crevillo 
I strongly believe that the intention was to have one full string when we don't have access to parent location. So it should be like
`tabs.urls.add.site_root.helper.default_parent_name = Unchecked will create the new alias under the parent of this location`

crevillo:
i'll change that asap.

crevillo:
@ViniTou @mikadamczyk is that if better?

src/bundle/Resources/views/content/tab/url/modal_add_custom_url.html.twig
```diff
@@ -28,7 +28,11 @@
                 <span class="ez-modal--custom-url-alias__label">{{ 'tab.urls.add.site_root'|trans|desc('Place alias at the site root') }}</span>
                 {{ form_widget(form.site_root) }}
                 <div class="ez-modal--custom-url-alias__info-text">{{ 'tab.urls.add.site_root.helper.checked'|trans|desc('Checked will create the alias at the site root.') }}</div>
-                <div class="ez-modal--custom-url-alias__info-text">{{ 'tab.urls.add.site_root.helper.unchecked'|trans({'%parent_name%': parent_name})|desc('Unchecked will create the new alias under %parent_name%') }}</div>
+                {% if parent_name is not null %}
+                    <div class="ez-modal--custom-url-alias__info-text">{{ 'tab.urls.add.site_root.helper.unchecked'|trans({'%parent_name%': parent_name})|desc('Unchecked will create the new alias under %parent_name%') }}</div>
```
mikadamczyk: Would you change it?
```php
<div class="ez-modal--custom-url-alias__info-text">
  {% if parent_name is not null %}
    {{ 'tab.urls.add.site_root.helper.unchecked'|trans({'%parent_name%': parent_name})|desc('Unchecked will create the new alias under %parent_name%') }}
  {% else %}
    {{ 'tab.urls.add.site_root.helper.no_parent_name'|trans|desc('Unchecked will create the new alias under the parent of this location') }}
  {% endif %}
</div>

``` 
It is no reason to duplicate `<div>`

crevillo:
changed to avoid the duplication of the div @mikadamczyk @ViniTou

mikadamczyk:
Thank you @crevillo :+1:

PR #570, opened 2018-07-20
EZP-29430: Missing pagination in Notification pop-up

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29430
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | ?
| Doc needed?   | no

This PR introduces pagination to notifications' modal. Everything else should work as it was working previously.

screen shot 2018-07-20 at 12 35 20

#### Checklist:
- [ ] Ready for Code Review


src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
```diff
@@ -1,89 +1,171 @@
...
     const INTERVAL = 30000;
-    const token = document.querySelector('meta[name="CSRF-Token"]').content;
+    const token = doc.querySelector('meta[name="CSRF-Token"]').content;
     const modal = doc.querySelector('.n-notifications-modal');
-    const markAsRead = (notification) => {
-        const requestOptions = {
+    const handleResponseError = eZ.helpers.notification.showErrorNotification;
+    const handleRequestResponseJson = eZ.helpers.request.handleRequestResponseJson;
+    const handleRequestResponseText = eZ.helpers.request.handleRequestResponseText;
+    const onNotificationMarkedAsRead = (notification, response) => {
```
sunpietro: `markAsRead` is easier to read

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
```diff
@@ -1,89 +1,171 @@
...
 
-        fetch(modal.querySelector(SELECTOR_TABLE).dataset.notificationsCount, requestOptions)
-            .then(response => response.json())
-            .then(data => setPendingNotificationCount(data));
+        fetch(request)
+            .then(handleRequestResponseJson)
+            .then(setPendingNotificationCount)
+            .catch(handleResponseError);
+    };
+    const updateUserAvatar = (notificationsInfo) => {
```
sunpietro: It doesn't update avatar.

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
```diff
@@ -1,89 +1,171 @@
...
+
+        if (notificationsTotal !== notificationsTotalOld) {
+            notificationsTable.dataset.notificationsTotal = notificationsTotal;
+
+            refreshNotificationsPage();
+        }
     };
-    const getNotificationsList = () => {
-        const requestOptions = {
+    const setNotificationPage = (pageHtml) => {
```
sunpietro: I think the function name is misleading. Should be like `showNotificationPage`

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
```diff
@@ -1,89 +1,171 @@
...
+        const modalResults = modal.querySelector(SELECTOR_MODAL_RESULTS);
+
+        modalResults.innerHTML = pageHtml;
+        toggleLoading(true);
+    };
+    const toggleSpinner = (force) => {
```
sunpietro: What if `force` is undefined?

src/bundle/Resources/public/js/scripts/helpers/request.helper.js
```diff
@@ -0,0 +1,58 @@
...
+    };
+
+    /**
+     * Handles request JSON response
+     *
+     * @function handleRequestResponseJson
+     * @param {Response} response
+     * @returns {Error|Promise}
+     */
+    const handleRequestResponseJson = (response) => {
```
sunpietro: `getJsonFromResponse`

src/bundle/Resources/public/js/scripts/helpers/request.helper.js
```diff
@@ -0,0 +1,58 @@
...
+    };
+
+    /**
+     * Handles request text response
+     *
+     * @function handleRequestResponseText
+     * @param {Response} response
+     * @returns {Error|Promise}
+     */
+    const handleRequestResponseText = (response) => {
```
sunpietro: `getTextFromResponse`

src/bundle/Resources/public/js/scripts/helpers/request.helper.js
```diff
@@ -0,0 +1,58 @@
...
+    };
+
+    /**
+     * Handles request response; returns status if response is OK
+     *
+     * @function handleRequestResponseStatus
+     * @param {Response} response
+     * @returns {Error|Promise}
+     */
+    const handleRequestResponseStatus = (response) => {
```
sunpietro: `getStatusFromResponse`

src/bundle/Resources/public/js/scripts/helpers/request.helper.js
```diff
@@ -0,0 +1,58 @@
...
+    const eZ = (global.eZ = global.eZ || {});
+
+    /**
+     * Handles request error
+     *
+     * @function handleRequestError
+     * @param {Response} response
+     * @returns {Error|Promise}
+     */
+    const handleRequestError = (response) => {
```
sunpietro: `handleRequest` because it handles request. If there's an error then it throws an error, otherwise it returns a response.

src/bundle/Resources/public/scss/_notifications-modal.scss
```diff
@@ -112,4 +112,23 @@
...
         }
     }
+
+    $spinner-size: 2rem;
+    &__spinner {
+        width: $spinner-size;
+        height: $spinner-size;
+        position: absolute;
+        bottom: calc(50% + #{$spinner-size/2});
+        left: calc(50% - #{$spinner-size/2});
```
sunpietro: Please use `transform` CSS property with `translate()` functions instead doing such calculations. It applies both to `bottom` and `left`

src/bundle/Resources/views/notifications/notifications_modal.html.twig
```diff
@@ -24,24 +24,16 @@
...
-                        <th>{{ 'notification.type'|trans|desc('Type') }}</th>
-                        <th>{{ 'notification.description'|trans|desc('Description') }}</th>
-                        <th>{{ 'notification.date'|trans|desc('Date') }}</th>
-                    </tr>
-                    </thead>
-                    <tbody class="n-table__body">
-                    {{ render(controller('EzPlatformAdminUiBundle:Notification:renderNotifications', {
-                        'limit': 5,
-                        'offset': 0,
+                <div class="n-notifications-modal__spinner n-notifications-modal__spinner--invisible">
```
sunpietro: This is not `n-something` ... It should be `c-something` where `c` means component. Please, update all the HTML code for notifications.

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
tischsoic: This undefined value would be passed to `toggle` function and it should be OK.
Do you want me to change it? Better solution does not come to my mind.

src/bundle/Resources/public/js/scripts/helpers/request.helper.js
tischsoic: What do you think about `handleResponseError`?

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
sunpietro: Isn't `undefined` treated as `false`?

src/bundle/Resources/public/js/scripts/helpers/request.helper.js
sunpietro: No because it does 2 things. Handling the error case is one of them. It just handles a response from request.

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
tischsoic: I checked it and undefined is treated as if there was no 2 argument in `toggle`. 
Nevertheless, I noticed that in `toggleLoading` there is:
```js
toggleResults(!force);
```
which results in `true` if `force` is undefined, I will modify `toggleLoading`.

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
```diff
@@ -1,89 +1,171 @@
...
+        toggleLoading(false);
+        fetch(request)
+            .then(handleRequestResponseText)
+            .then(setNotificationPage)
+            .catch(handleResponseError);
     };
-    const setPendingNotificationCount = (notificationsInfo) => {
-        const methodName = notificationsInfo.pending ? 'add' : 'remove';
-        const userAvatar = doc.querySelector('.ez-user-menu__avatar-wrapper');
+    const refreshNotificationsPage = () => {
```
dew326: I don't see any reason to have this wrapper function. This is called once and you could call `fetchNotificationPage(currentPageLink);` directly.

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
```diff
@@ -1,89 +1,171 @@
...
+            return;
+        }
 
-        if (notificationsCount !== notificationsInfo.total) {
-            notificationsCount = notificationsInfo.total;
+        const notificationsPageLink = target.href;
 
-            getNotificationsList();
-        }
+        event.preventDefault();
```
dew326: please use `preventDefault` always on the top of the method.

src/bundle/Resources/views/notifications/notifications_modal.html.twig
dew326: It was `n` as this was the `notifications-bundle`, now this should be `ez-notifications`.
Hard to tell if it's a BC break. I mean technically it is but I think we don't have a strong BC policy on templates.

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
tischsoic: But before I check `if (!isPaginationBtn)` I shoud not handle this. So should I put it over: `const notificationsPageLink = target.href;`?

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
```diff
@@ -1,89 +1,171 @@
...
+
+        if (!notification) {
+            return;
+        }
+
+        handleNotificationClick(notification);
     };
-    const getNotificationsStatus = () => {
-        const requestOptions = {
+    const updateNotificationsStatus = () => {
```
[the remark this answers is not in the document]
tischsoic: I always take into consideration what other functions this function is calling and it is calling `setPendingNotificationCount` which updates pending notifications status and updates the notifications' table if needed.
Do you think it is wrong approach?

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
```diff
@@ -1,89 +1,171 @@
...
+        const spinner = modal.querySelector(SELECTOR_MODAL_SPINNER);
+
+        spinner.classList.toggle(CLASS_MODAL_SPINNER_INVISIBLE, force);
+    };
+    const toggleResults = (force) => {
+        const results = modal.querySelector(SELECTOR_MODAL_RESULTS);
+
+        results.classList.toggle(CLASS_MODAL_RESULTS_INVISIBLE, force);
+    };
+    const toggleLoading = (force) => {
```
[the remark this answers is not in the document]
tischsoic: Good idea!

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
dew326: Now I see it better. 
The check `if (!isPaginationBtn)` should not be in this method. The `handleModalResultsClick` should work as a dispatcher, here should be checked what was clicked and called the proper function. Not calling everything and inside the method check if it's correct one.

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
tischsoic: @dew326 I have chosen `force` because in MDN in `toggle` definition there is `force`: https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

dew326: Yes but when you calling method `toggleSpinner` it is more natural to do `show/hide` not to `force`

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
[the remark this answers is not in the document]
dew326: For me the flow is more like:
`get notifications status` then `set notifications count`
not like:
`update notifications status` then `set notifications count`

any opinion @sunpietro ?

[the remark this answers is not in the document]
sunpietro: I agree with @dew326

tischsoic:
Ping @webhdx @dew326 
Code improvements has been made according to comments.

tischsoic:
Ping @Nattfarinn 
Can you check changes on behalf of @webhdx , who is absent today, please? :)

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
```diff
@@ -1,89 +1,164 @@
...
+    };
+    const fetchNotificationPage = (link) => {
+        if (!link) {
+            return;
+        }
+
+        const request = new Request(link, {
+            method: 'GET',
             headers: {
-                'X-CSRF-Token': token
```
dew326: We don't have to send `CSRF token` for notification page but we have to fo notification count?

src/bundle/Resources/public/scss/_notifications-modal.scss
```diff
@@ -112,4 +116,29 @@
...
+        width: 2rem;
+        height: 2rem;
+        position: absolute;
+        bottom: 50%;
+        left: 50%;
+        transform: translate(-50%, -50%);
+        display: none;
+
+        .ez-icon {
+            width: 2rem;
```
dew326: I think 2 rem is default value for `ez-icon`, do we need this rule?

lserwatka:
@dew326 @Nattfarinn are you OK with changes here?

PR #578, opened 2018-07-25
EZP-29091: Publishing Content from the frontend results in error

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29091
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Added new FormProcessor into adminUi in order to properly handle redirects after publishing/canceling content inside admin panel. There is change in default behavior in repository-forms as right now it by defaults redirect onto system-url instead of /content/location/{id}.

Relevant PR:
https://github.com/ezsystems/repository-forms/pull/244

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/RepositoryForms/Form/Processor/Content/UrlRedirectProcessor.php
+++ b/src/lib/RepositoryForms/Form/Processor/Content/UrlRedirectProcessor.php
@@ -8,12 +8,16 @@ declare(strict_types=1);
 
 namespace EzSystems\EzPlatformAdminUi\RepositoryForms\Form\Processor\Content;
 
+use eZ\Publish\API\Repository\Exceptions\InvalidArgumentException;
+use eZ\Publish\API\Repository\Exceptions\NotFoundException;
+use eZ\Publish\API\Repository\URLAliasService;
+use eZ\Publish\API\Repository\Values\Content\URLAlias;
+use eZ\Publish\Core\MVC\Symfony\SiteAccess;
 use EzSystems\EzPlatformAdminUi\Specification\SiteAccess\IsAdmin;
 use EzSystems\RepositoryForms\Event\FormActionEvent;
 use EzSystems\RepositoryForms\Event\RepositoryFormEvents;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
-use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Routing\RouterInterface;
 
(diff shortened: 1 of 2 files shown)


ViniTou:
@webhdx 
Not fan of it either, but we didn't fount better solution for this. Unless, as I suggested somewhere during our internal discussion we would remove RedirectResponse from event and then setup late listener to build redirect response from Location but that would evolve some major changes I believe.

webhdx:
As far as I understand the redirection is invalid because one of the AdminUI or PageBuilder form processor is changing it to the back office limited route. Easiest fix to this is simply adding a condition to run them only on admin siteaccesses.

webhdx:
Which route user is being redirected to on content create? Find the processor which is responsible for setting this redirection and create a condition there. I'm not sure how if it's done in repository-forms if so I'd change redirection there to "safe" one using url aliases and create a new processor which will override the redirection to AdminUI specific route when content is being added from admin siteaccess.

ViniTou:
@webhdx 
Are you by any chance talking about this linked PR?
https://github.com/ezsystems/repository-forms/pull/244

webhdx:
Yes, I mistook the PRs with my first comment. Sorry.

lserwatka:
@ViniTou @webhdx we have to move forward with this fix and it is blocking Behat tests. What was the conclusion for fix here? Is current version acceptable?

src/lib/RepositoryForms/Form/Processor/Content/UrlRedirectProcessor.php
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        /** @var \Symfony\Component\HttpFoundation\RedirectResponse $response */
+        $response = $event->getResponse();
+
+        if (!$response instanceof RedirectResponse) {
+            return;
+        }
+
+        $targetUrlAlias = $this->URLAliasService->lookup(
```
webhdx: Can you make it fail safe? Check if the alias can be loaded first as someone could change the redirection in his processor to something different than Url Alias.

src/lib/RepositoryForms/Form/Processor/Content/UrlRedirectProcessor.php
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * @return bool
+     *
+     * @throws \EzSystems\EzPlatformAdminUi\Exception\InvalidArgumentException
+     */
+    protected function isAdminSiteAccess(): bool
+    {
+        $siteaccess = $this->requestStack->getCurrentRequest()->attributes->get('siteaccess');
```
webhdx: Get current siteaccess from the service `ezpublish.siteaccess`, works with autowiring as well: `\eZ\Publish\Core\MVC\Symfony\SiteAccess`.

lserwatka:
@ViniTou could you merge it up?

ViniTou:
@lserwatka 
:+1:

PR #580, opened 2018-07-26
EZP-29104:  Implemented UI for ImageAsset Field Type

adamwojs:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29104
| **Depends on** | https://github.com/ezsystems/repository-forms/pull/246, https://github.com/ezsystems/ezpublish-kernel/pull/2403
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

#### Checklist:
- [x] Extract translations
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/AssetController.php
+++ b/src/bundle/Controller/AssetController.php
@@ -1,0 +1,164 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformAdminUiBundle\Controller;
+
+use Exception;
+use eZ\Publish\Core\FieldType\Image\Value as ImageValue;
+use eZ\Publish\Core\FieldType\ImageAsset\AssetMapper as ImageAssetMapper;
+use EzSystems\EzPlatformAdminUi\Form\Data\Asset\ImageAssetUploadData;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Csrf\CsrfToken;
+use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
+use Symfony\Component\Translation\TranslatorInterface;
+use Symfony\Component\Validator\ConstraintViolationListInterface;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+class AssetController extends Controller
+{
+    const CSRF_TOKEN_HEADER = 'X-CSRF-Token';
+
+    const LANGUAGE_CODE_KEY = 'languageCode';
+    const FILE_KEY = 'file';
+
+    /** @var \Symfony\Component\Validator\Validator\ValidatorInterface */
+    private $validator;
+
+    /** @var \Symfony\Component\Security\Csrf\CsrfTokenManagerInterface */
+    private $csrfTokenManager;
+
+    /** @var \eZ\Publish\Core\FieldType\ImageAsset\AssetMapper */
+    private $imageAssetMapper;
+
+    /** @var \Symfony\Component\Translation\TranslatorInterface */
+    private $translator;
+
+    /**
+     * @param \Symfony\Component\Validator\Validator\ValidatorInterface $validator
+     * @param \Symfony\Component\Security\Csrf\CsrfTokenManagerInterface $csrfTokenManager
+     * @param \eZ\Publish\Core\FieldType\ImageAsset\AssetMapper $imageAssetMapper
+     * @param \Symfony\Component\Translation\TranslatorInterface $translator
+     */
+    public function __construct(
+        ValidatorInterface $validator,
+        CsrfTokenManagerInterface $csrfTokenManager,
+        ImageAssetMapper $imageAssetMapper,
+        TranslatorInterface $translator)
+    {
+        $this->validator = $validator;
+        $this->csrfTokenManager = $csrfTokenManager;
+        $this->imageAssetMapper = $imageAssetMapper;
+        $this->translator = $translator;
+    }
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     *
+     * @throws \eZ\Publish\Core\Base\Exceptions\InvalidArgumentType
+     */
+    public function uploadImageAction(Request $request): Response
+    {
+        if ($this->isValidCsrfToken($request)) {
+            $data = new ImageAssetUploadData(
+                $request->files->get(self::FILE_KEY),
+                $request->request->get(self::LANGUAGE_CODE_KEY)
+            );
+
+            $errors = $this->validator->validate($data);
+            if ($errors->count() === 0) {
+                try {
+                    $file = $data->getFile();
+
+                    $content = $this->imageAssetMapper->createAsset(
+                        $file->getClientOriginalName(),
+                        new ImageValue([
+                            'path' => $file->getRealPath(),
+                            'fileSize' => $file->getSize(),
+                            'fileName' => $file->getClientOriginalName(),
+                            'alternativeText' => null,
+                        ]),
+                        $data->getLanguageCode()
+                    );
+
+                    return new JsonResponse([
+                        'destinationContent' => [
+                            'id' => $content->contentInfo->id,
+                            'name' => $content->getName(),
+                            'locationId' => $content->contentInfo->mainLocationId,
+                        ],
+                        'value' => $this->imageAssetMapper->getAssetValue($content),
+                    ]);
+                } catch (Exception $e) {
+                    return $this->createGenericErrorResponse($e->getMessage());
+                }
+            } else {
+                return $this->createInvalidInputResponse($errors);
+            }
+        }
+
+        return $this->createInvalidCsrfResponse();
+    }
+
+    /**
+     * @return \Symfony\Component\HttpFoundation\JsonResponse
+     */
+    private function createInvalidCsrfResponse(): JsonResponse
+    {
+        $errorMessage = $this->translator->trans(
+/** @Desc("Missing or invalid CSRF token") */ 'asset.upload.invalid_csrf', [], 'assets'
+        );
+
+        return $this->createGenericErrorResponse($errorMessage);
+    }
+
+    /**
+     * @param \Symfony\Component\Validator\ConstraintViolationListInterface $errors
+     *
+     * @return \Symfony\Component\HttpFoundation\JsonResponse
+     */
+    private function createInvalidInputResponse(ConstraintViolationListInterface $errors): JsonResponse
+    {
+        $errorMessages = [];
+        foreach ($errors as $error) {
+            $errorMessages[] = $error->getMessage();
+        }
+
+        return $this->createGenericErrorResponse(implode(', ', $errorMessages));
+    }
+
+    /**
+     * @param string $errorMessage
+     *
+     * @return \Symfony\Component\HttpFoundation\JsonResponse
+     */
+    private function createGenericErrorResponse(string $errorMessage): JsonResponse
+    {
+        return new JsonResponse([
+            'status' => 'failed',
+            'error' => $errorMessage,
+        ]);
+    }
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     *
+     * @return bool
+     */
+    private function isValidCsrfToken(Request $request): bool
+    {
+        $csrfTokenValue = $request->headers->get(self::CSRF_TOKEN_HEADER);
+
+        return $this->csrfTokenManager->isTokenValid(
+            new CsrfToken('authenticate', $csrfTokenValue)
+        );
+    }
+}
--- a/src/bundle/Resources/config/universal_discovery_widget.yml
+++ b/src/bundle/Resources/config/universal_discovery_widget.yml
@@ -37,3 +37,6 @@ system:
                     multiple: false
                     content_on_the_fly:
                         allowed_content_types: ['image']
+                image_asset:
+                    multiple: false
+                    visible_tabs: ['browse', 'search', 'bookmarks']
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
@@ -1,0 +1,276 @@
+(function(global, doc, eZ, React, ReactDOM, Translator) {
+    const SELECTOR_FIELD = '.ez-field-edit--ezimageasset';
+    const SELECTOR_INPUT_FILE = 'input[type="file"]';
+    const SELECTOR_INPUT_DESTINATION_CONTENT_ID = '.ez-data-source__destination-content-id';
+    const SELECTOR_LABEL_WRAPPER = '.ez-field-edit__label-wrapper';
+    const token = doc.querySelector('meta[name="CSRF-Token"]').content;
+    const siteaccess = doc.querySelector('meta[name="SiteAccess"]').content;
+    const showErrorNotification = eZ.helpers.notification.showErrorNotification;
+    const showSuccessNotification = eZ.helpers.notification.showSuccessNotification;
+    const getJsonFromResponse = eZ.helpers.request.getJsonFromResponse;
+    const imageAssetMapping = eZ.adminUiConfig.imageAssetMapping;
+
+    /**
+     * Handles response status
+     *
+     * @function handleResponseStatus
+     * @param {Response} response
+     * @returns {Error|Promise}
+     */
+    const handleResponseStatus = (response) => {
+        if (response.status === 'failed') {
+            throw new Error(response.error);
+        }
+
+        return response;
+    };
+
+    class EzImageAssetPreviewField extends eZ.BasePreviewField {
+        /**
+         * Creates a new Image Asset
+         *
+         * @method createAsset
+         * @param {File} file
+         * @param {String} languageCode
+         */
+        createAsset(file, languageCode) {
+            const assetCreateUri = global.Routing.generate('ezplatform.asset.upload_image');
+            const form = new FormData();
+
+            form.append('languageCode', languageCode);
+            form.append('file', file);
+
+            const options = {
+                method: 'POST',
+                headers: {
+                    Accept: 'application/json',
+                    'X-CSRF-Token': token,
+                },
+                body: form,
+                mode: 'same-origin',
+                credentials: 'same-origin',
+            };
+
+            this.toggleLoading(true);
+
+            fetch(assetCreateUri, options)
+                .then(getJsonFromResponse)
+                .then(handleResponseStatus)
+                .then(this.onAssetCreateSuccess.bind(this))
+                .catch(this.onAssetCreateFailure.bind(this));
+        }
+
+        /**
+         * Handle a successfully created Image Asset
+         *
+         * @method onAssetCreateSuccess
+         * @param {Object} response
+         */
+        onAssetCreateSuccess(response) {
+            const destinationContent = response.destinationContent;
+
+            this.updateData(destinationContent.id, destinationContent.name, destinationContent.locationId, response.value);
+            this.toggleLoading(false);
+
+            showSuccessNotification(
+                Translator.trans(
+                    /* @Desc("Image has been published and can now be reused") */ 'ezimageasset.create.message.success',
+                    {},
+                    'fieldtypes_edit'
+                )
+            );
+        }
+
+        /**
+         * Handle a failure while creating Image Asset
+         *
+         * @method onAssetCreateFailure
+         */
+        onAssetCreateFailure(error) {
+            const message = Translator.trans(
+                /* @Desc("Error while creating image asset: %error%") */ 'ezimageasset.create.message.error',
+                { error: error.message },
+                'fieldtypes_edit'
+            );
+
+            this.toggleLoading(false);
+            showErrorNotification(message);
+        }
+
+        /**
+         * Loads selected Image Asset
+         *
+         * @method loadAsset
+         * @param {Object} response
+         */
+        loadAsset(response) {
+            const imageField = response.CurrentVersion.Version.Fields.field.find((field) => {
+                return field.fieldDefinitionIdentifier === imageAssetMapping['contentFieldIdentifier'];
+            });
+
+            this.updateData(response.ContentInfo.Content._id, response.ContentInfo.Content.Name, response.id, imageField.fieldValue);
+        }
+
+        /**
+         * Toggle visibility of the loading spinner
+         *
+         * @method toggleLoading
+         * @param {boolean} show
+         */
+        toggleLoading(show) {
+            this.fieldContainer.classList.toggle('ez-field-edit--is-preview-loading', show);
+        }
+
+        /**
+         * Updates Image Asset preview data
+         *
+         * @method updateData
+         * @param {Number} destinationContentId
+         * @param {String} destinationContentName
+         * @param {Number} destinationLocationId
+         * @param {Object} image
+         */
+        updateData(destinationContentId, destinationContentName, destinationLocationId, image) {
+            const preview = this.fieldContainer.querySelector('.ez-field-edit__preview');
+            const previewImg = preview.querySelector('.ez-field-edit-preview__media');
+            const previewActionPreview = preview.querySelector('.ez-field-edit-preview__action--preview');
+            const assetNameContainer = preview.querySelector('.ez-field-edit-preview__asset-name a');
+            const destinationLocationUrl = global.Routing.generate('_ezpublishLocation', {
+                locationId: destinationLocationId
+            });
+
+            previewImg.setAttribute('src', image ? image.uri : '://0');
+            previewImg.classList.toggle('d-none', image === null);
+            previewActionPreview.setAttribute('href', destinationLocationUrl);
+            assetNameContainer.innerHTML = destinationContentName;
+            assetNameContainer.setAttribute('href', destinationLocationUrl);
+
+            this.inputDestinationContentId.value = destinationContentId;
+            this.inputField.value = '';
+            this.showPreview();
+        }
+
+        /**
+         * Open UDW to select an existing Image Asset
+         *
+         * @method openUDW
+         * @param {Event} event
+         */
+        openUDW(event) {
+            event.preventDefault();
+
+            const udwContainer = doc.getElementById('react-udw');
+            const config = JSON.parse(event.currentTarget.dataset.udwConfig);
+            const title = Translator.trans(/*@Desc("Select Image Asset")*/ 'ezimageasset.title', {}, 'universal_discovery_widget');
+            const closeUDW = () => ReactDOM.unmountComponentAtNode(udwContainer);
+            const onCancel = closeUDW;
+            const onConfirm = (items) => {
+                closeUDW();
+                this.loadAsset(items[0]);
+            };
+            const canSelectContent = ({ item }, callback) => {
+                const itemContentType = item.ContentInfo.Content.ContentTypeInfo.identifier;
+                const isAllowedContentType = itemContentType === imageAssetMapping['contentTypeIdentifier'];
+
+                callback(isAllowedContentType);
+            };
+
+            ReactDOM.render(
+                React.createElement(
+                    eZ.modules.UniversalDiscovery,
+                    Object.assign(
+                        {
+                            onConfirm,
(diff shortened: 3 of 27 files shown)


src/bundle/Resources/config/universal_discovery_widget.yml
```diff
@@ -37,3 +37,6 @@ system:
                     multiple: false
                     content_on_the_fly:
                         allowed_content_types: ['image']
+                image_asset:
+                    multiple: false
+                    visible_tabs: ['browse', 'search', 'bookmarks']
```
dew326: Why you should not be able to create content while picking from UDW?

src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
```diff
@@ -0,0 +1,240 @@
+(function (global, doc, eZ, React, ReactDOM, Translator) {
+    const SELECTOR_FIELD = '.ez-field-edit--ezimageasset';
+    const SELECTOR_INPUT_FILE = 'input[type="file"]';
+    const SELECTOR_LABEL_WRAPPER = '.ez-field-edit__label-wrapper';
+    const SELECTOR_INPUT_DESTINATION_CONTENT_ID = '.ez-field-edit__destination_content_id';
+
```
dew326: No need to separate `const`

src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
```diff
@@ -0,0 +1,240 @@
...
+    const token = doc.querySelector('meta[name="CSRF-Token"]').content;
+    const siteaccess = doc.querySelector('meta[name="SiteAccess"]').content;
+
+    const showErrorNotification = eZ.helpers.notification.showErrorNotification;
+    const showSuccessNotification = eZ.helpers.notification.showSuccessNotification;
+    const imageAssetMapping = eZ.adminUiConfig.imageAssetMapping;
+
+    class EzImageAssetPreviewField extends eZ.BasePreviewField {
+
```
dew326: Please remove empty line

src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
```diff
@@ -0,0 +1,240 @@
...
+        /**
+         * Creates a new Image Asset
+         *
+         * @method createImage
+         * @param {File} file
+         * @param {String} languageCode
+         */
+        createImage(file, languageCode) {
+            const assetCreateUri = '/asset/image';
```
dew326: Should be generate by the JS routing

src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
```diff
@@ -0,0 +1,240 @@
...
+                    'X-Siteaccess': siteaccess,
+                    'X-CSRF-Token': token
+                },
+                body: form,
+                mode: 'same-origin',
+                credentials: 'same-origin'
+            };
+
+            this.toggleLoading(true);
+            fetch(assetCreateUri, options).then((response) => {
```
dew326: Please move this function to some method

src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
```diff
@@ -0,0 +1,240 @@
...
+                    const message = Translator.trans(
+                        /* @Desc("Error while creating image asset") */  'ezimageasset.create.message.error', {}, 'fieldtypes_edit'
+                    );
+
+                    throw Error(message);
+                }
+
+                return response.json();
+            }).then((response) => {
+                showSuccessNotification(
```
dew326: The same, please create a method.

src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
```diff
@@ -0,0 +1,240 @@
...
+         * Open UDW to select an existing Image Asset
+         *
+         * @method openUDW
+         * @param {Event} event
+         */
+        openUDW(event) {
+            const udwContainer = doc.getElementById('react-udw');
+            const config = JSON.parse(event.currentTarget.dataset.udwConfig);
+            const title = Translator.trans(/*@Desc("Select Image Asset")*/ 'ezimageasset.title', {}, 'universal_discovery_widget');
+
```
dew326: remove empty line

src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
```diff
@@ -0,0 +1,240 @@
...
+                this.loadPreview(items[0]);
+            };
+            const canSelectContent = ({item, itemsCount}, callback) => {
+                const itemContentType = item.ContentInfo.Content.ContentTypeInfo.identifier;
+                const isAllowedContentType = itemContentType === imageAssetMapping['content_type_identifier'];
+
+                callback(isAllowedContentType);
+            };
+
+            ReactDOM.render(React.createElement(eZ.modules.UniversalDiscovery, Object.assign(config, {
```
dew326: I think this should be the other way, the `config` should be able to override you settings

src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
```diff
@@ -0,0 +1,240 @@
...
+                onConfirm,
+                onCancel,
+                canSelectContent,
+                title,
+                multiple: false,
```
dew326: I think you have this in `config`

src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
```diff
@@ -0,0 +1,240 @@
...
+                startingLocationId: parseInt(imageAssetMapping['parent_location_id'], 10),
```
dew326: This can't be moved to the `config`?

src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
```diff
@@ -0,0 +1,240 @@
...
+        /**
+         * Initializes the preview
+         *
+         * @method init
+         */
+        init() {
+            super.init();
+
+            this.btnSelect = this.fieldContainer.querySelector('.ez-data-source__btn-select');
+            this.btnSelect.addEventListener('click', (event) => {
```
dew326: I do not see a point of this middle function which will only `preventDefault`, you can do it in `openUDW`

src/bundle/Resources/public/scss/fieldType/edit/_ezimageasset.scss
```diff
@@ -0,0 +1,78 @@
...
+
+    &__details {
+      color: $ez-color-base-light;
+      font-size: .75rem;
+      padding-top: .5rem;
+    }
+  }
+
+  .ez-field-edit__spinner {
+
```
dew326: this can be remove

src/bundle/Resources/public/scss/fieldType/edit/_ezimageasset.scss
```diff
@@ -0,0 +1,78 @@
+.ez-field-edit--ezimageasset {
+  .ez-field-edit-preview {
+    padding: 1.25rem 0 1.25rem 1.25rem;
```
dew326: Indentation 4 spaces

src/bundle/Resources/public/scss/fieldType/edit/_ezimageasset.scss
```diff
@@ -0,0 +1,78 @@
...
+        text-align: center;
+        font-size: 1.5rem;
+        border-top-left-radius: 5px;
+        border-bottom-left-radius: 5px;
+      }
+    }
+  }
+}
+
+.ez-field-edit {
```
dew326: This will override styles for all fields. Cannot be done like this.

src/bundle/Resources/config/universal_discovery_widget.yml
adamwojs: Good question. I going to check this with PM team 😉

adamwojs:
PR rebased and updated according to @dew326 suggestions.

src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
```diff
@@ -0,0 +1,248 @@
...
+            const udwContainer = doc.getElementById('react-udw');
+            const config = JSON.parse(event.currentTarget.dataset.udwConfig);
+            const title = Translator.trans(/*@Desc("Select Image Asset")*/ 'ezimageasset.title', {}, 'universal_discovery_widget');
+            const closeUDW = () => ReactDOM.unmountComponentAtNode(udwContainer);
+            const onCancel = closeUDW;
+            const onConfirm = (items) => {
+                closeUDW();
+                this.loadAsset(items[0]);
+            };
+            const canSelectContent = ({item, itemsCount}, callback) => {
```
dew326: You are not using `itemsCount`, can be removed.

src/bundle/Resources/public/scss/fieldType/edit/_ezimageasset.scss
```diff
@@ -0,0 +1,74 @@
...
+        &__media {
+            max-width: 100%;
+            width: auto;
+            height: auto;
+            display: block;
+        }
+
+        &__details {
+            color: $ez-color-base-light;
+            font-size: .75rem;
```
dew326: `0.75rem`, Prettier adding zero at the beginning. Please install Prettier and run on all JS and CSS files.

src/bundle/Resources/public/scss/fieldType/edit/_ezimageasset.scss
```diff
@@ -0,0 +1,74 @@
...
+                text-align: center;
+                font-size: 1.5rem;
+                border-top-left-radius: 5px;
+                border-bottom-left-radius: 5px;
+            }
+        }
+    }
+}
+
+.ez-field-edit--ezimageasset {
```
dew326: This can be combined with the rules above

src/bundle/Resources/public/scss/fieldType/edit/_ezimageasset.scss
```diff
@@ -0,0 +1,74 @@
...
+    &__preview-wrapper {
```
dew326: I think this will be compiled to: `ez-field-edit--ezimageasset__preview-wrapper` and this is not correct usage of BEM

src/bundle/Resources/public/scss/fieldType/preview/_ezimageasset.scss
```diff
@@ -0,0 +1,44 @@
+.ez-field-preview--ezimageasset {
+    display: flex;
+
+    .ez-imageasset__wrapper {
```
dew326: Seems like not correct BEM

src/bundle/Controller/AssetController.php
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @throws \eZ\Publish\Core\Base\Exceptions\InvalidArgumentType
+     */
+    public function uploadImageAction(Request $request): Response
+    {
+        $data = new AssetUploadData();
+
+        $form = $this->formFactory->assetUpload($data, 'form');
+        $form->handleRequest($request);
+        if ($form->isValid()) {
```
webhdx: I believe it's advised by Symfony to use `$form->isSubmitted() && $form->isValid()` together.

src/bundle/Resources/views/fieldtypes/edit/ezimageasset.html.twig
```diff
@@ -0,0 +1,126 @@
...
+        <div{% with { attr: widget_wrapper_attr } %}{{ block('attributes') }}{% endwith %}>
+            <div class="ez-data-source">
+                {{ block('ezplatform_fieldtype_ezimageasset_widget') }}
+            </div>
+        </div>
+    </div>
+{% endblock %}
+
+{% block ezplatform_fieldtype_ezimageasset_widget %}
+    <div class="ez-data-source__message--main">
```
dew326: Seems like not correct BEM usage

src/bundle/Resources/views/fieldtypes/edit/ezimageasset.html.twig
```diff
@@ -0,0 +1,126 @@
...
+        <svg class="ez-icon">
+            <use xlink:href="/bundles/ezplatformadminui/img/ez-icons.svg#browse"></use>
+        </svg>
+        <span class="ez-data-source__btn-label">
+            {{ 'fieldtype.ezimageasset.select.label'|trans|desc('Select from your repository') }}
+        </span>
+    </div>
+
+    {% if max_file_size is defined and max_file_size > 0 %}
+        <div class="ez-data-source__message--filesize">
```
dew326: The same, not correct BEM

src/bundle/Controller/AssetController.php
```diff
@@ -0,0 +1,75 @@
...
+
+use eZ\Publish\Core\FieldType\Image\Value as ImageValue;
+use eZ\Publish\Core\FieldType\ImageAsset\Mapper as ImageAssetMapper;
+use EzSystems\EzPlatformAdminUi\Form\Data\Asset\AssetUploadData;
+use EzSystems\EzPlatformAdminUi\Form\Factory\FormFactory;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+class AssetController extends Controller
```
webhdx: I doubt it's a good idea to keep the Controller in AdminUI - what if we need this fieldtype to be available for frontend editing?

src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
```diff
@@ -0,0 +1,248 @@
...
+            const form = new FormData();
+
+            form.append('form[languageCode]', languageCode);
+            form.append('form[file]', file);
+
+            const options = {
+                method: 'POST',
+                headers: {
+                    'Accept': 'application/json',
+                    //'X-CSRF-Token': token
```
webhdx: Commented out code should not be commited.

src/bundle/Resources/views/admin/content_type/field_types.html.twig
```diff
@@ -223,3 +223,6 @@
         {{- form_row(form.textRows) -}}
     </div>
 {% endblock %}
+
+{% block ezimageasset_field_definition_edit %}
```
webhdx: Is an empty block needed?

src/bundle/Resources/views/fieldtypes/edit/ezimageasset.html.twig
```diff
@@ -0,0 +1,126 @@
...
+
+    {% if file_is_empty or has_error %}
+        {% set preview_attr = preview_attr|default({})|merge({'hidden': 'hidden'}) %}
+    {% else %}
+        {% set widget_wrapper_attr = widget_wrapper_attr|merge({'hidden': 'hidden'}) %}
+    {% endif %}
+
+    {% set widget_container_block = block('ezplatform_fieldtype_ezimageasset_widget_container') %}
+
+    {{ block('form_row_fieldtype') }}
```
webhdx: You are duplicating a lot of variables from block `binary_base_row` (defined in `binary_base.html.twig`). Can't you override only some of the variables (it might require to do some changes to `binary_base_row` to make it possible) and display `binary_base_row`?

src/bundle/Resources/views/fieldtypes/preview/ezimageasset.html.twig
```diff
@@ -0,0 +1,52 @@
...
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td>{{ 'ezimageasset.file_name'|trans|desc('File name') }}:</td>
+            <td>{{ field.value.fileName }}</td>
+        </tr>
+        <tr>
+            <td>{{ 'ezimageasset.size'|trans|desc('Size') }}:</td>
+            <td>{{ field.value.fileSize|ez_file_size( 1 ) }}</td>
```
webhdx: Spaces no needed: `ez_file_size(1)`

src/bundle/Controller/AssetController.php
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     *
+     * @throws \eZ\Publish\Core\Base\Exceptions\InvalidArgumentType
+     */
+    public function uploadImageAction(Request $request): Response
+    {
+        $data = new AssetUploadData();
+
+        $form = $this->formFactory->assetUpload($data, 'form');
```
webhdx: Name `form` feels too generic to me.

src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
```diff
@@ -0,0 +1,248 @@
...
+         *
+         * @method createAsset
+         * @param {File} file
+         * @param {String} languageCode
+         */
+        createAsset(file, languageCode) {
+            const assetCreateUri = global.Routing.generate('ezplatform.asset.upload');
+            const form = new FormData();
+
+            form.append('form[languageCode]', languageCode);
```
webhdx: You shouldn't manually compose Data object here. Please use existing form:

```js
        const request = new Request(form.action, {
            method: 'POST',
            body: new FormData(form),
            mode: 'same-origin',
            credentials: 'same-origin',
        });
```

dew326:
I have been thinking about the whole idea of using a form in this case and this makes no sense.
As far as I can see you are creating a form, rendering only a few fields and send two fields in POST request. The endpoint gets the form and returning JSON response (so there is no possibility to submit this form in a regular way). Thus we do not need this form at all.

This should be done like this:
- no form
- a regular POST endpoint which takes in the body: 
```js
{
    languageCode: languageCode,
    file: file
}
```
- JS sending POST with the body and CSRF token

webhdx:
Totally overlooked how it's really done yesterday, sorry. I second what @dew326 said. Using Form is convenient, especially when validation and creating data object is needed but in your case it's totally misused. It works but it will be confusing for anyone who maintains the code in the future.

src/bundle/Controller/AssetController.php
```diff
@@ -0,0 +1,75 @@
...
+
+            return new JsonResponse([
+                'status' => 'success',
+                'destinationContentId' => $content->contentInfo->id,
+                'value' => $this->imageAssetMapper->getAssetValue($content),
+            ]);
+        }
+
+        return new JsonResponse([
+            'status' => 'error',
```
mikadamczyk: Is this status field really needed?

src/lib/Form/Data/Asset/AssetUploadData.php
```diff
@@ -0,0 +1,64 @@
...
+     * @param \Symfony\Component\HttpFoundation\File\UploadedFile|null $file
+     * @param string|null $languageCode
+     */
+    public function __construct(?UploadedFile $file = null, string $languageCode = null)
+    {
+        $this->file = $file;
+        $this->languageCode = $languageCode;
+    }
+
+    public function getFile(): ?UploadedFile
```
mikadamczyk: missing docblock here and below

src/bundle/Resources/views/fieldtypes/preview/ezimageasset.html.twig
```diff
@@ -0,0 +1,52 @@
+{% trans_default_domain 'fieldtypes_preview' %}
+
+{% set field = ez_field(content, ez_image_asset_content_field_identifier()) %}
+{% set versionInfo = content.versionInfo  %}
+{% set imageAlias = ez_image_alias( field, versionInfo, parameters.alias|default( 'original' ) ) %}
+{% set src = imageAlias ? asset( imageAlias.uri ) : "//:0" %}
+
+<div class="ez-imageasset__wrapper">
```
dew326: `ez-field-preview__image-wrapper`

adamwojs:
PR ready for the final code review @webhdx @dew326  @mikadamczyk

src/bundle/Resources/views/fieldtypes/edit/ezimageasset.html.twig
```diff
@@ -0,0 +1,106 @@
...
+        </span>
+    </div>
+
+    {% if max_file_size is defined and max_file_size > 0 %}
+        <div class="ez-data-source__message--filesize">
+            {{ 'fieldtype.binary_base.max_file_size'|trans({'%size%': max_file_size|ez_file_size(2)})|desc('Max file size: %size%') }}
+        </div>
+    {% endif %}
+
+    {{- form_widget(form.file, {'attr': attr}) -}}
```
dew326: I think we drop the idea of creating a form for this?

src/bundle/Resources/views/fieldtypes/edit/binary_base.html.twig
```diff
@@ -20,7 +22,10 @@
         {% set widget_wrapper_attr = widget_wrapper_attr|merge({'hidden': 'hidden'}) %}
     {% endif %}
 
-    {% set widget_container_block = block('binary_base_widget_container') %}
+    {% if not widget_container_block_name is defined %}
```
webhdx: Why not `{% if widget_container_block_name is not defined %}`? Looks better and reads better.

webhdx:
I need some clarification on the controller uploading image assets being a part of AdminUI - I don't think this is a good move if we want to bring UGC in the near future.

barbaragr:
**I Error 500 after deleting Image used in ImageAssets field**
1. Create Content Type with ImageAsset FieldType
2. Go to Content/Content structure and create new Content Item based on Content Type from step 1
3. Upload an image and publish
4. Go to Media/Images and send to Trash image from previous step
5. Go back to Content/Content structure and go to view mode of recently published CI
**Actual result**: error 500: `Type error: Argument 1 passed to eZ\Publish\Core\SignalSlot\BookmarkService::isBookmarked() must be an instance of eZ\Publish\API\Repository\Values\Content\Location, null given, called in /var/www/ezplatformee/vendor/ezsystems/ezplatform-admin-ui/src/bundle/Controller/ContentViewController.php on line 366`

**II Error 500 after editing added image in Media/Image**
1. Create Content Type with ImageAsset FieldType
2. Go to Content/Content structure and create new Content Item based on Content Type from step 1
3. Upload an image and publish
4. Go to Media/Images and find the image and click Edit button to go to edit mode
5. Remove the image from field Image (Trash button) and publish
6. Go back to Content/Content structure and go to view mode of recently published CI
**Actual result:** Error 500: `An exception has been thrown during the rendering of a template ("Argument 'BinaryFile::id' is invalid: 'NULL' is wrong value in class 'BinaryFile'").`

**III Question:**
When I'm creating a CI with ImageAssets and I have uploaded an image, but next I remove it using TRash button, should it stay in images repository? Specifications says only about uploading the image and saving a draft. In my situation I saved nothing, so maybe it shouldn't be publish in Media/Images yet?

**IV Question:**
For scenario: 
When I create a new Article 
And decide to select an existing image for the Photo field 
And try to select any content item which is not an Image content type 
Then I can not select the content item 
Now I can select, but I can't confirm. Maybe It could be disabled like some CIs are disabled in UDW when user want to swap locations (containers vs non containers). But I'm just asking :)

SylvainGuittard:
Thanks @barbaragr for testing.

**I Error 500 after deleting Image used in ImageAssets field**
I updated the specifications. Please check it again.

**II Error 500 after editing added image in Media/Image**
This field is not mandatory in the `Image` content type. So the best solution is to not display the image but the name of the content item. We keep the `Open in a new tab` link for the user to edit it again.

**III Question:**
The image should be deleted

**IV Question:**
> Maybe It could be disabled like some CIs are disabled in UDW when user want to swap locations (containers vs non containers).

Totally agree.

adamwojs:
@barbaragr , @SylvainGuittard 

1) This problem is already solved by https://github.com/ezsystems/ezplatform-admin-ui/pull/581. Please  confirm it by cherry picking `501190d8ec6d88bffb37ed6455ebee89e78a3975`. 

2) I already handled this case by displaying _Image asset is not available message but  definitely it might be improved. I see what it can do :wink: 

zrzut ekranu z 2018-09-05 10-08-04

3) This is edge case but worth to mention. As the image content item is created and published immediately after upload, this gives a time frame to reuse it in other content. if "Trash" button in content editor will delete related content item so it need same confirmation modal window as impl. as part of https://github.com/ezsystems/ezplatform-admin-ui/pull/591 and it need to be permission aware.  

PS. I will update PR today :wink:

ViniTou:
I can confirm that 1 is solved just by rebasing.

adamwojs:
PR updated according to @barbaragr suggestions and ready to second round of QA.  

2) I decided to handle this issue as @SylvainGuittard suggested. Now when image is empty, the content name will be displayed: 

image_value_is_empty

When the related content item is deleted or current user doesn't have permission to see it the the following message will be display:

image_asset_is_not_available

From technical POV, I implemented `Parameter Provider` (similar as in case of `ezrelation` and `ezrelation list` field types) in which pass `available` parameter to FT block. See https://github.com/ezsystems/ezpublish-kernel/pull/2403/files#diff-a69bf20f91185b4b1c044ed1b71d5bceR41

AD3) No changes here. 

AD4) No changes here. I'm missing proper API in UDW to be able to do this.

AD5) Behavior has been aligned with the `ezimage` field type. You cannot drop file which are not images into `ezimageasset` FT, also proper filter is now active by default in the file select dialog. 

When you try to force upload of non-image file the "The file is not valid image" error will be displayed:   

file_is_not_a_valid_image

AD6) In this case the following error will be displayed: 
insufficent_permissions

barbaragr:
**I Image is still visible after deleting Image used in ImageAssets field**

1. Create Content Type with ImageAsset FieldType
2. Go to Content/Content structure and create new Content Item based on Content Type from step 1
3. Upload an image and publish
4. Go to Media/Images and send to Trash image from previous step
5. Go back to Content/Content structure and go to view mode of recently published CI
*Actual result:* In view mode user sees a message: `Image asset is not available (related content has been deleted or insufficient permissions)` but if he go to the Edit mode, he can still see the image in image asset field.

barbaragr:
**AD VI Question:**
User receives a message: `Error while creating image asset: User does not have access to 'read' 'content'` It could by misleading, because, as default, editor has got all content policies and assign limitations for Home subtree. Maybe the information message  should be more general?

barbaragr:
**VII I can't create Landing Page**
1. Go to Content/Content structure
2. Click Create and choose Landing Page
**Actual result:** There is no layout modal and user is unable to publish LP.  Console shows: `Uncaught TypeError: window.eZ.addConfig is not a function` and  `Uncaught TypeError: Cannot read property 'config' of undefined`

SylvainGuittard:
> Actual result: According to @SylvainGuittard and @adamwojs in view mode content name is displayed (with a link to open in a new tab), but when I go to edit mode ezimageasset field is empty. When I change nothing and click Publish content name with a link is still visible. Are we OK with that? Or should it be overridden and This field is empty message should be displayed?

Displaying `This field is empty message` might confuse editors. Is it the field of the current content item or the content item image? I think we should include a link like in view mode.
image

barbaragr:
**VIII Error when previewing LP after sending to trash image from image asset field**
1. Create Content Type with ImageAsset FieldType and Landing Page
2. Go to Content/Content structure and create new Content Item based on Content Type from step 1
3. Upload an image and publish
4. Go to Media/Images and send to Trash image from previous step
5. Go back to Content/Content structure and CI created in step 3
6. Click Page/site to see page view
**Actual result:** There is an error: `An exception has been thrown during the rendering of a template ("Parameter "locationId" for route "_ezpublishLocation" must match "[^/]++" ("" given) to generate a corresponding URL.").`

src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
```diff
@@ -0,0 +1,276 @@
...
+         */
+        onAssetCreateFailure(error) {
+            const message = Translator.trans(
+                /* @Desc("Error while creating image asset: %error%") */ 'ezimageasset.create.message.error',
+                { error: error.message },
+                'fieldtypes_edit'
+            );
+
+            this.toggleLoading(false);
+            showErrorNotification({ message });
```
barbaragr: I think that these curly braces are needless. It should be:
```js
showErrorNotification(message);
```

barbaragr:
All right! To sum up:
AD I - OK
AD II - OK
AD III - works fine, updated spec (_Discard of the parent content item using the image asset_)
AD IV - follow up: https://jira.ez.no/browse/EZP-29589
AD V - OK
AD VI - Probably this will fix it https://github.com/ezsystems/ezpublish-kernel/pull/2408 (I will check after merge)
AD VII - OK
AD VIII - OK

I didn't check DBP (blocked by: https://jira.ez.no/browse/EZEE-2308) and translations (blocked by: https://jira.ez.no/browse/EZEE-2316)

PR #588, opened 2018-08-06
[WIP] EZP-29269: User draft is not discarded on edition cancel

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       |  https://jira.ez.no/browse/EZP-29269
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no?

This is part of introducing new route for creating and editing users. Instead of treating them as content all the way.

#### Checklist:
- [ ] Ready for Code Review


src/bundle/Controller/User/UserEditController.php
```diff
@@ -0,0 +1,59 @@
...
+        $this->submitHandler = $submitHandler;
+    }
+
+    public function userEditAction(Request $request): Response
+    {
+        $form = $this->formFactory->editUser(null);
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            $result = $this->submitHandler->handle($form, function (UserEditData $data) {
```
mikadamczyk: `handle()` contain `isValidate()` method

src/bundle/Resources/config/routing.yml
```diff
@@ -553,6 +553,12 @@ ezplatform.link_manager.view:
 # User
 #
 
+ezplatform.user.edit:
+    path: /user/edit/
```
mikadamczyk: `/user/edit/` without last `/`

src/bundle/Resources/views/content/widgets/user_edit.html.twig
```diff
@@ -0,0 +1,16 @@
+{% trans_default_domain 'locationview' %}
+
+<div class="ez-extra-actions ez-extra-actions--edit-user ez-extra-actions--hidden bg-secondary {% if form.language.vars.choices|length == 1 %} ez-extra-actions--prevent-show {% endif %}" data-actions="edit-user">
```
mikadamczyk: This line is too long

lserwatka:
@ViniTou rebase is needed

src/bundle/Resources/public/js/scripts/sidebar/btn/user.edit.js
```diff
@@ -0,0 +1,16 @@
+(function (global, doc) {
+    const editButton = doc.querySelector('.ez-btn--edit-user');
+    const languageRadioOption = doc.querySelector('.ez-extra-actions--edit-user.ez-extra-actions--prevent-show [type="radio"]');
```
dew326: I think you cannot assume there is only one language. After discussing with @webhdx and @Nattfarinn it should be possible to tranlsate user. You should keep in mind there might be more then one language option.

src/bundle/Resources/public/js/scripts/sidebar/btn/user.edit.js
```diff
@@ -0,0 +1,16 @@
...
+    const editActions = doc.querySelector('.ez-extra-actions--edit-user');
+    const form = editActions.querySelector('form');
```
dew326: I think the `editActions` might be null in some cases or should be.

src/bundle/Resources/public/js/scripts/sidebar/btn/user.edit.js
```diff
@@ -0,0 +1,16 @@
...
+
+    if (!languageRadioOption) {
```
dew326: Why in edit form there is no language at all?

src/bundle/Resources/public/js/scripts/sidebar/btn/user.edit.js
```diff
@@ -0,0 +1,16 @@
...
+        return;
+    }
+
+    editButton.addEventListener('click', () => {
+        languageRadioOption.checked = true;
+        languageRadioOption.dispatchEvent(new CustomEvent('change'));
```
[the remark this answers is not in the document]
ViniTou: If I remember correctly, version conflict modal is listening on that

src/bundle/Resources/public/js/scripts/sidebar/btn/user.edit.js
ViniTou: There is only one language when there is `ez-extra-actions--prevent-show` class

src/bundle/Resources/public/js/scripts/sidebar/btn/user.edit.js
ViniTou: connected to `ez-extra-actions--prevent-show` class.

src/bundle/Resources/public/js/scripts/sidebar/btn/user.edit.js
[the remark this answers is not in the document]
dew326: Ok, so you should not do `form.submit()` and let to do it by the version conflict script.

src/bundle/Controller/User/UserEditController.php
```diff
@@ -0,0 +1,68 @@
...
+        $this->formFactory = $formFactory;
+        $this->submitHandler = $submitHandler;
+    }
+
+    /**
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function userEditAction(Request $request): Response
```
webhdx: You don't need a form just to provide redirection. For content it made sense because it would create a draft thus it needed some kind of protection from misclicks and CSRF. For users where you don't create drafts, you only display editing form you can use simple redirection. Less code less maintenance hassle.

src/bundle/Resources/config/routing.yml
```diff
@@ -553,6 +553,12 @@ ezplatform.link_manager.view:
 # User
 #
 
+ezplatform.user.edit:
```
webhdx: See my comment above.

src/bundle/Resources/views/content/tab/versions/table.html.twig
```diff
@@ -52,8 +52,8 @@
             {% if is_draft_conflict %}
                 <td class="ez-table__cell ez-table__cell--has-action-btns text-right">
                     {% set edit_draft_disabled = version.author.id != admin_ui_config.user.user.id %}
-                    <a href="{{ path('ez_content_draft_edit', { 'contentId': version.contentInfo.id, 'versionNo': version.versionNo, 'language': version.translations[0].languageCode, 'locationId': location.id }) }}"
-                       class="btn btn-icon mx-2 {% if edit_draft_disabled %}ez-btn--prevented{% endif %}"
+                    <a href="{{ path(edit_url, { 'contentId': version.contentInfo.id, 'versionNo': version.versionNo, 'language': version.translations[0].languageCode, 'locationId': location.id }) }}"
```
webhdx: It would be better to do generate needed URL and assign it to the variable to avoid issues when parameters to these routes change. They now have the same parameters but it might be changed sometime in the future.

src/lib/Form/Data/Content/Draft/ContentEditData.php
```diff
@@ -21,11 +21,7 @@ class ContentEditData
     /** @var Location|null */
     protected $location;
 
-    /**
-     * @deprecated Deprecated in 1.1 and will be removed in 2.0. Please use ContentEditData::$location instead.
-     *
-     * @var ContentInfo|null
-     * */
+    /** @var ContentInfo|null */
```
webhdx: FQCN

src/lib/Menu/ContentRightSidebarBuilder.php
```diff
@@ -144,6 +144,18 @@ public function createStructure(array $options): ItemInterface
...
         ];
+        $editUserAttributes = [
+            'class' => 'ez-btn--extra-actions ez-btn--edit-user',
+            'data-actions' => 'edit-user',
+        ];
+        $deleteAttributes = [
+            'data-toggle' => 'modal',
+            'data-target' => '#delete-user-modal',
+        ];
+        $sendToTrashAttributes = [
```
webhdx: Is `$sendToTrashAttributes` needed? I don't see any usage of it in your PR.

src/lib/Menu/ContentRightSidebarBuilder.php
```diff
@@ -169,16 +183,36 @@ public function createStructure(array $options): ItemInterface
...
-                    'extras' => ['icon' => 'edit'],
-                    'attributes' => $canEdit
-                        ? $editAttributes
-                        : array_merge($editAttributes, ['disabled' => 'disabled']),
-                ]
-            ),
-            self::ITEM__MOVE => $this->createMenuItem(
+        ]);
+
+        if ($contentIsUser) {
```
webhdx: Can you extract it to a method? It's becoming harder and harder to maintain. `addEditMenuItem()` is a good candidate.

src/lib/UserSetting/Setting/Timezone.php
```diff
@@ -81,7 +81,8 @@ public function mapFieldForm(FormBuilderInterface $formBuilder, ValueDefinitionI
     private function getTranslatedName(): string
     {
         return $this->translator->trans(
-            /** @Desc("Time Zone") */ 'settings.timezone.value.title',
+            /** @Desc("Time Zone") */
```
webhdx: Thanks :)

src/bundle/Resources/public/js/scripts/sidebar/btn/location.edit.js
```diff
@@ -1,8 +1,9 @@
 (function (global, doc, $) {
-    const editActions = doc.querySelector('.ez-extra-actions--edit');
+    const editActions = doc.querySelector('.ez-extra-actions--edit') || doc.querySelector('.ez-extra-actions--edit-user') ;
     const btns = [...editActions.querySelectorAll('.form-check [type="radio"]')];
     const form = editActions.querySelector('form');
-    const contentId = form.querySelector('#content_edit_content_info').value;
+    const contentIdSelector = form.querySelector('#content_edit_content_info') || form.querySelector('#user_edit_content_info');
```
dew326: This is not `selector` it's probably input so `contentIdInput`

src/bundle/Resources/public/js/scripts/sidebar/btn/location.edit.js
```diff
@@ -14,17 +15,31 @@
...
             [...wrapper.querySelectorAll('.ez-btn--prevented')].forEach(btn => btn.addEventListener('click', event => event.preventDefault(), false));
             $('#version-draft-conflict-modal').modal('show').on('hidden.bs.modal', resetRadioButtons);
         };
 
+        const changeFormActionUrl = () => {
+            if (!form.querySelector('#user_edit_version_info')) {
+                return;
+            }
+            const versionNo = form.querySelector('#user_edit_version_info_version_no').value;
+            const language = btns.filter(btn => btn.checked = true)[0].value;
```
dew326: oxmdj7p

should be `btn.checked === true` event better `btn.checked` and you can use `array.find` so in the end:
`const language = btns.find(btn => btn.checked).value;`

src/bundle/Resources/public/js/scripts/sidebar/btn/user.edit.js
```diff
@@ -0,0 +1,18 @@
+(function (global, doc) {
+    const editButton = doc.querySelector('.ez-btn--edit-user');
+    const languageRadioOption = doc.querySelector('.ez-extra-actions--edit-user.ez-extra-actions--prevent-show [type="radio"]');
+    const editActions = doc.querySelector('.ez-extra-actions--edit-user');
+
+    if (!editActions) {
+        return;
+    }
+
+    if (!languageRadioOption) {
```
dew326: those ifs can be in one `if (!editActions || !languageRadioOption) {`

src/bundle/Resources/views/content/modal_draft_conflict.html.twig
```diff
@@ -22,29 +22,37 @@
...
-                        {{ 'draft.conflict.choice'|trans({})|desc('You can either edit any of your existing draft(s) or add a new one.') }}
+                        {% if content_is_user %}
+                            {{ 'draft.conflict.choice_content_is_user'|trans({})|desc('You can edit any of your existing draft(s)') }}
+                        {% else %}
+                            {{ 'draft.conflict.choice'|trans({})|desc('You can either edit any of your existing draft(s) or add a new one.') }}
+                        {% endif %}
                     </p>
                 </div>
-                {% include '@ezdesign/parts/table_header.html.twig' with { headerText: null, tools: version_modal_draft_conflict.table_header_tools() } %}
+                {% set show_add_draft_button = (content_is_user != true) %}
```
dew326: nitkpick: `content_is_user is not true`?

src/bundle/Resources/public/js/scripts/sidebar/btn/location.edit.js
ViniTou: image

src/bundle/Resources/views/content/modal_draft_conflict.html.twig
ViniTou: I think I tried all combinations of those and for some reason this is the only one not resulting in syntax error.

src/bundle/Resources/public/js/scripts/sidebar/btn/location.edit.js
```diff
@@ -14,17 +15,31 @@
...
+            }
             [...wrapper.querySelectorAll('.ez-btn--prevented')].forEach(btn => btn.addEventListener('click', event => event.preventDefault(), false));
             $('#version-draft-conflict-modal').modal('show').on('hidden.bs.modal', resetRadioButtons);
         };
 
+        const redirectToUserEdit = () => {
+            const versionNo = form.querySelector('#user_edit_version_info_version_no').value;
+            const language = btns.find(btn => btn.checked).value;
+
+            window.location.replace(global.Routing.generate('ez_user_update', { contentId, versionNo, language }));
```
dew326: I don't like the `replace` here. The last item from the history will be replaced so the user will not be able to click back button in the browser (he will be redirected to the previous page).
I think better is just: 'window.location.href = url'

lserwatka:
@webhdx are you OK with final changes here? Can we proceed with testing?

PR #589, opened 2018-08-06
EZP-29498: showDangerNotification duplicates the functionality of showErrorNotification

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29498
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | -
| Doc needed?   | no

Currently there are 4 types of notifications: 'info', 'success', 'warning' & 'danger', but in fact 'danger' should be an 'error' notification. Thus `showDangerNotification` should be removed; `showErrorNotification` already exists.

*For QA*: notifications should work as before.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/helpers/notification.helper.js
+++ b/src/bundle/Resources/public/js/scripts/helpers/notification.helper.js
@@ -4,13 +4,13 @@
     const NOTIFICATION_INFO_LABEL = 'info';
     const NOTIFICATION_SUCCESS_LABEL = 'success';
     const NOTIFICATION_WARNING_LABEL = 'warning';
-    const NOTIFICATION_DANGER_LABEL = 'danger';
+    const NOTIFICATION_ERROR_LABEL = 'danger';
 
     /**
      * Dispatches notification event
      *
      * @function showNotification
-     * @param {Object} detail
+     * @param {{message: string, label: string}} detail
      */
     const showNotification = (detail) => {
         const event = new CustomEvent('ez-notify', { detail });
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/scripts/helpers/notification.helper.js
```diff
@@ -55,31 +55,27 @@
...
-    const showErrorNotification = (error) => showDangerNotification(error.message);
+    };
 
     eZ.helpers = eZ.helpers || {};
     eZ.helpers.notification = {
+        showNotification,
         showInfoNotification,
         showSuccessNotification,
         showWarningNotification,
-        showDangerNotification,
```
dew326: We cannot just remove this because of BC. We should add deprecation warning and remove this in 2.0 of admin-ui

dew326: We should always use the version of the package. 
`ezplatform-admin-ui` is now 1.x so next version for removing deprecations is 2.0 (which most likely will be released as eZPlatform 3.0).

src/bundle/Resources/public/js/scripts/helpers/notification.helper.js
```diff
@@ -60,22 +60,36 @@
      * @function showDangerNotification
      * @param {String} message
      */
-    const showDangerNotification = (message) =>
+    const showDangerNotification = (message) => {
+        console.warn('[DEPRECATED] showDangerNotification is deprecated');
+        console.warn('[DEPRECATED] it will be removed from ezplatform-admin-ui 2.0');
+        console.warn('[DEPRECATED] use showErrorNotification instead');
+
         showNotification({
```
sunpietro: it should be `showErrorNotification(message)` after changes

tischsoic: Fixed!

PR #595, opened 2018-08-13
EZP-29530: Simplify extending global eZ config

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29530
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | yes

## For QA
The app should work like before

## How to use it?

Previously, when trying to extend an existing global config developers had to write a piece of code similar to this one:

```javascript
window.eZ = window.eZ || {};
window.eZ.formBuilder = window.eZ.formBuilder || {};
window.eZ.formBuilder.config = JSON.parse('{{ config|json_encode|raw }}');
```

With a new feature they can write it in a simpler way:

```javascript
window.eZ.addConfig('formBuilder.config', JSON.parse('{{ config|json_encode|raw }}'));
```


src/bundle/Resources/views/layout.html.twig
```diff
@@ -4,31 +4,70 @@
...
+
+                            if (isOldValueArray && isNewValueArray) {
+                                return [...oldValue, ...newValue];
+                            }
+
+                            if (isOldValueArray && !isNewValueArray) {
+                                return [...oldValue, newValue];
+                            }
+
+                            if (oldValue.toString() === objectString && newValue.toString() === objectString) {
```
tischsoic: Why not `oldValue instanceof Object`?

src/bundle/Resources/views/layout.html.twig
```diff
@@ -4,31 +4,70 @@
...
+                    const setValue = (newValue, oldValue) => {
+                        if (shouldMerge) {
+                            const isOldValueArray = Array.isArray(oldValue);
+                            const isNewValueArray = Array.isArray(newValue);
+
+                            if (isOldValueArray && isNewValueArray) {
+                                return [...oldValue, ...newValue];
+                            }
+
+                            if (isOldValueArray && !isNewValueArray) {
```
dew326: I'm thinking about this one. So if oldValue is an array and the new value isn't it will be merged into one array.
Example:

oldvalue: `[1, 2, 3, 4]`
newValue: `{key: value}`

Will be merged into `[1, 2, 3, 4, {key: value}]`
I'm not sure if we want to do such things.

src/bundle/Resources/views/layout.html.twig
```diff
@@ -4,31 +4,70 @@
...
-                    notSamePasswords: '{{ 'js.error.not_same_passwords'|trans({}, 'validators')|desc('Passwords do not match') }}',
-                    invalidValue: '{{ 'js.error.invalid_value'|trans({}, 'validators')|desc('{fieldName} field has invalid values') }}',
-                    outOfRangeValue: '{{ 'js.error.out_of_range_value'|trans({}, 'validators')|desc('{fieldName} field value is out of range. The value has to be between {min} and {max}') }}'
-                },
+                addConfig: (path, value, shouldMerge = false) => {
+                    const keys = path.split('.');
+                    const lastIndex = keys.length - 1;
+                    const objectString = '[object Object]';
+                    const setValue = (newValue, oldValue) => {
+                        if (shouldMerge) {
```
tischsoic: For me it would be cleaner to do:
```js
if (!shouldMerge) {
    return newValue;
}

// merging code
```

src/bundle/Resources/views/layout.html.twig
```diff
@@ -4,31 +4,70 @@
...
-                    addressNotFound: '{{ 'js.error.address_not_found'|trans({}, 'validators')|desc('Provided address does not exist') }}',
-                    notSamePasswords: '{{ 'js.error.not_same_passwords'|trans({}, 'validators')|desc('Passwords do not match') }}',
-                    invalidValue: '{{ 'js.error.invalid_value'|trans({}, 'validators')|desc('{fieldName} field has invalid values') }}',
-                    outOfRangeValue: '{{ 'js.error.out_of_range_value'|trans({}, 'validators')|desc('{fieldName} field value is out of range. The value has to be between {min} and {max}') }}'
-                },
+                addConfig: (path, value, shouldMerge = false) => {
+                    const keys = path.split('.');
+                    const lastIndex = keys.length - 1;
+                    const objectString = '[object Object]';
+                    const setValue = (newValue, oldValue) => {
```
tischsoic: New function will be created each time, maybe it would be good to extract it above `window.eZ = {`?

sunpietro: Then you would create an unwanted global function. I wanted to keep it private..

src/bundle/Resources/views/layout.html.twig
sunpietro: I was testing a couple of approaches. I'll fix the order of returning values.

src/bundle/Resources/views/layout.html.twig
sunpietro: Because arrays are also objects.

src/bundle/Resources/views/layout.html.twig
tischsoic: You're right, we don't have self invoking function here. Is there any reason to not to have one here?

src/bundle/Resources/views/layout.html.twig
tischsoic: So we may check: `oldValue instanceof Object && !(oldValue instanceof Array)`

tischsoic: If Object has overriden `toString` this solution would not work.

sunpietro: It doesn't make sense. There are two things to consider: 
1. you double check,
2. there's also a function `Array.isArray()` that checks if something is an Array,
3. I'm checking for arrays before.

src/bundle/Resources/views/layout.html.twig
```diff
@@ -4,31 +4,65 @@
...
+                tooShort: '{{ 'js.error.too_short'|trans({}, 'validators')|desc('{fieldName} value should be greater than or equal to {minLength} characters') }}',
+                isNotInteger: '{{ 'js.error.is_not_integer'|trans({}, 'validators')|desc('{fieldName} value should be an integer') }}',
+                isNotFloat: '{{ 'js.error.is_not_float'|trans({}, 'validators')|desc('{fieldName} value should be a float number') }}',
+                isLess: '{{ 'js.error.is_less'|trans({}, 'validators')|desc('{fieldName} value should be greater than or equal to {minValue}') }}',
+                isGreater: '{{ 'js.error.is_greater'|trans({}, 'validators')|desc('{fieldName} value should be less than or equal to {maxValue}') }}',
+                invalidFileSize: '{{ 'js.error.invalid_file_size'|trans({}, 'validators')|desc('{fieldName}: Cannot upload. File exceeds file size limit.') }}',
+                provideLatitudeValue: '{{ 'js.error.provide_latitude_value'|trans({}, 'validators')|desc('Please, provide latitude value in the Latitude field') }}',
```
DominikaK: Get rid of comma after `please` (or, ideally, get rid of `please` altogether), but otherwise +1 for the string text.

PR #605, opened 2018-08-22
EZEE-2062: Unable to embed second item in Richtext field

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZEE-2062 <br /> https://jira.ez.no/browse/EZP-28994 <br /> https://jira.ez.no/browse/EZP-28835
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes/no
| Doc needed?   | yes/no

Partially fixes problem. Partially because when you have empty richtext and then:
1. add new embed
2. delete paragraph above

the Add Toolbar will not be displayed. This is because there is no selection region on Ebed field. Style Toolbar is displayed because it is not positioned based on selection.region directly. 

It fixes situation when nothing is selected, but editor has focus. 

#### Checklist:
- [x] Ready for Code Review


lserwatka:
@sunpietro can we proceed with QA here?

src/bundle/Resources/public/js/alloyeditor/src/toolbars/ez-add.js
```diff
@@ -7,34 +7,60 @@ export default class EzToolbarAdd extends AlloyEditor.Toolbars.add {
...
     /**
      * Lifecycle. Renders the UI of the button.
      *
      * @method render
-     * @return {Object} The content which should be rendered.
+     * @returns {Object} The content which should be rendered.
      */
     render() {
         const selectionData = this.props.selectionData;
+        const nothingSelected = this.checkIsNothingSelected();
```
sunpietro: Nitpick: this variable is not required. The function itself is pretty self-explanatory, so I would avoid declaring this variable.

lserwatka:
@tischsoic @dew326 what is the status here?

tischsoic:
@lserwatka, I think that @dew326 is right. I think we may close this PR and try other solution in other PR.

tischsoic:
Closing this PR, because one issue was solved in: https://github.com/ezsystems/ezplatform-admin-ui/pull/745 and for the 2nd we need different approach.

PR #629, opened 2018-09-18
EZP-29638: Fixed Uncaught TypeError exception when removing action fr…

sunpietro:
…om Form button

| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29638
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

## FOR QA
The `ezselection` fieldtype should be tested.

### DIFF
--- a/src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
+++ b/src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
@@ -44,17 +44,13 @@
         }
 
         [selectFirstItem]() {
-            const firstItem = this.itemsContainer.querySelector(`${SELECTOR_ITEM}:not([disabled])`);
+            const firstItem = this.itemsContainer.querySelector(`${SELECTOR_ITEM}`);
 
             firstItem.classList.add(CLASS_ITEM_SELECTED);
 
-            if (!firstItem.dataset.value) {
-                return;
-            }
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
```diff
@@ -124,14 +121,14 @@
         init() {
             const isEmpty = !this.container.querySelectorAll(SELECTOR_SELECTED_ITEM).length;
 
-            if (isEmpty && this[canSelectOnlyOne]) {
+            if (isEmpty && this[canSelectOnlyOne] && this.hasDefaultSelection) {
                 this[selectFirstItem]();
             }
 
-            this.container.querySelector(SELECTOR_SELECTION_INFO).addEventListener('click', this[onInputClick], false);
+            this.container.querySelector(SELECTOR_SELECTION_INFO).onclick = this[onInputClick];
```
tischsoic: Why not `addEventListener `?

sunpietro: Because it's easier to reinitialize, when you don't have to call `removeEventListener` and then `addEventListener`.

PR #630, opened 2018-09-18
WIP EZP-29639: Labels for checkboxes and radioboxes are misaligned with inputs

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29639
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | ?
| Tests pass?   | yes
| Doc needed?   | ?

Fixes rendering & CSS styles of all checkboxes + radioboxes. 
For QA: All checkboxes and radioboxes shoul look like nothing happened.

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_forms.scss
+++ b/src/bundle/Resources/public/scss/_forms.scss
@@ -61,18 +61,12 @@ form:not(.form-inline) {
 
 .form-check {
     margin-bottom: 0;
+}
 
-    &-label {
-        display: flex;
-        align-items: center;
-        margin-bottom: 0;
-    }
-
-    &-input {
-        position: initial;
-        margin-bottom: .25rem;
-        margin-right: .5rem;
-    }
+.form-check-input {
+    position: initial;
+    margin-bottom: .25rem;
+    margin-right: .5rem;
 }
 
 .ez-field {
--- a/src/bundle/Resources/views/form_fields.html.twig
+++ b/src/bundle/Resources/views/form_fields.html.twig
@@ -141,3 +141,43 @@
         {{ block('form_widget') }}
     </div>
 {%- endblock -%}
+
+{% block checkbox_radio_label -%}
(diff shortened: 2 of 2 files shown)


src/bundle/Resources/public/scss/_forms.scss
```diff
@@ -61,18 +61,12 @@ form:not(.form-inline) {
...
-        align-items: center;
-        margin-bottom: 0;
-    }
-
-    &-input {
-        position: initial;
-        margin-bottom: .25rem;
-        margin-right: .5rem;
-    }
+.form-check-input {
```
mikadamczyk: Should it be `.ez-form-check__input` ? @sunpietro

sunpietro: He's overriding Bootstrap styles, so no.

src/bundle/Resources/views/form_fields.html.twig
```diff
@@ -141,3 +141,43 @@
         {{ block('form_widget') }}
     </div>
 {%- endblock -%}
+
+{% block checkbox_radio_label -%}
```
webhdx: Is it copied from Symfony? What was changed here exactly? I'm against duplicating the code, maybe we can somehow extend Symfony's template or use different approach.

tischsoic: I added: `{%- if label is not same as(false) -%}` (code rendering label here) `{%- endif -%}`. 
Now, if label is set to false it is rendered with no text what is breaking tables. It was fixed by overwriting `.form-check-input` but this again is breaking checkboxes and radioboxes.

tischsoic: Hmm, maybe the other approach would be to use different template for checkboxes in tables, what do you think?

mikadamczyk:
@webhdx you are right, this should be fixed on css side

PR #634, opened 2018-09-19
EZP-29602: Large images do not fit in the editor window

konradoboza:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-29602](https://jira.ez.no/browse/EZP-29602)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Before:
before

After:
after

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/css/alloyeditor/alloyeditor-ez.css
+++ b/src/bundle/Resources/public/css/alloyeditor/alloyeditor-ez.css
@@ -989,6 +989,10 @@
   line-height: 18px;
 }
 
+.ez-embed-type-image img {
+  max-width: 100%;
+}
+
 [class*="ae-icon-"],
 [class*=" ae-icon-"] {
   font-size: 14px;


src/bundle/Resources/public/css/alloyeditor/alloyeditor-ez.css
```diff
@@ -989,6 +989,10 @@
   line-height: 18px;
 }
 
+.ez-embed-type-image img {
```
tischsoic: I wonder: `.ez-embed-type-image` is not enough, we need to apply this directly to `img` element?

konradoboza: Apparently, it didn't work when applied to parent divs.

sunpietro: What was the reason it didn't work?

I would expect a generic CSS rule for images, like `img { max-width: 100%; }`.
If something overrides it, then it has to be checked why and fixed.

konradoboza: There is no generic rule for img. A really similar approach is used for rendering images in content/view: https://github.com/ezsystems/ezplatform-admin-ui/blob/1.2/src/bundle/Resources/public/scss/_general.scss#L42

content_view

If you meant `img { max-width: 100%; }` set globally, it would work as it isn't overriden, but I am a bit unsure it it wouldn't ruin something else.

lserwatka:
@sunpietro I would proceed as it is, as it is more defensive fix, than setting it globally. Agree?

PR #653, opened 2018-10-02
EZP-29688: eZ Map Location fieldtype validation error message is not displaying correctly

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29688
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Implemented new behaviour on field type validation while trying to publish a content item. It differs from regular validation while typing incorrect data in fields. Thanks to this approach it will be easier to implement: https://jira.ez.no/browse/EZP-29679

zrzut ekranu 2018-10-02 o 10 20 21


sunpietro:
@micszo issues mentioned above are fixed

PR #663, opened 2018-10-03
EZP-29610: 'Create new URL alias' modal is rendered when user select URL tab

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29610
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

When the user can not add Custom URL modal is not rendered (it depends on `content/urltranslator` policy.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/content/tab/urls.html.twig
+++ b/src/bundle/Resources/views/content/tab/urls.html.twig
@@ -57,19 +57,34 @@
     }) }}
 {% endif %}
 
-{% include '@ezdesign/content/tab/url/modal_add_custom_url.html.twig' with {'form': form_custom_url_add, 'parent_name': parent_name} only %}
+{% if can_edit_custom_url %}
+    {% include '@ezdesign/content/tab/url/modal_add_custom_url.html.twig' with {
+        'form': form_custom_url_add,
+        'parent_name': parent_name
+    } only %}
+{% endif %}
 
 {% macro table_header_tools(form_custom_url_remove, can_edit_custom_url) %}
     {% if can_edit_custom_url %}
-        <button class="btn btn-primary ez-btn--prevented" data-toggle="modal" data-target="#ez-modal--custom-url-alias" title="{{ 'tab.urls.action.add'|trans|desc('Add Custom URL') }}">
+        <button
+            class="btn btn-primary ez-btn--prevented"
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/views/content/tab/urls.html.twig
```diff
@@ -57,19 +57,34 @@
...
+        'form': form_custom_url_add,
+        'parent_name': parent_name
+    } only %}
+{% endif %}
 
 {% macro table_header_tools(form_custom_url_remove, can_edit_custom_url) %}
     {% if can_edit_custom_url %}
-        <button class="btn btn-primary ez-btn--prevented" data-toggle="modal" data-target="#ez-modal--custom-url-alias" title="{{ 'tab.urls.action.add'|trans|desc('Add Custom URL') }}">
+        <button
+            class="btn btn-primary ez-btn--prevented"
```
sunpietro: `ez-btn--prevented` should come along with `ez-btn` class name as well.

src/bundle/Resources/views/content/tab/urls.html.twig
```diff
@@ -57,19 +57,34 @@
...
+            data-toggle="modal"
+            data-target="#ez-modal--custom-url-alias"
```
sunpietro: That's weird `id`. The BEM naming convention applies only to CSS class names, not to `id` attributes.

mikadamczyk: same here

src/bundle/Resources/views/content/tab/urls.html.twig
mikadamczyk: I have only reformated the code here for better readability and I think it is not the scope of this bug

lserwatka: I agree, let's not mess with it now as it might break things.

micszo:
On Platform v2.2.3 with this branch I get console error `Uncaught TypeError: Cannot read property 'querySelectorAll' of null` in Content structure. Is it something relevant?
screen shot 2018-10-04 at 11 36 00

PR #667, opened 2018-10-04
EZP-29663: Login page redirects to notifications/count

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29663
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
+++ b/src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
@@ -40,7 +40,8 @@
     const getNotificationsStatus = () => {
         const requestOptions = {
             headers: {
-                'X-CSRF-Token': token
+                'X-CSRF-Token': token,
+                'X-Requested-With' : 'XMLHttpRequest'
             },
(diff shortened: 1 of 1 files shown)


ViniTou:
Is there a correct way for us to apply this globally for every request?

ViniTou:
Hint for QA:

I opened two tabs, both on dashboard. Then logout on one and go to /admin first (not /admin/login). Then on second tab you should have two request every 30 seconds - count and login. Try to login then on the first tab, and you should be redirected to json from /notification/count.

andrerom:
So we rely on session for redirect or?
If so, that is to ask for problems :)

ViniTou:
@andrerom 
that's how symfony does it.

andrerom:
ok

ViniTou:
Yes, thought about it right now, definitely we could try this but in separate PR.

PR #670, opened 2018-10-04
Add twig filter to translate error message

JoseManuelGonzalezCantarero:
Using the bundle in a spanish project, where I need to show this message in spanish

### DIFF
--- a/src/bundle/Resources/views/Security/login.html.twig
+++ b/src/bundle/Resources/views/Security/login.html.twig
@@ -4,7 +4,7 @@
     <form action="{{ path( 'login_check' ) }}" method="post" role="form">
         <fieldset>
             {% if error %}
-                <span>{{ error.message }}</span>
+                <span>{{ error.message|trans }}</span>
             {% endif %}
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/views/Security/login.html.twig
```diff
@@ -4,7 +4,7 @@
     <form action="{{ path( 'login_check' ) }}" method="post" role="form">
         <fieldset>
             {% if error %}
-                <span>{{ error.message }}</span>
+                <span>{{ error.message|trans }}</span>
```
sunpietro: Thank you for contribution. Can you add a description filter as well? Something like `error.message|transdesc('Cannot login because of error.')` or similarly.

ViniTou: I think that would change whole behavior of this. That's like really bad solution.

crevillo: hi @ViniTou , you mean bad solution the error.message|trans part or the desc part? thanks!.

PR #673, opened 2018-10-05
EZP-29700: Syntax error after logging out

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | 
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | yes/no
| Tests pass?   | yes/no
| Doc needed?   | yes/no

screen shot 2018-10-05 at 4 19 56 pm

When a user has logged out a notification with information is show one time. When the user is logged back in and log out again information is shown again.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
+++ b/src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
@@ -1,5 +1,6 @@
 (function(global, doc, eZ, React, ReactDOM, Translator) {
     let currentPageLink = null;
+    let getNotificationsStatusErrorShowed = false;
     const SELECTOR_MODAL_ITEM = '.ez-notifications-modal__item';
     const SELECTOR_MODAL_RESULTS = '.ez-notifications-modal__results';
     const SELECTOR_MODAL_TITLE = '.modal-title';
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
```diff
@@ -66,7 +66,7 @@
                 setPendingNotificationCount(notificationsInfo);
                 updateModalTitleTotalInfo(notificationsInfo.total);
             })
-            .catch(console.error);
+            .catch();
```
dew326: Removing error handling is not a solution.
Here should be checked what was the response and do a proper thing to a response.

mikadamczyk:
@dew, @sunpietro can you take a look at it?

src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
```diff
@@ -11,6 +12,7 @@
     const INTERVAL = 30000;
     const modal = doc.querySelector('.ez-notifications-modal');
     const showErrorNotification = eZ.helpers.notification.showErrorNotification;
+    const showWarningNotification = eZ.helpers.notification.showWarningNotification;
```
dew326: You can combine two `const` and write it like:
```js
const { showErrorNotification, showWarningNotification } = eZ.helpers.notification;
```

PR #676, opened 2018-10-08
EZP-29328: As a editor I want OE to support formatted text

andrerom:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-29328](https://jira.ez.no/browse/EZP-29328)
| Bug fix?      | yes and no
| New feature?  | yes and no
| BC breaks?    | no
| Tests pass?   | yes _(there are none for js code)_
| Doc needed?   | yes

screenshot 2018-10-08 at 11 59 04

Besides being new feature in eZ Platform Admin UI, this allows migration of plain literal tags to eZ Platform as supported since 2.3.0 in kernel and migration scripts. This is why this is suggested for 1.3.1, to avoid customers reporting it as bug once they migrate.

#### Checklist:
- [x] Ready for Code Review


lserwatka:
Wow, 97 files changed for such a simple feature enablement.

andrerom:
@lserwatka No, there are only 3-4 files changed. Maybe I miss-understood something but IU tought I was supported to run webpack to generate /dist files. README or PR template todo's didn't say much about that.

andrerom:
Updated.

Removed webpack changes not relevant, needed to keep changes in all toolbars where option for formatted is shown, they change because of the change in `/base.js` which they extend.

src/bundle/Resources/public/js/alloyeditor/src/toolbars/config/ez-formatted.js
```diff
@@ -0,0 +1,35 @@
+import EzConfigBase from './base';
+
+export default class EzFormattedConfig extends EzConfigBase {
+    constructor(config) {
+        super(config);
+
+        this.name = 'paragraph';
```
dew326: Probably not a paragraph

andrerom: Indeed :)

src/bundle/Resources/public/js/alloyeditor/src/toolbars/config/ez-formatted.js
```diff
@@ -0,0 +1,36 @@
...
+export default class EzFormattedConfig extends EzConfigBase {
+    constructor(config) {
+        super(config);
+
+        this.name = 'formatted';
+        this.buttons = [
+            'ezmoveup',
+            'ezmovedown',
+            this.getStyles(config.customStyles),
+            // NOTE: alignment currently not supported on <pre> so skipping buttons for that
```
andrerom: @vidarl Actually did not check if this is supported or not, besides noticing there where no handling of this in the xsl. However if we did not have it in legacy either there is probably no need to add it on formatted.

andrerom:
@dew326 updated 👍 

_Note that I have *not* added this to the "Add block" toolbar, I think we should, but then we probably need a new icon for it and I'm unsure how you guys do that now._

lserwatka:
OK, let's move it to QA, we could make 2.3.1 on Friday as I have some other fixes in the pipeline.

src/bundle/Resources/public/js/alloyeditor/src/toolbars/config/ez-formatted.js
vidarl: We do not have align in legacy ( editor disables align buttons ) on literal.
Docbook does not support it on ProgramListing either, but we are free to add support for `ezxhtml:align` attribute if we choose to.

However, since it was not a need for it in legacy, it is likely no needed now either.

andrerom:
@katarzynazawada which of these issues are unique to the change here? As in, safari issue and chrome add button issue sounds like its unrelated => generic that might also happen with other blocks, or?

The Firefox and the chrome scroll bar issues are probably due to `<pre>` styling, I/someone might need to check it against Alloy's css and see if issue comes from there or if it comes from our own css.

andrerom:
@katarzynazawada If you have those browsers still running, could you cross check against https://alloyeditor.com/ ? _(front page, scroll down to "In live demo, click/touch here")_

katarzynazawada:
@andrerom I’ve checked these issues again both on master and your branch. From my observations, all of them only appear for Formatted block. What is more, on Chrome, there is one more case:  

1. Click on Rich Text Editor 
2. Change style to Formatted. 
3. Click outside the editor. 
4. Go back to Rich Text Editor.

  Result: 
screen shot 2018-10-12 at 12 29 48

andrerom:
Pushed CSS that tries to solve some of the issues, from what I can see these are fixed in Alloy 1.5!:
- https://github.com/liferay/alloy-editor/pull/867
- https://github.com/liferay/alloy-editor/pull/868
- https://github.com/liferay/alloy-editor/pull/888

@dew326 do you have suggestion on how we can force some content inside the tag if it's empty, the CSS I added for that does not work as intended atm.
_(Also would like to not having to set height to avoid it collapsing when not in focus, so suggestions welcome on that too.)_

lserwatka:
https://github.com/ezsystems/ezplatform-admin-ui-assets/blob/master/package.json

@andrerom We are using Alloy Editor v1.5.1

andrerom:
Great, was tricked by package.json here _(which is also using React 15 btw, so maybe we can clean that up)_.

@lserwatka Could we do a asset update with [Alloy 1.5.14](https://github.com/liferay/alloy-editor/blob/master/CHANGELOG.md)?

lserwatka:
@andrerom we are doing sanities at the moment. When result will be positive, we will update of course. We need 1 or 2 days.

andrerom:
> @andrerom we are doing sanities at the moment. When result will be positive, we will update of course. We need 1 or 2 days.

👍 _Note: RC tag could work as well so we can test this PR and do sanity without risking anything for current installs_

andrerom:
Pushed [ezplatform-admin-ui-assets 3.2.0-rc1](https://github.com/ezsystems/ezplatform-admin-ui-assets/releases/tag/v3.2.0-rc1) to test with newer Alloy version.

However the toolbar issue here is not caused by Alloy, as it's our own toolbar. Code responsibi,e for this set of bugs is `class EzConfigBase`

lserwatka:
@andrerom there was no need for this rc tag. We discovered set of regression while testing upgrade to latest JavaScript components, inc. Alloy Editor. @katarzynazawada and @dew326 are aware of the issues.

andrerom:
@lserwatka I (thought I ) needed it for testing, but it's dropped now.

andrerom:
@katarzynazawada Added fix for:
- toolbar position when scrolling* & when empty**
- height of formatted box when empty***
- made pre use scroll in all browsers _(added comment inline in css for how you can enable wrapped style instead, but non wrapped is the normal behavior for pre, so I stayed with that)_
- added `<>` styling when empty as placeholder.

You should not need to update Alloy for this, so you don't need new asset package to test this.

For devs:
\* _Needed to take .scrollLeft into account._
\*\*_ Caused by block not being considered empty by code._
\*\*\* _Added pre:empty::after where some placeholder content is set when empty._

katarzynazawada:
@andrerom After fixes I noticed some issues:

- **Safari**

Toolbar skips:

**A)**
1. Click on Rich Text Editor
2. Change style to Formatted.
3. Click outside Rich Text Editor but within a short distance.
4. Click on Rich Text Editor again. 

See movie: http://recordit.co/tsiwhvKKza

**B)**
1. Click on Rich Text Editor
2. Change style to Formatted.
3. Change style to Paragraph (or another one)

screenshot 2018-10-30 at 12 34 58

- **Chrome:**

Placeholder sometimes disappears 

1. Click on Rich Text Editor
2. Change style to Formatted.
3. Change style to Paragraph (or another one)
4. Change style to Formatted again. 

screenshot 2018-10-30 at 11 14 28

After publishing ,in Content item view, there is a blank space in field instead of _This field is empty_ text

screenshot 2018-10-30 at 11 14 38

- **Firefox:**

Pre use scroll is not added.



- _If clicking Add button in situation when cursor is outside the current view, page and toolbar skips._ - this issue still occurs for scroll.

andrerom:
@SylvainGuittard If you want this feature, please get someone else to take over at some point. The level of deep issues in Alloy / browsers here is beyond what I have time to reverse engineer any time soon.

andrerom:
UPDATE: Placeholder issue was fixed branch here in 45feb2fe323132751a92ba07cbabbc48283de4ff _(by removing placeholder, it's not important)_. However the toolbar issue is something else, maybe deeper in the editor, so if we want to cover all edge cases on first try here, it's better if a frontend guy can handle. /cc @lserwatka

PR #678, opened 2018-10-08
EZP-29709: RequestLocaleListener tries to match SiteAccess on /_fos_user_context_hash

kmadejski:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-29709](https://jira.ez.no/browse/EZP-29709)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

The `RequestLocaleListener` is called too early. In case of request to `/_fos_user_context_hash` it tries to resolve SiteAccess from the request, but since this route is not SiteAccess aware it ends-up with an Exception thrown from: https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/lib/Specification/SiteAccess/IsAdmin.php#L39

The `FOS\HttpCacheBundle\EventListener\UserContextSubscriber::onKernelRequest()` should be called before, therefore I've decreased the priority of the `RequestLocaleListener`.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/EventListener/RequestLocaleListener.php
+++ b/src/lib/EventListener/RequestLocaleListener.php
@@ -41,7 +41,7 @@ class RequestLocaleListener implements EventSubscriberInterface
     public static function getSubscribedEvents(): array
     {
         return [
-            KernelEvents::REQUEST => ['onKernelRequest', 13],
+            KernelEvents::REQUEST => ['onKernelRequest', 6],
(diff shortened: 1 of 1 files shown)


lserwatka:
@kmadejski please merge it up to master (1.3.x)

PR #652, opened 2018-10-01
EZP-29682: Cannot set default Location for objectrelationlist

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29682
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no
| License       | [GPL-2.0](https://github.com/ezsystems/ezplatform-admin-ui/blob/master/LICENSE)

#### Checklist:
- [x] Coding standards (`$ composer fix-cs`)
- [x] Ready for Code Review

PR #679, opened 2018-10-08
EZP-29660: Ezobjectrelationlist ignores allowed content types when creating content via UDW

konradoboza:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-29660](https://jira.ez.no/browse/EZP-29660)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Currentlym ezobjectrelationlist doesn't take CT's set as allowed via UI into account - only settings from `/src/bundle/Resources/config/universal_discovery_widget.yml` are visible.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/fieldtypes/edit/relation_base.html.twig
+++ b/src/bundle/Resources/views/fieldtypes/edit/relation_base.html.twig
@@ -4,6 +4,7 @@
     {% set limit = limit|default(0) %}
     {% set default_location = default_location|default(1) %}
     {% set has_relations = relations|length > 0 %}
+    {% set allowed_content_types = form.parent.vars.value.fieldDefinition.fieldSettings.selectionContentTypes %}
     <div class="ez-relations__wrapper" {% if not has_relations %}hidden="true"{% endif %}>
         <table class="table ez-relations__table">
             <thead>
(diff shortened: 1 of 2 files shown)


src/lib/UniversalDiscovery/Event/Subscriber/RelationListAllowedContentTypes.php
```diff
@@ -0,0 +1,51 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformAdminUi\UniversalDiscovery\Event\Subscriber;
+
+use EzSystems\EzPlatformAdminUi\UniversalDiscovery\Event\ConfigResolveEvent;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class RelationListAllowedContentTypes implements EventSubscriberInterface
```
adamwojs: I could be wrong but `ObjectRelationStartingLocationId` seems to have similar role but for the `starting_location_id` parameter. Should we keep convention and name this listener `ObjectRelationAllowedContentTypes` and also limit the scope of it  same as in the following example: https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/lib/UniversalDiscovery/Event/Subscriber/ObjectRelationStartingLocationId.php#L37-L43 ?

konradoboza: @adamwojs when it comes to naming, I could probably agree, yet I didn't notice any convention, e.g.: https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/lib/UniversalDiscovery/Event/Subscriber/RichTextAllowedLanguages.php. When it comes to the condition: wouldn't it be less readable? Introduced Subscriber applies only to allowed CT's for ezobjectrelationlist so checking them should be enough to make it work properly. Or?

konradoboza:
All the remarks from @adamwojs discussed and applied. Please review @kmadejski @Nattfarinn @ViniTou.

micszo:
@konradoboza the patch failes on 2.2.3 (1.2.3), please advise.
screen shot 2018-10-11 at 09 09 19

konradoboza:
From my understanding, PR https://github.com/ezsystems/ezplatform-admin-ui/pull/652 was released only with eZ Platform v2.3.0. My changes extend some of the changes which were introduced there, so either apply patch form mentioned PR and then from this on, or use `1.2` branch instead of tag in `composer.json`.

micszo:
Ok, fine by me. What about customer version? :) In ticket 2.2.2 is marked.

konradoboza:
I guess I will provide them with both patches. :)

lserwatka:
Please merge this up to master

konradoboza:
@lserwatka done.

PR #683, opened 2018-10-12
EZP-29712 Run nightly Behat tests on demo metarepos

m-tyrala:
| Question      | Answer
| ------------- | ---
| Tickets       | 
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

Adds pagination handling for administration pages. Changes in notifications were obligatory because new notification covers over the pagination section with `Next` button, which causes errors.

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/lib/Behat/BusinessContext/AdministrationContext.php
+++ b/src/lib/Behat/BusinessContext/AdministrationContext.php
@@ -84,11 +84,9 @@ class AdministrationContext extends BusinessContext
      */
     public function isElementOnTheList(string $listElementName, string $page, ?string $parameter = null): void
     {
-        $isElementOnTheList = PageObjectFactory::createPage($this->utilityContext, $page, $parameter)
-            ->adminList->table->isElementInTable($listElementName);
-
-        if (!$isElementOnTheList) {
-            Assert::fail(sprintf('Element "%s" is not on the %s list.', $listElementName, $page));
+        $pageElement = PageObjectFactory::createPage($this->utilityContext, $page, $parameter);
+        if (!$this->findElementOnTheList($pageElement, $listElementName)) {
+            Assert::fail(sprintf('Element "%s" is on the %s list.', $listElementName, $page));
         }
     }
 
@@ -98,10 +96,8 @@ class AdministrationContext extends BusinessContext
      */
     public function isElementNotOnTheList(string $listElementName, string $page, string $parameter = null): void
     {
-        $isElementOnTheList = PageObjectFactory::createPage($this->utilityContext, $page, $parameter)
-            ->adminList->table->isElementInTable($listElementName);
-
-        if ($isElementOnTheList) {
+        $pageElement = PageObjectFactory::createPage($this->utilityContext, $page, $parameter);
+        if ($this->findElementOnTheList($pageElement, $listElementName)) {
             Assert::fail(sprintf('Element "%s" is on the %s list.', $listElementName, $page));
         }
     }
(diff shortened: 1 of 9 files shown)


m-tyrala:
[Tests passed](https://travis-ci.org/ezsystems/ezplatform-admin-ui/jobs/441597218) - Normally problem occurs only on Demo envs (more content types) so to test it I simply ran ContentType tests after ContentTypeFields tests.

src/lib/Behat/BusinessContext/AdministrationContext.php
```diff
@@ -278,4 +278,24 @@ public function listIsEmpty(string $listName, string $itemType, string $itemName
         $pageObject = PageObjectFactory::createPage($this->utilityContext, $itemType, $itemName);
         $pageObject->verifyListIsEmpty($listName);
     }
+
+    public function findElementOnTheList(Page $pageElement, string $listElementName): bool
```
mnocon: Either this function should return Element or be named `isElementOnTheList`

src/lib/Behat/BusinessContext/AdministrationContext.php
```diff
@@ -278,4 +278,24 @@ public function listIsEmpty(string $listName, string $itemType, string $itemName
...
+    {
+        $isElementOnTheList = false;
+        while (true) {
+            $isElementOnTheListPage = $pageElement->adminList->table->isElementInTable($listElementName);
+            if ($isElementOnTheListPage) {
+                $isElementOnTheList = $isElementOnTheListPage;
```
mnocon: `return true` simplifies it a lot.

src/lib/Behat/BusinessContext/AdministrationContext.php
```diff
@@ -278,4 +278,24 @@ public function listIsEmpty(string $listName, string $itemType, string $itemName
...
+                break;
+            }
+
+            if ($pageElement->adminList->isPaginationNextButtonActive()) {
+                $pageElement->adminList->clickPaginationNextButton();
+            } else {
+                break;
```
mnocon: `return false` also simplifies IMHO.

src/lib/Behat/BusinessContext/AdministrationContext.php
```diff
@@ -98,10 +96,8 @@ public function isElementOnTheList(string $listElementName, string $page, ?strin
      */
     public function isElementNotOnTheList(string $listElementName, string $page, string $parameter = null): void
     {
-        $isElementOnTheList = PageObjectFactory::createPage($this->utilityContext, $page, $parameter)
-            ->adminList->table->isElementInTable($listElementName);
-
-        if ($isElementOnTheList) {
+        $pageElement = PageObjectFactory::createPage($this->utilityContext, $page, $parameter);
+        if ($this->findElementOnTheList($pageElement, $listElementName)) {
```
mnocon: Why is this function in Context and not in AdminList/Table?

src/lib/Behat/BusinessContext/AdministrationContext.php
m-tyrala: I called it `findElementOnTheList` because it searches for the element through the pages and `isElementOnTheList` is already taken by one of the step functions. Maybe I will just change it to `is` and step functions to `verify` as we call it in other places.

src/lib/Behat/BusinessContext/AdministrationContext.php
m-tyrala: Yeah, you are right

m-tyrala:
@lserwatka ready to merge

PR #695, opened 2018-10-18
EZP-29223: Implement permissions for "Content/Translate"

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29223
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no
| License       | [GPL-2.0](https://github.com/ezsystems/ezplatform-admin-ui/blob/master/LICENSE)

depends on [https://github.com/ezsystems/ezpublish-kernel/pull/2465](https://github.com/ezsystems/ezpublish-kernel/pull/2465)

#### Checklist:
- [x] Coding standards (`$ composer fix-cs`)
- [x] Ready for Code Review

PR #696, opened 2018-10-19
EZP-29734: As an Administrator I want to add Content Type limitation for Section/Assign policy

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29734
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/SectionController.php
+++ b/src/bundle/Controller/SectionController.php
@@ -15,9 +15,11 @@ use eZ\Publish\API\Repository\Values\Content\ContentInfo;
 use eZ\Publish\API\Repository\Values\Content\Query;
 use eZ\Publish\API\Repository\Values\Content\Query\SortClause;
 use eZ\Publish\API\Repository\Values\Content\Section;
+use eZ\Publish\API\Repository\Values\User\Limitation\NewSectionLimitation;
 use eZ\Publish\Core\MVC\Symfony\Security\Authorization\Attribute;
 use eZ\Publish\Core\Pagination\Pagerfanta\ContentSearchAdapter;
 use eZ\Publish\API\Repository\SearchService;
+use eZ\Publish\API\Repository\PermissionResolver;
 use EzSystems\EzPlatformAdminUi\Form\Data\Section\SectionContentAssignData;
 use EzSystems\EzPlatformAdminUi\Form\Data\Section\SectionCreateData;
 use EzSystems\EzPlatformAdminUi\Form\Data\Section\SectionDeleteData;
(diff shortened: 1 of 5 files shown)


src/lib/UniversalDiscovery/Event/Subscriber/SectionAssign.php
```diff
@@ -0,0 +1,110 @@
...
+        ];
+    }
+
+    /**
+     * @param \EzSystems\EzPlatformAdminUi\UniversalDiscovery\Event\ConfigResolveEvent $event
+     */
+    public function onUdwConfigResolve(ConfigResolveEvent $event): void
+    {
+        $configName = $event->getConfigName();
+        if ('single' !== $configName && 'multiple' !== $configName) {
```
dew326: I'm not sure if we need to check for `single` - I think we have only `multiple`.

src/lib/Util/PermissionUtil.php
```diff
@@ -0,0 +1,38 @@
...
+namespace EzSystems\EzPlatformAdminUi\Util;
+
+class PermissionUtil
+{
+    /**
+     * @param array $hasAccess
+     *
+     * @return array
+     */
+    public function flattenArrayOfLimitations(array $hasAccess): array
```
andrerom: As there is no usage of $this here, it could technically be a static util class.

src/bundle/Controller/SectionController.php
```diff
@@ -490,9 +507,25 @@ private function getSectionsNumbers(array $sections): array
...
+        $hasAccess = $this->permissionResolver->hasAccess('section', 'assign');
+
+        if (is_bool($hasAccess)) {
+            return $hasAccess;
+        }
+
+        foreach ($this->permissionUtil->flattenArrayOfLimitations($hasAccess) as $limitation) {
+            if ($limitation instanceof NewSectionLimitation) {
+                // If one of user limitation is NewSectionLimitation we check if user can assign Content to $section
+                return in_array($section->id, $limitation->limitationValues);
```
andrerom: As this is iterating several roles and policies, this should not return here unless value is true.
Afaik you can do like in `checkAllowedContentTypes()` and collect all allowed sections before you do a in_array.

src/bundle/Controller/SectionController.php
```diff
@@ -490,9 +507,25 @@ private function getSectionsNumbers(array $sections): array
      * @param \eZ\Publish\API\Repository\Values\Content\Section $section
      *
      * @return bool
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
      */
     private function canUserAssignSectionToAnyContent(Section $section): bool
```
andrerom: naming, but something like `canUserAssignSectionToSomeContent` given a true here is more like a "maybe", while a false is pretty accurate.

mikadamczyk:
@adamwojs can you take a look at it?

barbaragr:
1. Add policies to the Editor role:
* Setup/Administrate
* Section/View
* User/Login
* Content/*
* Section/Assign
2. Log in as an Editor
3. Go to Content/Structure and create an Article
4. Go to Admin/Sections, click assign button  next to Design section
5. Select created article, click (+) button and click Confirm

*Actual result:* Error 403 is show: Access denied

barbaragr:
For Admin user after publishing an Article error 500 is thrown:
1. Go to Content/Content structure
2. Click Create and select Article
3. Fill Title and Intro, click Publish

*Actual result:* Error 500 is thrown: None of the routers in the chain matched url '/admin/content/location/61' but article was published.

lserwatka:
@mikadamczyk could you rebase here, as I merged https://github.com/ezsystems/ezplatform-admin-ui/pull/695

mikadamczyk:
@lserwatka done

PR #697, opened 2018-10-19
EZP-29355: As a developer, I want to edit the ezcontent of a custom tag

dew326:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29355
| Bug fix?      | no
| New feature?  | yes/
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review


src/bundle/Resources/public/js/scripts/fieldType/ezrichtext.js
```diff
@@ -211,16 +207,16 @@
...
 
-        container.addEventListener('blur', event => {
+        container.addEventListener('blur', (event) => {
             const data = alloyEditor.get('nativeEditor').getData();
             const doc = document.createDocumentFragment();
             const root = document.createElement('div');
             const ezNamespace = 'http://ez.no/namespaces/ezpublish5/xhtml5/edit';
             const xhtmlNamespace = 'http://www.w3.org/1999/xhtml';
-            const emptyEmbed = function (embedNode) {
+            const emptyEmbed = function(embedNode) {
```
sunpietro: Shouldn't it be extracted outside of callback scope, should it?

src/bundle/Resources/public/js/scripts/fieldType/ezrichtext.js
```diff
@@ -230,21 +226,28 @@
...
+                        removeClass = function() {
                             embedNode.classList.remove(cl);
                             prevRemoveClass();
                         };
                     }
                 });
                 removeClass();
             };
-            const xhtmlify = function (data) {
+            const clearCustomTag = (customTag) => {
```
sunpietro: Shouldn't it be extracted outside of callback scope, should it?

src/bundle/Resources/public/js/scripts/fieldType/ezrichtext.js
```diff
@@ -230,21 +226,28 @@
...
+                const attributesNodes = [...customTag.querySelectorAll('[data-ezelement="ezattributes"]')];
+                const headers = [...customTag.querySelectorAll('.ez-custom-tag__header')];
+
+                attributesNodes.forEach((attributesNode) => attributesNode.remove());
+                headers.forEach((header) => header.remove());
+            };
+            const xhtmlify = function(data) {
```
sunpietro: Shouldn't it be extracted outside of callback scope, should it?

src/bundle/Resources/public/scss/_custom-tag.scss
```diff
@@ -0,0 +1,72 @@
...
+
+    &--attributes-visible {
+        [data-ezelement='ezattributes'] {
+            display: block;
+        }
+
+        .ez-custom-tag__header-btn--attributes {
+            background-color: $ez-color-secondary;
+
+            .ez-icon {
```
sunpietro: Maybe you should apply `ez-icon--light` CSS class instead?

src/bundle/Resources/public/scss/_custom-tag.scss
```diff
@@ -0,0 +1,72 @@
...
+
+    &--content-visible {
+        [data-ezelement='ezcontent'] {
+            display: block;
+        }
+
+        .ez-custom-tag__header-btn--content {
+            background-color: $ez-color-secondary;
+
+            .ez-icon {
```
sunpietro: Maybe you should apply `ez-icon--light` CSS class instead?

src/bundle/Resources/public/scss/_custom-tag.scss
dew326: No, color depends on the parent class (state).

src/bundle/Resources/public/scss/_custom-tag.scss
dew326: Same case.

tischsoic:
I think we should reformat all *.js files according to our Prettify rules. Reading further PRs without formatting changes would be a lot easier. :) ping @sunpietro @dew326

bdunogier:
What is the status of this PR ? I can't really tell based on the latest contributions to it.

dew326:
@bdunogier This waits for QA, but we cannot test it without "nested custom tags".

bdunogier:
Why does testing require nested custom tags, @dew326 ? If you test editing of regular text, without tags in the `ezcontent`, isn't it enough ?

dew326:
@micszo tried to test it but we couldn't determinate which problems are because of the ezcontent and which because of nested custom tags. For example, embed and embed image doesn't work because problems in nesting `ezconfig` (probably will be fixed in nested custom tags). 

WIthout nested custom tags it's not possible to make full and proper QA, and probably the QA would need to be duplicated while testing nested custom tags.

bdunogier:
I understand. 

On the other hand, we have customers waiting for this, in order to edit content that doesn't include nested tags. I really see benefits in having this feature without the nesting part. The scope would then be very limited, but it would still bring something valuable.

Would it be complicated to disable all the OE buttons when the ezcontent editing box has focus ?

dew326:
But the nested custom tags are planned for 2.4, the same as this `ezcontent`. 

I don't see a point in spending time on limiting this and later enable all buttons as it's all planned for 2.4 and will not be released earlier.

bdunogier:
I knew you would say that, and it of course makes sense. As long as we don't end up skipping it in 2.4 for some reason, it will have to do. Thank you.

dew326:
I think we are safe in this case, @alongosz already has the nested custom tags in the current sprint.

bdunogier:
Okay, that sounds good. Thank you for the infos, @dew326.

PR #711, opened 2018-11-08
EZP-29766: As an editor, I want to filter the search using subtree parameter

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29766
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

## Description
This PR only adds filtering by subtree and simple displaying selected path.

## Restyling
This PR also includes restyling of Search filters.

screen shot 2018-11-20 at 11 15 51

###  Subtree button styling:

screen shot 2018-11-20 at 11 23 57

screen shot 2018-11-28 at 14 15 27

screen shot 2018-11-20 at 11 23 50

#### in case of longer breadcrumb:

screen shot 2018-11-20 at 14 57 45

## Basic responsiveness to window width:

View should adapt to window width; 
e.g number of columns in "Content type" filters now changes accordingly to window width:

screen shot 2018-11-20 at 11 18 41

Arrow always point to the center of the filters button: 

screen shot 2018-11-20 at 11 17 06

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/SearchController.php
+++ b/src/bundle/Controller/SearchController.php
@@ -108,6 +108,7 @@ class SearchController extends Controller
         $contentTypes = [];
         $lastModified = $search['last_modified'] ?? [];
         $created = $search['created'] ?? [];
+        $subtree = $search['subtree'] ?? null;
 
         if (!empty($search['section'])) {
             $section = $this->sectionService->loadSection($search['section']);
@@ -119,7 +120,7 @@ class SearchController extends Controller
         }
 
         $form = $this->formFactory->createSearchForm(
-            new SearchData($limit, $page, $query, $section, $contentTypes, $lastModified, $created, $creator),
+            new SearchData($limit, $page, $query, $section, $contentTypes, $lastModified, $created, $creator, $subtree),
             'search',
             [
                 'method' => Request::METHOD_GET,
@@ -139,6 +140,7 @@ class SearchController extends Controller
                 $lastModified = $data->getLastModified();
                 $created = $data->getCreated();
                 $creator = $data->getCreator();
+                $subtree = $data->getSubtree();
                 $query = new Query();
                 $criteria = [];
 
@@ -171,6 +173,9 @@ class SearchController extends Controller
                         Criterion\Operator::EQ,
                         $creator->id
                     );
+                }
+                if (null !== $subtree) {
+                    $criteria[] = new Criterion\Subtree($subtree);
                 }
                 if (!empty($criteria)) {
                     $query->filter = new Criterion\LogicalAnd($criteria);
--- a/src/bundle/Resources/public/js/scripts/admin.search.filters.js
+++ b/src/bundle/Resources/public/js/scripts/admin.search.filters.js
@@ -20,7 +20,6 @@
     const listGroupsTitle = [...doc.querySelectorAll('.ez-content-type-selector__group-title')];
     const contentTypeCheckboxes = [...doc.querySelectorAll('.ez-content-type-selector__item [type="checkbox"]')];
     const subtreeInput = doc.querySelector('#search_subtree');
-
     const clearFilters = (event) => {
         event.preventDefault();
 
--- a/src/bundle/Resources/public/js/scripts/admin.search.filters.js
+++ b/src/bundle/Resources/public/js/scripts/admin.search.filters.js
@@ -63,7 +63,7 @@
         const isModifiedSelected = !!lastModifiedSelect.value;
         const isCreatedSelected = !!lastCreatedSelect.value;
         const isCreatorSelected = !!searchCreatorInput.value;
-        const isSubtreeSelected = !!subtreeInput.value;
+        const isSubtreeSelected = !!subtreeInput.value.trim().length;
         const isEnabled = isContentTypeSelected || isSectionSelected || isModifiedSelected || isCreatedSelected || isCreatorSelected || isSubtreeSelected;
         const methodName = isEnabled ? 'removeAttribute' : 'setAttribute';
 
--- a/src/bundle/Resources/public/js/scripts/admin.search.filters.js
+++ b/src/bundle/Resources/public/js/scripts/admin.search.filters.js
@@ -19,6 +19,8 @@
     const resetCreatorBtn = doc.querySelector('.ez-filters__item--creator .ez-icon--reset');
     const listGroupsTitle = [...doc.querySelectorAll('.ez-content-type-selector__group-title')];
     const contentTypeCheckboxes = [...doc.querySelectorAll('.ez-content-type-selector__item [type="checkbox"]')];
+    const subtreeInput = doc.querySelector('#search_subtree');
+
     const clearFilters = (event) => {
         event.preventDefault();
 
@@ -48,6 +50,7 @@
         lastModifiedEnd.value = '';
         lastCreatedPeriod.value = '';
         lastCreatedEnd.value = '';
+        subtreeInput.value = '';
 
         handleResetUser();
 
@@ -60,7 +63,8 @@
         const isModifiedSelected = !!lastModifiedSelect.value;
         const isCreatedSelected = !!lastCreatedSelect.value;
         const isCreatorSelected = !!searchCreatorInput.value;
-        const isEnabled = isContentTypeSelected || isSectionSelected || isModifiedSelected || isCreatedSelected || isCreatorSelected;
+        const isSubtreeSelected = !!subtreeInput.value;
+        const isEnabled = isContentTypeSelected || isSectionSelected || isModifiedSelected || isCreatedSelected || isCreatorSelected || isSubtreeSelected;
         const methodName = isEnabled ? 'removeAttribute' : 'setAttribute';
 
         applyBtn[methodName]('disabled', !isEnabled);
@@ -276,6 +280,7 @@
         sectionSelect.addEventListener('change', toggleDisabledStateOnApplyBtn, false);
     }
 
+    subtreeInput.addEventListener('change', toggleDisabledStateOnApplyBtn, false);
     lastModifiedSelect.addEventListener('change', toggleModalVisibility, false);
     lastCreatedSelect.addEventListener('change', toggleModalVisibility, false);
     creatorInput.addEventListener('keyup', handleTyping, false);
--- a/src/bundle/Resources/public/js/scripts/udw/select.location.js
+++ b/src/bundle/Resources/public/js/scripts/udw/select.location.js
@@ -1,0 +1,94 @@
+(function (global, doc, React, ReactDOM, Translator) {
+    const btns = doc.querySelectorAll('.ez-btn--udw-select-location');
+    const udwContainer = doc.getElementById('react-udw');
+    const token = doc.querySelector('meta[name="CSRF-Token"]').content;
+    const siteaccess = doc.querySelector('meta[name="SiteAccess"]').content;
+
+    const findLocationsByIdList = (idList, callback) => {
+        const body = JSON.stringify({
+            ViewInput: {
+                identifier: `udw-locations-by-path-string-${idList.join('-')}`,
+                public: false,
+                LocationQuery: {
+                    Criteria: {},
+                    FacetBuilders: {},
+                    SortClauses: { SectionIdentifier: 'ascending' },
+                    Filter: { LocationIdCriterion: idList.join(',') },
+                    limit: 50,
+                    offset: 0,
+                },
+            },
+        });
+
+        const request = new Request('/api/ezp/v2/views', {
+            method: 'POST',
+            headers: {
+                Accept: 'application/vnd.ez.api.View+json; version=1.1',
+                'Content-Type': 'application/vnd.ez.api.ViewInput+json; version=1.1',
+                'X-Requested-With' : 'XMLHttpRequest',
+                'X-Siteaccess': siteaccess,
+                'X-CSRF-Token': token
+            },
+            body,
+            mode: 'same-origin',
+            credentials: 'same-origin',
+        });
+
+        const errorMessage = Translator.trans(
+            /*@Desc("Cannot find children locations with given id list")*/ 'select_location.error',
+            {},
+            'universal_discovery_widget'
+        );
+        fetch(request)
+            .then(window.eZ.helpers.request.getJsonFromResponse)
+            .then((json) => callback(json))
+            .catch(() => window.eZ.helpers.notification.showErrorNotification(errorMessage));
+    };
+
+    const removeRootFromPathString = (pathString) => {
+        const pathArray = pathString.split('/').filter((val) => val);
+
+        return pathArray.splice(1, pathArray.length - 1);
+    };
+    const closeUDW = () => ReactDOM.unmountComponentAtNode(udwContainer);
+    const onConfirm = (btn, items) => {
+        closeUDW();
+
+        const pathString = items[0].pathString;
+        const displayWrapper = doc.querySelector(btn.dataset.displayWrapper);
+        const pathStringInput = doc.querySelector(btn.dataset.locationPathInputSelector);
+
+        pathStringInput.value = pathString;
+        pathStringInput.dispatchEvent(new Event('change'));
+
+        if (!displayWrapper) {
+            return;
+        }
+
+        findLocationsByIdList(removeRootFromPathString(pathString), (data) => {
+            const searchHitList = data.View.Result.searchHits.searchHit;
+            const pathStringNames = searchHitList.map((searchHit) => {
+                return searchHit.value.Location.ContentInfo.Content.Name;
+            }).join(' / ');
+
+            displayWrapper.querySelector(btn.dataset.selectedContentNameSelector).innerHTML = pathStringNames;
+        });
+    };
+    const onCancel = () => closeUDW();
+    const openUDW = (event) => {
+        event.preventDefault();
+
+        const config = JSON.parse(event.currentTarget.dataset.udwConfig);
+
+        ReactDOM.render(React.createElement(global.eZ.modules.UniversalDiscovery, Object.assign({
+            onConfirm: onConfirm.bind(null, event.currentTarget),
+            onCancel,
+            title: event.currentTarget.dataset.universalDiscoveryTitle,
+            multiple: false,
+            startingLocationId: window.eZ.adminUiConfig.universalDiscoveryWidget.startingLocationId,
+            restInfo: { token, siteaccess }
+        }, config)), udwContainer);
+    };
+
+    btns.forEach(btn => btn.addEventListener('click', openUDW, false));
+})(window, window.document, window.React, window.ReactDOM, window.Translator);
--- a/src/bundle/Resources/public/js/scripts/udw/select.location.js
+++ b/src/bundle/Resources/public/js/scripts/udw/select.location.js
@@ -64,7 +64,7 @@
     const setLocationPath = (btn, pathString) => {
         const pathStringInput = doc.querySelector(btn.dataset.locationPathInputSelector);
         const contentBreadcrumbsContainer = doc.querySelector(
-            `${btn.dataset.contentBreadcrumbsSelector} .ez-filters__subtree__breadcrumbs`
+            `${btn.dataset.contentBreadcrumbsSelector} .ez-filters__breadcrumbs`
         );
 
         pathStringInput.value = pathString;
--- a/src/bundle/Resources/public/js/scripts/udw/select.location.js
+++ b/src/bundle/Resources/public/js/scripts/udw/select.location.js
@@ -1,4 +1,4 @@
-(function(global, doc, eZ, React, ReactDOM, Translator){
+(function(global, doc, eZ, React, ReactDOM, Translator) {
     const btns = doc.querySelectorAll('.ez-btn--udw-select-location');
     const udwContainer = doc.getElementById('react-udw');
     const token = doc.querySelector('meta[name="CSRF-Token"]').content;
@@ -38,9 +38,9 @@
         );
 
         fetch(request)
-            .then(window.eZ.helpers.request.getJsonFromResponse)
-            .then((json) => callback(json))
-            .catch(() => window.eZ.helpers.notification.showErrorNotification(errorMessage));
+            .then(eZ.helpers.request.getJsonFromResponse)
+            .then(callback)
+            .catch(() => eZ.helpers.notification.showErrorNotification(errorMessage));
     };
     const removeRootFromPathString = (pathString) => {
         const pathArray = pathString.split('/').filter((val) => val);
@@ -52,22 +52,23 @@
 
         return searchHitList.map((searchHit) => searchHit.value.Location.ContentInfo.Content.Name).join(' / ');
     };
-    const toggleVisibility = (btn, pathString) => {
-        btn.hidden = !!pathString;
-
+    const toggleVisibility = (btn, isLocationSelected) => {
         const contentBreadcrumbsWrapper = doc.querySelector(btn.dataset.contentBreadcrumbsSelector);
 
+        btn.hidden = isLocationSelected;
+
         if (contentBreadcrumbsWrapper) {
-            contentBreadcrumbsWrapper.hidden = !pathString;
+            contentBreadcrumbsWrapper.hidden = !isLocationSelected;
         }
     };
     const setPath = (btn, pathString) => {
         const pathStringInput = doc.querySelector(btn.dataset.locationPathInputSelector);
+        const contentBreadcrumbsContainer = doc.querySelector(
+            `${btn.dataset.contentBreadcrumbsSelector} .ez-filters__subtree__breadcrumbs`
+        );
 
         pathStringInput.value = pathString;
         pathStringInput.dispatchEvent(new Event('change'));
-
-        const contentBreadcrumbsContainer = doc.querySelector(`${btn.dataset.contentBreadcrumbsSelector} ez-filters__subtree__breadcrumbs`);
 
         if (!contentBreadcrumbsContainer) {
             return;
@@ -88,7 +89,7 @@
         const pathString = items[0].pathString;
 
         setPath(btn, pathString);
-        toggleVisibility(btn, pathString);
+        toggleVisibility(btn, !!pathString);
     };
     const onCancel = () => closeUDW();
     const openUDW = (event) => {
@@ -98,7 +99,7 @@
 
         ReactDOM.render(
             React.createElement(
-                global.eZ.modules.UniversalDiscovery,
+                eZ.modules.UniversalDiscovery,
                 Object.assign(
                     {
                         onConfirm: onConfirm.bind(null, event.currentTarget),
@@ -116,7 +117,7 @@
     };
     const clearValue = (btn) => {
         setPath(btn, '');
-        toggleVisibility(btn, '');
+        toggleVisibility(btn, false);
     };
 
     btns.forEach((btn) => {
--- a/src/bundle/Resources/public/js/scripts/udw/select.location.js
+++ b/src/bundle/Resources/public/js/scripts/udw/select.location.js
@@ -61,7 +61,7 @@
             contentBreadcrumbsWrapper.hidden = !isLocationSelected;
         }
     };
-    const setPath = (btn, pathString) => {
+    const setLocationPath = (btn, pathString) => {
         const pathStringInput = doc.querySelector(btn.dataset.locationPathInputSelector);
         const contentBreadcrumbsContainer = doc.querySelector(
             `${btn.dataset.contentBreadcrumbsSelector} .ez-filters__subtree__breadcrumbs`
@@ -88,7 +88,7 @@
 
         const pathString = items[0].pathString;
 
-        setPath(btn, pathString);
+        setLocationPath(btn, pathString);
         toggleVisibility(btn, !!pathString);
     };
     const onCancel = () => closeUDW();
@@ -115,8 +115,8 @@
             udwContainer
         );
     };
-    const clearValue = (btn) => {
-        setPath(btn, '');
+    const clearSelection = (btn) => {
+        setLocationPath(btn, '');
         toggleVisibility(btn, false);
     };
 
@@ -126,7 +126,7 @@
         const clearBtn = doc.querySelector(btn.dataset.clearBtnSelector);
 
         if (clearBtn) {
-            clearBtn.addEventListener('click', clearValue.bind(null, btn), false);
+            clearBtn.addEventListener('click', clearSelection.bind(null, btn), false);
         }
     });
 })(window, document, window.eZ, window.React, window.ReactDOM, window.Translator);
--- a/src/bundle/Resources/public/js/scripts/udw/select.location.js
+++ b/src/bundle/Resources/public/js/scripts/udw/select.location.js
@@ -55,10 +55,10 @@
     const toggleVisibility = (btn, pathString) => {
         btn.hidden = !!pathString;
 
-        const contentNameWrapper = doc.querySelector(btn.dataset.contentBreadcrumbsSelector);
+        const contentBreadcrumbsWrapper = doc.querySelector(btn.dataset.contentBreadcrumbsSelector);
 
-        if (contentNameWrapper) {
-            contentNameWrapper.hidden = !pathString;
+        if (contentBreadcrumbsWrapper) {
+            contentBreadcrumbsWrapper.hidden = !pathString;
         }
     };
     const setPath = (btn, pathString) => {
@@ -67,17 +67,17 @@
         pathStringInput.value = pathString;
         pathStringInput.dispatchEvent(new Event('change'));
 
-        const contentNameContainer = doc.querySelector(`${btn.dataset.contentBreadcrumbsSelector} div`);
+        const contentBreadcrumbsContainer = doc.querySelector(`${btn.dataset.contentBreadcrumbsSelector} ez-filters__subtree__breadcrumbs`);
 
-        if (!contentNameContainer) {
+        if (!contentBreadcrumbsContainer) {
             return;
         }
 
         if (!pathString) {
-            contentNameContainer.innerHTML = '';
+            contentBreadcrumbsContainer.innerHTML = '';
         } else {
             findLocationsByIdList(removeRootFromPathString(pathString), (data) => {
-                contentNameContainer.innerHTML = buildBreadcrumbsString(data);
+                contentBreadcrumbsContainer.innerHTML = buildBreadcrumbsString(data);
             });
         }
     };
--- a/src/bundle/Resources/public/js/scripts/udw/select.location.js
+++ b/src/bundle/Resources/public/js/scripts/udw/select.location.js
@@ -63,26 +63,26 @@
     };
     const updateBreadcrumbsState = (btn, pathString) => {
         const pathStringInput = doc.querySelector(btn.dataset.locationPathInputSelector);
-        const contentBreadcrumbs = doc.querySelector(btn.dataset.contentBreadcrumbsSelector);
-        const contentBreadcrumbsContainer = contentBreadcrumbs.querySelector('.ez-filters__breadcrumbs');
-        const contentBreadcrumbsSpinner = contentBreadcrumbs.querySelector('.ez-filters__breadcrumbs-spinner');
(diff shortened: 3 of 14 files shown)


src/bundle/Controller/SearchController.php
```diff
@@ -172,6 +174,10 @@ public function searchAction(Request $request): Response
                         $creator->id
                     );
                 }
+                $content = null;
```
dew326: Looks unused?

src/bundle/Resources/config/services/components/search.yml
```diff
@@ -0,0 +1,10 @@
+services:
+    search.udw.subtree:
+        parent: EzSystems\EzPlatformAdminUi\Component\ScriptComponent
+        autowire: true
+        autoconfigure: false
+        public: false
+        arguments:
+            $src: '/bundles/ezplatformadminui/js/scripts/udw/select.location.js'
```
dew326: Why do you need this add by the component? It's was created to add JS/CSS from another bundle, not from the same. I think you can add this easily in Twig.

src/bundle/Resources/public/js/scripts/admin.search.filters.js
```diff
@@ -276,6 +280,8 @@
         sectionSelect.addEventListener('change', toggleDisabledStateOnApplyBtn, false);
     }
 
+    subtreeInput.addEventListener('change', toggleDisabledStateOnApplyBtn, false);
+
```
dew326: Please remove this empty line.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,88 @@
...
+        fetch(request)
+            .then(window.eZ.helpers.request.getJsonFromResponse)
+            .then((json) => callback(json))
+            .catch(() => window.eZ.helpers.notification.showErrorNotification('Cannot find children locations by a parent location id'));
+    };
+
+    const removeRootFromPathString = (pathString) => {
+        const pathArray = pathString.split('/').filter((val) => val);
+
+        return pathArray.splice(1, pathArray.length - 1).join('/');
```
dew326: If you return here an array you will not need to do `replace` but just `join('-')` and second `join()`.
So it's 2 joins vs join, 2 replaces.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,88 @@
...
+            },
+            body,
+            mode: 'same-origin',
+            credentials: 'same-origin',
+        });
+
+        fetch(request)
+            .then(window.eZ.helpers.request.getJsonFromResponse)
+            .then((json) => callback(json))
+            .catch(() => window.eZ.helpers.notification.showErrorNotification('Cannot find children locations by a parent location id'));
```
dew326: missing translation

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,88 @@
...
+        const pathArray = pathString.split('/').filter((val) => val);
+
+        return pathArray.splice(1, pathArray.length - 1).join('/');
+    };
+    const closeUDW = () => ReactDOM.unmountComponentAtNode(udwContainer);
+    const onConfirm = (btn, items) => {
+        closeUDW();
+
+        const pathString = items[0].pathString;
+        const displayWrapper = doc.querySelector(btn.dataset.displayWrapper) || null;
```
dew326: If the querySelector will not find element it will return null, no point in adding `|| null`

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,88 @@
...
+        const pathStringInput = doc.querySelector(btn.dataset.locationPathInputSelector);
+
+        pathStringInput.value = pathString;
+        pathStringInput.dispatchEvent(new Event('change'));
+
+        if (!displayWrapper) {
+
```
dew326: no needed empty line

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,88 @@
...
+            return;
+        }
+
+        findLocationsByPathString(removeRootFromPathString(pathString), function (data) {
```
dew326: `(data) => {`

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,88 @@
...
+            displayWrapper.querySelector(btn.dataset.selectedContentNameSelector).innerHTML =
+                data.View.Result.searchHits.searchHit.map(function (searchHit) {
```
dew326: `(searchHit) => {`

src/bundle/Resources/views/admin/search/search_form.html.twig
```diff
@@ -55,6 +55,30 @@
...
         <ul class="ez-filters__user-list ez-filters__user-list--hidden"></ul>
     </div>
+    <div class="ez-filters__item ez-filters__item--subtree">
+        <label class="ez-filters__item-label">{{ 'search.subtree'|trans|desc('Subtree:') }}</label>
+        {% set subtree = form.vars.data.subtree %}
+        <button data-universaldiscovery-title="{{'search.udw.select_content'|trans|desc('Select Content')}}"
+                class="ez-button-tree pure-button ez-font-icon ez-button btn btn-secondary btn--udw-select-location"
+                data-location-path-input-selector="#{{form.subtree.vars.id}}"
+                data-selected-content-name-selector="#{{form.subtree.vars.id}}-selected-content-name"
+                data-starting-location-id="1"
```
dew326: Why this is hardcoded?

src/bundle/Resources/views/admin/search/search_form.html.twig
```diff
@@ -55,6 +55,30 @@
         </svg>
         <ul class="ez-filters__user-list ez-filters__user-list--hidden"></ul>
     </div>
+    <div class="ez-filters__item ez-filters__item--subtree">
+        <label class="ez-filters__item-label">{{ 'search.subtree'|trans|desc('Subtree:') }}</label>
+        {% set subtree = form.vars.data.subtree %}
+        <button data-universaldiscovery-title="{{'search.udw.select_content'|trans|desc('Select Content')}}"
```
dew326: `data-universal-discovery-title`

src/bundle/Resources/views/admin/search/search_form.html.twig
```diff
@@ -55,6 +55,30 @@
...
+                class="ez-button-tree pure-button ez-font-icon ez-button btn btn-secondary btn--udw-select-location"
+                data-location-path-input-selector="#{{form.subtree.vars.id}}"
+                data-selected-content-name-selector="#{{form.subtree.vars.id}}-selected-content-name"
+                data-starting-location-id="1"
+                data-udw-config="{{ ez_udw_config('single_container', {}) }}"
+                data-display-wrapper=".ez-filters__item--subtree"
+        >{{'search.select_content'|trans|desc('Select Content')}}</button>
+        <div id="{{form.subtree.vars.id}}-selected-content-name">
+            <label class="{{form.subtree.vars.id}}-selected-content-name">
```
dew326: I think it's no a `label`.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,88 @@
...
+                    return searchHit.value.Location.ContentInfo.Content.Name;
+                }).join(' / ');
```
dew326: This whole callback can be organized better because it's not readable.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,88 @@
+(function (global, doc, React, ReactDOM) {
+    const btns = doc.querySelectorAll('.btn--udw-select-location');
```
dew326: `ez-btn--udw-select-location` I think the prefix is missing.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,94 @@
...
+            'universal_discovery_widget'
+        );
+        fetch(request)
+            .then(window.eZ.helpers.request.getJsonFromResponse)
+            .then((json) => callback(json))
+            .catch(() => window.eZ.helpers.notification.showErrorNotification(errorMessage));
+    };
+
+    const removeRootFromPathString = (pathString) => {
+        const pathArray = pathString.split('/').filter((val) => val);
```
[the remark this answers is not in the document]
ViniTou: Otherwise we have empty string in first and last element of array.

src/bundle/Resources/views/admin/search/search_form.html.twig
```diff
@@ -55,6 +55,29 @@
         </svg>
         <ul class="ez-filters__user-list ez-filters__user-list--hidden"></ul>
     </div>
+    <div class="ez-filters__item ez-filters__item--subtree">
+        <label class="ez-filters__item-label">{{ 'search.subtree'|trans|desc('Subtree:') }}</label>
+        {% set subtree = form.vars.data.subtree %}
+        <button data-universal-discovery-title="{{'search.udw.select_content'|trans|desc('Select Content')}}"
+                class="ez-button-tree pure-button ez-font-icon ez-button btn btn-secondary ez-btn--udw-select-location"
```
sunpietro: Too big indentation. Please, stick to 4 spaces.

src/bundle/Resources/public/js/scripts/admin.search.filters.js
```diff
@@ -60,7 +63,8 @@
         const isModifiedSelected = !!lastModifiedSelect.value;
         const isCreatedSelected = !!lastCreatedSelect.value;
         const isCreatorSelected = !!searchCreatorInput.value;
-        const isEnabled = isContentTypeSelected || isSectionSelected || isModifiedSelected || isCreatedSelected || isCreatorSelected;
+        const isSubtreeSelected = !!subtreeInput.value;
```
sunpietro: I suppose this is a regular input, right? If so, then it should be: `!!subtreeInput.value.trim().length`. Thanks to this you'll avoid checking a value with spaces only.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,94 @@
...
+                'X-Siteaccess': siteaccess,
+                'X-CSRF-Token': token
+            },
+            body,
+            mode: 'same-origin',
+            credentials: 'same-origin',
+        });
+
+        const errorMessage = Translator.trans(
+            /*@Desc("Cannot find children locations with given id list")*/ 'select_location.error',
```
sunpietro: I'd expose the `id` in the message, as well.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,94 @@
...
+        const pathStringInput = doc.querySelector(btn.dataset.locationPathInputSelector);
+
+        pathStringInput.value = pathString;
+        pathStringInput.dispatchEvent(new Event('change'));
+
+        if (!displayWrapper) {
+            return;
+        }
+
+        findLocationsByIdList(removeRootFromPathString(pathString), (data) => {
```
sunpietro: What the callback does? Is it building a kind of breadcrumb?

src/bundle/Resources/views/admin/search/search_form.html.twig
sunpietro: Another thing, I believe there are too many different classes for buttons. I believe they override each other. What's your take on this @dew326 ?

src/bundle/Resources/views/admin/search/search_form.html.twig
```diff
@@ -55,6 +55,29 @@
...
+                data-location-path-input-selector="#{{form.subtree.vars.id}}"
+                data-selected-content-name-selector="#{{form.subtree.vars.id}}-selected-content-name"
+                data-udw-config="{{ ez_udw_config('single_container', {}) }}"
+                data-display-wrapper=".ez-filters__item--subtree"
+        >{{'search.select_content'|trans|desc('Select Content')}}</button>
+        <div id="{{form.subtree.vars.id}}-selected-content-name">
+            <span class="{{form.subtree.vars.id}}-selected-content-name">
```
sunpietro: This class seems to be unused, am I right?

src/bundle/Resources/views/admin/search/search_form.html.twig
```diff
@@ -55,6 +55,29 @@
...
+        <label class="ez-filters__item-label">{{ 'search.subtree'|trans|desc('Subtree:') }}</label>
+        {% set subtree = form.vars.data.subtree %}
+        <button data-universal-discovery-title="{{'search.udw.select_content'|trans|desc('Select Content')}}"
+                class="ez-button-tree pure-button ez-font-icon ez-button btn btn-secondary ez-btn--udw-select-location"
+                data-location-path-input-selector="#{{form.subtree.vars.id}}"
+                data-selected-content-name-selector="#{{form.subtree.vars.id}}-selected-content-name"
+                data-udw-config="{{ ez_udw_config('single_container', {}) }}"
+                data-display-wrapper=".ez-filters__item--subtree"
+        >{{'search.select_content'|trans|desc('Select Content')}}</button>
+        <div id="{{form.subtree.vars.id}}-selected-content-name">
```
sunpietro: This is strange combination. What is the output value for `id` attribute?

src/bundle/Resources/views/admin/search/search_form.html.twig
dew326: Well, I think the `pure-button` is from PlatformUI v1.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,94 @@
+(function (global, doc, React, ReactDOM, Translator) {
+    const btns = doc.querySelectorAll('.ez-btn--udw-select-location');
+    const udwContainer = doc.getElementById('react-udw');
+    const token = doc.querySelector('meta[name="CSRF-Token"]').content;
+    const siteaccess = doc.querySelector('meta[name="SiteAccess"]').content;
+
```
dew326: You can remove this empty line.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,94 @@
...
+            onConfirm: onConfirm.bind(null, event.currentTarget),
+            onCancel,
+            title: event.currentTarget.dataset.universalDiscoveryTitle,
+            multiple: false,
+            startingLocationId: window.eZ.adminUiConfig.universalDiscoveryWidget.startingLocationId,
+            restInfo: { token, siteaccess }
+        }, config)), udwContainer);
+    };
+
+    btns.forEach(btn => btn.addEventListener('click', openUDW, false));
```
dew326: I think we are missing `clear selected subtree`? Because I don't see any JS for it.
screenshot at nov 15 14-32-36

src/bundle/Resources/public/js/scripts/udw/select.location.js
ViniTou: Yup.

src/bundle/Resources/public/js/scripts/udw/select.location.js
ViniTou: I left that out as redesign is not in scope of this PR. It will be done by @tischsoic

src/bundle/Resources/public/js/scripts/admin.search.filters.js
ViniTou: This is hidden input filled from UDW, so don't know if this make a lot of difference in that case.

sunpietro: Well, when I see the `selected` or any similar word in the variable name, then I would expect a boolean value by default. Having an input, not checkbox or radio, in this case might be confusing.
Read it out loud. To select a subtree you have to type anything in the input and the variable name is about selection.

src/bundle/Resources/views/admin/search/search_form.html.twig
ViniTou: selected location Id.

src/bundle/Resources/public/js/scripts/admin.search.filters.js
ViniTou: Same logic is applied to creator or dates input as well as you don't have checkboxes there.

src/bundle/Resources/public/js/scripts/udw/select.location.js
tischsoic: I will provide PR with styling after merge of this one, because there will be also changes in other places.

tischsoic:
ping @dew326 @sunpietro

src/bundle/Resources/public/js/scripts/udw/select.location.js
dew326: Why not push changes into this branch? We will have a better overview of changes and we will be able to perform proper QA on this.

tischsoic: Because it involves restyling whole filters menu, not only Sub-tree filter.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,93 @@
...
+            .catch(() => window.eZ.helpers.notification.showErrorNotification(errorMessage));
+    };
+    const removeRootFromPathString = (pathString) => {
+        const pathArray = pathString.split('/').filter((val) => val);
+
+        return pathArray.splice(1, pathArray.length - 1);
+    };
+    const buildBreadCrumbsString = (viewData) => {
+        const searchHitList = viewData.View.Result.searchHits.searchHit;
+        return searchHitList.map((searchHit) => {
```
dew326: please add empty line before `return`

src/bundle/Resources/public/js/scripts/udw/select.location.js
micszo: Do you mean that only styling is out of scope of this PR or also some core actions related to the new feature? In the latter case completing QA will not be possible here. We had a similar case here 
https://github.com/ezsystems/ezplatform-admin-ui/pull/697 and that's why it wasn't merged.

tischsoic: In offline discussion with @ViniTou we decided that we I will push my changes to this PR. :)

tischsoic:
ping @dew326 @sunpietro 
Restyling changes are in the last commit.

src/bundle/Resources/views/form_fields.html.twig
```diff
@@ -37,10 +37,10 @@
 {%- endblock -%}
 
 {%- block _search_content_types_widget -%}
-    <div class="ez-content-type-selector ez-content-type-selector--collapsed">
+    <div class="ez-content-type-selector">
     {%- set options = choices -%}
     {%- for group_label, choice in options -%}
-        <div class="ez-content-type-selector__group">
+        <div class="ez-content-type-selector__group {% if not loop.first %}ez-content-type-selector__group--collapsed{% endif %}">
```
dew326: So we will still have this collapsed content types?

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,131 @@
...
+        ReactDOM.render(
+            React.createElement(
+                global.eZ.modules.UniversalDiscovery,
+                Object.assign(
+                    {
+                        onConfirm: onConfirm.bind(null, event.currentTarget),
+                        onCancel,
+                        title: event.currentTarget.dataset.universalDiscoveryTitle,
+                        multiple: false,
+                        startingLocationId: window.eZ.adminUiConfig.universalDiscoveryWidget.startingLocationId,
```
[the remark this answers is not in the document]
tischsoic: This must be copied from other place, I will change this.

src/bundle/Resources/views/form_fields.html.twig
tischsoic: Yes, as you can see on screens in PR description and prototypes. :)

tischsoic: But we no longer display them in the popup-like `div`.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,131 @@
...
+    };
+    const buildBreadCrumbsString = (viewData) => {
+        const searchHitList = viewData.View.Result.searchHits.searchHit;
+
+        return searchHitList.map((searchHit) => searchHit.value.Location.ContentInfo.Content.Name).join(' / ');
+    };
+    const toggleVisibility = (btn, pathString) => {
+        btn.hidden = !!pathString;
+
+        const contentNameWrapper = doc.querySelector(btn.dataset.selectedContentNameSelector);
```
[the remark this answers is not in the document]
tischsoic: Right, I will change name of this variable + name of the data-attribute.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,131 @@
...
+    };
+    const buildBreadcrumbsString = (viewData) => {
+        const searchHitList = viewData.View.Result.searchHits.searchHit;
+
+        return searchHitList.map((searchHit) => searchHit.value.Location.ContentInfo.Content.Name).join(' / ');
+    };
+    const toggleVisibility = (btn, pathString) => {
+        btn.hidden = !!pathString;
+
+        const contentNameWrapper = doc.querySelector(btn.dataset.contentBreadcrumbsSelector);
```
dew326: `contentBreadcrumbsWrapper`

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,131 @@
...
+            contentNameWrapper.hidden = !pathString;
+        }
+    };
+    const setPath = (btn, pathString) => {
+        const pathStringInput = doc.querySelector(btn.dataset.locationPathInputSelector);
+
+        pathStringInput.value = pathString;
+        pathStringInput.dispatchEvent(new Event('change'));
+
+        const contentNameContainer = doc.querySelector(`${btn.dataset.contentBreadcrumbsSelector} div`);
```
dew326: `contentBreadcrumbsContainer`

dew326: The selector `div` is not a good one. Why this has no class?

tischsoic: Because we decided not to create separate block (BEM terminology) for `select.location` component and I didn't want to add too much elements to `ez-filters` block. 
I can add some `ez-filters` class to this `div` but then I think we should change this selector:
`const btns = doc.querySelectorAll('.ez-btn--udw-select-location');`
What do you think?

sunpietro: @tischsoic this is highly incorrect approach to use selectors like `.some-class div` in JS code without seriously good reason. Whenever you want to access a specific DOM node you should add a CSS class to it, to avoid selector chaining and CSS specificity.
This `div` belongs to some component, right? So it has some meaning. It is a part of something that can be named. Can you point me the place this `div` resides, maybe I can help you with constructing the selector properly?

tischsoic: The `div` is here: https://github.com/ezsystems/ezplatform-admin-ui/pull/711/files#diff-f5eca6a947a5a0955802071e5723d9c5R53
It actually has a class, I didn't use it because it corresponds to the `ez-filters` which for me is not generic enough as we use this selector: `doc.querySelectorAll('.ez-btn--udw-select-location');` (https://github.com/ezsystems/ezplatform-admin-ui/pull/711/files#diff-490854292fbfb1577b3b5bbefa54df28R2) to search for "select location" buttons.

Should I use some other CSS class or add id to this div?

dew326: Wait, are you trying to do some generic solution to be able to reuse this in other places? If so, this is a bad approach to achieve that.

You should create a class where you can override the selectors/methods on so.

Do we need this in other places? I think there is no point to spend time on this. When we will need this in other places we can change it (Once we will introduce this as a generic solution we will need to keep BC on this).

tischsoic: I will just use this class from `div`.

tischsoic:
ping @dew326 @sunpietro

lserwatka:
What is the status here? Can we move this PR to QA?

sunpietro:
@lserwatka it's in review.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,131 @@
...
+            credentials: 'same-origin',
+        });
+        const errorMessage = Translator.trans(
+            /*@Desc("Cannot find children locations with given id - %idList%")*/ 'select_location.error',
+            { idList: idList.join(',') },
+            'universal_discovery_widget'
+        );
+
+        fetch(request)
+            .then(window.eZ.helpers.request.getJsonFromResponse)
```
dew326: `eZ.helpers.request.getJsonFromResponse`

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,131 @@
...
+            .then((json) => callback(json))
+            .catch(() => window.eZ.helpers.notification.showErrorNotification(errorMessage));
```
dew326: `eZ.helpers.notification.showErrorNotification(errorMessage)`

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,131 @@
...
+    };
+    const buildBreadcrumbsString = (viewData) => {
+        const searchHitList = viewData.View.Result.searchHits.searchHit;
+
+        return searchHitList.map((searchHit) => searchHit.value.Location.ContentInfo.Content.Name).join(' / ');
+    };
+    const toggleVisibility = (btn, pathString) => {
+        btn.hidden = !!pathString;
+
+        const contentBreadcrumbsWrapper = doc.querySelector(btn.dataset.contentBreadcrumbsSelector);
```
dew326: this should be at the top o the method.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,131 @@
...
+            contentBreadcrumbsWrapper.hidden = !pathString;
+        }
+    };
+    const setPath = (btn, pathString) => {
+        const pathStringInput = doc.querySelector(btn.dataset.locationPathInputSelector);
+
+        pathStringInput.value = pathString;
+        pathStringInput.dispatchEvent(new Event('change'));
+
+        const contentBreadcrumbsContainer = doc.querySelector(`${btn.dataset.contentBreadcrumbsSelector} ez-filters__subtree__breadcrumbs`);
```
dew326: should be at the top of the method

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,131 @@
...
+        const pathArray = pathString.split('/').filter((val) => val);
+
+        return pathArray.splice(1, pathArray.length - 1);
+    };
+    const buildBreadcrumbsString = (viewData) => {
+        const searchHitList = viewData.View.Result.searchHits.searchHit;
+
+        return searchHitList.map((searchHit) => searchHit.value.Location.ContentInfo.Content.Name).join(' / ');
+    };
+    const toggleVisibility = (btn, pathString) => {
```
dew326: This method doesn't need `pathString`, it should take the boolean `show`.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,131 @@
+(function(global, doc, eZ, React, ReactDOM, Translator){
```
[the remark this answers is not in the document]
tischsoic: I don't know what you mean. :)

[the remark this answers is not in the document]
sunpietro: I believe Prettier haven't been run in this file. Because, there's missing space before the function opening curly bracket.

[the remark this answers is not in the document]
tischsoic: I run prettier in this file. I think that at one point at the end of editing I copied `(function(global, doc, eZ, React, ReactDOM, Translator){` from other file, thus missing space. Everything else seems to be formatted.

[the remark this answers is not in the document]
tischsoic: Hmm, I don't know, I was also surprised.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,132 @@
...
+        btn.hidden = isLocationSelected;
+
+        if (contentBreadcrumbsWrapper) {
+            contentBreadcrumbsWrapper.hidden = !isLocationSelected;
+        }
+    };
+    const setLocationPath = (btn, pathString) => {
+        const pathStringInput = doc.querySelector(btn.dataset.locationPathInputSelector);
+        const contentBreadcrumbsContainer = doc.querySelector(
+            `${btn.dataset.contentBreadcrumbsSelector} .ez-filters__subtree__breadcrumbs`
```
sunpietro: Improper usage of BEM naming convention. BEM stands for `Block`, `Element`, `Modifier`. The way you created a CSS class makes it BEEM: `Block`, `Element`, `Element`, `Modifier`.
I think the class name should be `ez-filters__breadcrumbs` or make the subtree part as a separate component.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,132 @@
...
+    const toggleVisibility = (btn, isLocationSelected) => {
+        const contentBreadcrumbsWrapper = doc.querySelector(btn.dataset.contentBreadcrumbsSelector);
+
+        btn.hidden = isLocationSelected;
+
+        if (contentBreadcrumbsWrapper) {
+            contentBreadcrumbsWrapper.hidden = !isLocationSelected;
+        }
+    };
+    const setLocationPath = (btn, pathString) => {
```
sunpietro: Well, what function does is filling the breadcrumbs label with a location path. Hence, I would name it: `updateBreadcrumbsLabel` or `updateBreadcrumbsPath`

src/bundle/Resources/public/scss/_filters.scss
```diff
@@ -37,26 +57,47 @@
...
+        .ez-icon--reset {
+            display: block;
+            position: absolute;
+            top: calculateRem(7px);
+            right: calculateRem(6px);
+            cursor: pointer;
+            fill: $ez-black;
+        }
+
+        &__breadcrumbs {
```
sunpietro: Wrong usage of BEM

src/bundle/Resources/views/admin/search/search_form.html.twig
```diff
@@ -12,48 +12,86 @@
...
+                {{ form_widget(form.section, {'attr': {'class': 'ez-filters__select'}}) }}
+            </div>
+        {% endif %}
+        <div class="ez-filters__item ez-filters__item--subtree">
+            <label class="ez-filters__item-label">{{ 'search.subtree'|trans|desc('Subtree:') }}</label>
+            {% set subtree = form.vars.data.subtree %}
+            <div>
+                <button class="btn btn-secondary ez-btn--udw-select-location"
+                    type="button"
+                    {% if form.subtree.vars.value is not empty %} hidden="true" {% endif %}
```
sunpietro: Just `hidden`

src/bundle/Resources/public/js/scripts/udw/select.location.js
tischsoic: Hmm, I thought that in some cases we can have `grandchild`, I will change this then.

src/bundle/Resources/public/js/scripts/udw/select.location.js
tischsoic: It also updates input value `pathStringInput.value = pathString;`.

src/bundle/Resources/views/admin/search/search_form.html.twig
tischsoic: I actually prefer `hidden` but I thought that we have convention to use `hidden="true"`. Also, we have many examples of: `{'attr': {'hidden': 'hidden', `. So I'm confused a little bit. Do we have some convention? Or maybe this `{'attr': {'hidden': 'hidden', ` is different from just `hidden`, because the second one is in tag?

tischsoic:
ping @sunpietro

src/bundle/Resources/views/admin/search/search_form.html.twig
sunpietro: When using Twig variables and object you can't set the attribute without value. But when writing in HTML you don't have to write `hidden="hidden"` because `hidden` is enough.

sunpietro:
I did a small manual test and I found an issue with generating subtree breadcrumb thing.
After selecting a location it is generated empty first and shortly after it is given non-empty label.
IMO, this is incorrect behaviour. The full label should be visible right away with no _lag_ effect.

src/bundle/Resources/views/admin/search/search_form.html.twig
tischsoic: AFAIK you can have empty attribute in twig: `{'attr': {'hidden': null}}`. But I think we use `{'attr': {'hidden': 'hidden'}}` so that it is more obvious, because `'hidden': null` may suggest that attribute won't be set. 

But here we have HTML, so I will change this to just `hidden`.

tischsoic:
ping @sunpietro

sunpietro:
> @sunpietro Label elements are fetch from the server, there will always be a lag. Do you want to uncover label UI after it is fetched, or display some spinner?

It should display some kind of loading state while the info is not ready and display a path when it's ready.

tischsoic:
ping @sunpietro  - spinner added.

src/bundle/Resources/public/js/scripts/udw/select.location.js
```diff
@@ -0,0 +1,136 @@
...
+
+        if (contentBreadcrumbsWrapper) {
+            contentBreadcrumbsWrapper.hidden = !isLocationSelected;
+        }
+    };
+    const updateBreadcrumbsState = (btn, pathString) => {
+        const pathStringInput = doc.querySelector(btn.dataset.locationPathInputSelector);
+        const contentBreadcrumbs = doc.querySelector(btn.dataset.contentBreadcrumbsSelector);
+        const contentBreadcrumbsContainer = contentBreadcrumbs.querySelector('.ez-filters__breadcrumbs');
+        const contentBreadcrumbsSpinner = contentBreadcrumbs.querySelector('.ez-filters__breadcrumbs-spinner');
```
sunpietro: cc @inakijv can you check the final result of showing the loading spinner while fetching data about the location path? Please, tell us whether you like it.

tischsoic:
As agreed with @SylvainGuittard, tags with selected filters (displayed above filter menu) are considered as another task. 😄 
You can proceed with tests @katarzynazawada . 😝

src/bundle/Resources/public/scss/_filters.scss
```diff
@@ -78,7 +78,6 @@
         padding: calculateRem(8px) calculateRem(48px) calculateRem(8px)  calculateRem(8px);
         display: inline-block;
         position: relative;
-        height: calculateRem(38px);
```
sunpietro: maybe just `min-height` instead?

tischsoic: Do we need `min-height`? You would like to have `min-height` because it will be then more error prone?

sunpietro: It was just a question. Maybe you wanted the element to be at least 38px tall, am I right?

PR #712, opened 2018-11-09
EZP-29747: As an editor, I want to see a list of content items impossible to delete

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29747
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Implements modal for bulk delete/move with list of failed items.

Related PR in `ezplatform-admin-ui-modules`: https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/113

## Screenshots

screen shot 2018-12-06 at 17 15 11

screen shot 2018-12-12 at 15 47 57

#### Checklist:
- [ ] Ready for Code Review


src/bundle/Resources/public/js/scripts/admin.location.view.js
```diff
@@ -55,13 +59,56 @@
...
+    const setModalTitle = (title) => {
+        const modalTitleNode = doc.querySelector(`${SELECTOR_BULK_OPERATION_FAILED_MODAL} .modal-header .modal-title`);
 
-    listContainers.forEach(container => {
+        modalTitleNode.innerHTML = title;
+    };
+    const setModalTableBody = (failedItemsData) => {
+        const modal = doc.querySelector(SELECTOR_BULK_OPERATION_FAILED_MODAL);
+        const table = modal.querySelector('table');
+        const tableBody = table.querySelector('tbody');
```
dew326: Selectors as a tag name are not the best. Please add some classes to this table.

src/bundle/Resources/views/content/modal_location_bulk_operation_failed.html.twig
```diff
@@ -0,0 +1,46 @@
...
+                    aria-label="Close">
+                    <svg
+                        class="ez-icon ez-icon--medium"
+                        aria-hidden="true">
+                        <use xlink:href="{{ asset('bundles/ezplatformadminui/img/ez-icons.svg') }}#discard"></use>
+                    </svg>
+                </button>
+            </div>
+            <div class="modal-body">
+                <div class="ez-table-header"></div>
```
dew326: Seems unused.

src/bundle/Resources/views/content/modal_location_bulk_operation_failed.html.twig
```diff
@@ -0,0 +1,46 @@
...
+    id="bulk-operation-failed-modal"
+    tabindex="-1"
+    role="dialog">
+    <div
+        class="modal-dialog"
+        role="document">
+        <div class="modal-content">
+            <div class="modal-header">
+                <h5 class="modal-title">
+
```
dew326: Not needed. Can be in one line

src/bundle/Resources/views/content/modal_location_bulk_operation_failed.html.twig
```diff
@@ -0,0 +1,46 @@
+<div
+    class="modal fade ez-bulk-operation-failed-modal"
+    id="bulk-operation-failed-modal"
+    tabindex="-1"
+    role="dialog">
+    <div
+        class="modal-dialog"
```
dew326: This is less readable. If the line is really long it's fine to split it in few but in this case, (and in some other here) I think it's not needed.

src/bundle/Resources/views/content/modal_location_bulk_operation_failed.html.twig
tischsoic: This table looks bad without this header.

dew326: I see, but on the mockups there is different table:
mockup

tischsoic: Yes, but during discussing on this feature with PM I've got feedback that we want a different table here. Probably there will be a title but I didn't get final feedback what title do we want.

src/bundle/Resources/public/js/scripts/admin.location.view.js
tischsoic: Do we want class on every table element?

tischsoic:
Changes applied in the last commit - ping @dew326 .

src/bundle/Resources/public/js/scripts/admin.location.view.js
dew326: On the elements that are relevant from JS or CSS perspective

tischsoic: My question was a bit imprecise, if we want to refer in JS `table` and `tbody` elements and `tbody` is inside `table`, we need class for both?

dew326: Of course. We should avoid using tag name as a selector.

tischsoic: OK, I added `.ez-bulk-operation-failed-modal__table` and `.ez-bulk-operation-failed-modal__table-body`.

src/bundle/Resources/views/content/modal_location_bulk_operation_failed.html.twig
micszo: @tischsoic what's the status here? :) is the table state final?

src/bundle/Resources/public/js/scripts/admin.location.view.js
```diff
@@ -1,4 +1,5 @@
 (function(global, doc, $, React, ReactDOM, eZ, Routing) {
+    const SELECTOR_BULK_OPERATION_FAILED_MODAL = '.ez-bulk-operation-failed-modal';
```
sunpietro: `SELECTOR_MODAL_BULK_ACTION_FAIL`

src/bundle/Resources/public/js/scripts/admin.location.view.js
```diff
@@ -1,4 +1,5 @@
...
     const listContainers = [...doc.querySelectorAll('.ez-sil')];
```
sunpietro: Can you convert it to: `const listContainers = doc.querySelectorAll('.ez-sil');`?

src/bundle/Resources/public/js/scripts/admin.location.view.js
```diff
@@ -59,6 +60,47 @@
         });
     };
     const generateLink = (locationId) => Routing.generate('_ezpublishLocation', { locationId });
+    const setModalTableTitle = (title) => {
+        const modalTableTitleNode = doc.querySelector(
+            `${SELECTOR_BULK_OPERATION_FAILED_MODAL} .ez-table-header .ez-table-header__headline`
```
sunpietro: It should be: 
```
`${SELECTOR_BULK_OPERATION_FAILED_MODAL} .ez-table-header__headline`
```

src/bundle/Resources/public/js/scripts/admin.location.view.js
sunpietro: I don't like this name: `.ez-bulk-operation-failed-modal`

src/bundle/Resources/public/js/scripts/admin.location.view.js
```diff
@@ -59,6 +60,47 @@
...
+            const container = doc.createElement('tbody');
+            const renderedItem = tableRowTemplate
+                .replace('{{ content_name }}', contentName)
+                .replace('{{ content_type_name }}', contentTypeName);
+
+            container.insertAdjacentHTML('beforeend', renderedItem);
+
+            const tableRowNode = container.querySelector('tr');
+
+            fragment.append(tableRowNode);
```
sunpietro: You should not do it this way. It's not performant. The `append` should be invoked when all the actions, for each item, in a loop are done.

sunpietro: You should care about it even if it's a `Fragment`;

src/bundle/Resources/public/js/scripts/admin.location.view.js
```diff
@@ -94,6 +136,7 @@
                 contentTypesMap,
                 totalCount: subItemsList.ChildrenCount,
                 udwConfigBulkMoveItems,
+                onShowBulkOperationFailedModal: showBulkOperationFailedModal,
```
sunpietro: Please, read it out loud: `onShowBulkOperationFailedModal: showBulkOperationFailedModal`. Does it make sense to you? Maybe you can simplify it and avoid a feeling of duplication?

src/bundle/Resources/public/js/scripts/helpers/notification.helper.js
```diff
@@ -10,7 +10,10 @@
      * Dispatches notification event
      *
      * @function showNotification
-     * @param {{message: string, label: string}} detail
+     * @param {Object} detail
+     * @param {String} detail.message
+     * @param {String} detail.label
+     * @param {Function} [detail.callback] Called after notification Node was added
      */
     const showNotification = (detail) => {
```
sunpietro: Ok, you're passing the callback here, but when it's invoked?

src/bundle/Resources/public/scss/_bulk-operation-failed-modal.scss
```diff
@@ -0,0 +1,25 @@
+.ez-bulk-operation-failed-modal {
+    .modal-dialog {
+        max-width: 60vw;
+    }
+
+    .modal-header {
+        background-color: $ez-color-base-pale;
+
+        .modal-title {
```
sunpietro: Do you need to nest this selector and the one below?

src/bundle/Resources/public/scss/_bulk-operation-failed-modal.scss
```diff
@@ -0,0 +1,25 @@
...
+            font-size: calculateRem(24px);
+        }
+
+        .close {
+            margin-top: calculateRem(-9px);
+        }
+    }
+
+    .modal-body {
```
sunpietro: Maybe you can use the Bootstrap utility classes instead? https://getbootstrap.com/docs/4.1/utilities/spacing/

src/bundle/Resources/views/content/modal_location_bulk_operation_failed.html.twig
```diff
@@ -0,0 +1,39 @@
+{% trans_default_domain 'locationview' %}
+
+<div
+    class="modal fade ez-bulk-operation-failed-modal"
```
sunpietro: Why do you duplicate `id` as a class name?

src/bundle/Resources/public/js/scripts/helpers/notification.helper.js
tischsoic: Hmm, `Called after notification Node was added` from jsDoc is not precise enough?
Callback is called right after we append notification element to the DOM.

src/bundle/Resources/public/js/scripts/admin.location.view.js
tischsoic: So you want to change all names `bulk-operation-failed` to `bulk-action-failed`?

src/bundle/Resources/public/js/scripts/helpers/notification.helper.js
sunpietro: The thing is you're dispatching it with an event. So it may be invoked and may be not as well. What's more, I'm missing an information about the kind of this callback. Is it invoked on success or on fail?

src/bundle/Resources/public/js/scripts/admin.location.view.js
sunpietro: Yes, it's shorter but still there's something in the class name I don't like. Can't it be even shorter by using different words describing the same element?

tischsoic: Unfortunately, I was unable to come up with shorter name. 😕

src/bundle/Resources/public/js/scripts/admin.location.view.js
tischsoic: I did this as it was done here: https://github.com/ezsystems/ezplatform-page-builder/blob/911eb290f55a5c98638e5518cd35d340cb7764e5/src/bundle/Resources/public/js/config-form/widgets/collection.js#L59 

> The append should be invoked when all the actions, for each item, in a loop are done.
I'm calling `.append` at the end of each iteration. I'm not sure what changes you propose.

src/bundle/Resources/public/js/scripts/admin.location.view.js
tischsoic: So, `onShowBulkOperationFailedModal: showBulkOperationFailedModal` -> `showBulkOperationFailedModal, ` ?

src/bundle/Resources/public/js/scripts/helpers/notification.helper.js
tischsoic: >  Is it invoked on success or on fail?
I don't know what do you mean by _fail_?
Could you please specify what how you want it to look like? You want different name for callback, different description in jsDoc, different behaviour of the whole mechanism?

src/bundle/Resources/views/content/modal_location_bulk_operation_failed.html.twig
tischsoic: I think that if we want to be compliant with BEM and use classes for elements (like `ez-bulk-operation-failed-modal__table-body`) we need class for whole block. I also use this class to style modal.

src/bundle/Resources/public/scss/_bulk-operation-failed-modal.scss
tischsoic: Hmm, we nest this selector in other places, but I checked it and we don't have to do this. 
Also, I modeled this on notifications modal, which has separate file with CSSes styling modal, but now I realised that we are storing modals CSSes in: https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/bundle/Resources/public/scss/_modals.scss
so these styles should probably be moved to this file?

src/bundle/Resources/public/js/scripts/helpers/notification.helper.js
sunpietro: 1. Please, change the description of `detail.callback` because it promises the callback will be invoked (but it depends on the final implementation of event listener),
2. The `callback` is very generic name. You don't know what is it for. I'm missing context here.

src/bundle/Resources/views/content/modal_location_bulk_operation_failed.html.twig
tischsoic: I checked how it works in other places in our codebase and it seems that we have convention to give `id`s to modals and later refer to them by these ids. It also matches examples from https://getbootstrap.com/docs/4.0/components/modal/.

But I don't use id in my piece of code, instead I have class selector:

```js
const SELECTOR_BULK_OPERATION_FAILED_MODAL = '.ez-bulk-operation-failed-modal';
```

I think that the right way would be to change above selector to:

```js
const SELECTOR_BULK_OPERATION_FAILED_MODAL = '#bulk-operation-failed-modal'
```

and leave `class` for styling puroposes. 

Are you OK with this? 🙂

src/bundle/Resources/public/js/scripts/helpers/notification.helper.js
tischsoic: 1. I will change description to: `to be called after notification Node was added`, which does not tell that it will be invoked but rather that it is intended to be invoked on notification show.
2. I think we may change name to `onShow`. I googled a bit and found that theres even something similar for notifications in PWA: https://developer.mozilla.org/en-US/docs/Web/API/Notification/onshow

sunpietro: I like it. Go on!

tischsoic:
ping @sunpietro @dew326

tischsoic:
Ready for merge, should be merged with: https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/113

PR #714, opened 2018-11-13
EZEE-2303: Impl. RichText block

adamwojs:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZEE-2303](https://jira.ez.no/browse/EZEE-2303)
| Depends on | ezsystems/ezplatform-richtext#13 |
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Refactored the RichText field type to be able to reuse it in the RichText block. 

#### Checklist:
- [x] Ready for Code Review


src/bundle/Resources/views/form_fields.html.twig
```diff
@@ -153,3 +153,18 @@
         {{ block('form_widget') }}
     </div>
 {%- endblock -%}
+
+{%- block richtext_widget -%}
+    {% set attr = attr|merge({ 'hidden': true }) %}
```
tischsoic: How does this work, do we use `attr`?

src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
```diff
@@ -0,0 +1,190 @@
+(function(global) {
+    const eZ = (global.eZ = global.eZ || {});
+
+    const BaseRichtText = class BaseRichtText {
```
tischsoic: BaseRich\*t\*Text :)

src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
```diff
@@ -0,0 +1,190 @@
...
+                    this.xhtmlNamespace,
+                    this.ezNamespace
+                );
+            });
+
+            return alloyEditor;
+        }
+    };
+
+    eZ.BaseRichtText = BaseRichtText;
```
tischsoic: Maybe we should use `eZ.addConfig` here?

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-linkedit.js
```diff
@@ -78,24 +78,24 @@ export default class EzBtnLinkEdit extends Component {
...
-                        },
-                        config
-                    )
-                ),
-                udwContainer
+            const alloyEditorCallbacks = eZ.ezAlloyEditor.callbacks;
+            const mergedConfig = Object.assign(
+                {
+                    onConfirm: udwOnConfirm,
+                    onCancel: () => ReactDOM.unmountComponentAtNode(udwContainer),
```
tischsoic: I wonder, should we pass this in config even if we are using `alloyEditorCallbacks.openUdw`? Probably there is no need to unmount component at `udwContainer` in this case. What do you think?

src/bundle/Resources/views/form_fields.html.twig
adamwojs: Yes, we do. Is used by the `widget_attributes` block called in https://github.com/ezsystems/ezplatform-admin-ui/pull/714/files#diff-c5ade6303c80a553058d3c44229e563bR165

src/bundle/Resources/views/form_fields.html.twig
```diff
@@ -153,3 +153,18 @@
...
+    {# FIXME: hardcoded language code #}
```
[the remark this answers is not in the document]
adamwojs: No, it's not supposed to stay. I will fix it.

src/bundle/Resources/views/form_fields.html.twig
tischsoic: Great, thanks for explanation. :)

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-linkedit.js
dew326: This is for a more generic solution. Someone can set own callback to openUDW and may rely on this method.

src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
dew326: I want to be in line with the rest of the AE files.

sunpietro:
There are conflicts in this branch @adamwojs

adamwojs:
PR updated according to the @tischsoic suggestions (I also squashed the commits)

adamwojs:
@webhdx My point in placing mentioned classes in the `ezplatform-admin-ui` repository was to allow the eZ Platform Developer to use RichText as any other front-end component in their custom administration controllers. 

Place where this feature might be useful is an e-mail templates editor (screenshot from demo prepared for test purposes) zrzut ekranu 2018-11-22 o 12 54 51
or CRUD of custom doctrine entity.  

But in general this feature is something out of scope, so I won't push for it (moving it to the `ezplatform-page-builder` might event simplify things). WDYT ?

src/bundle/Resources/views/form_fields.html.twig
```diff
@@ -153,3 +153,14 @@
...
     </div>
 {%- endblock -%}
+
+{%- block richtext_widget -%}
+    {% set attr = attr|merge({ 'hidden': true }) %}
+
+    <div class="ez-data-source">
+        <textarea {{ block('widget_attributes') }}>{{ value }}</textarea>
+        <div class="ez-data-source__richtext" id="{{ form.vars.id }}__editable"></div>
+        <div class="hidden" data-udw-config-name="richtext_embed" data-udw-config="{{ ez_udw_config('richtext_embed', udw_context) }}"></div>
```
mikadamczyk: nitpick: `richtext_embed` could be extracted to variable

PR #718, opened 2018-11-19
[EZP-29819] Set config providers which use dynamic settings to lazy

emodric:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29819
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

This fixes an issue where the dynamic parameters in arguments are injected to the config provider services in the `default` scope instead of `admin` scope, resulting in admin content edit interface without any styles.

I have not been able to deduce why this is happening, though, since it seems that clean eZ Platform install works fine. It maybe a combination of some other config, but anyway, setting these services to lazy helps.

#### Checklist:
- [x] Ready for Code Review


lserwatka:
Merged up to master.

PR #720, opened 2018-11-21
EZP-29825 Change form templates merge order to let me override form fragments

StephaneDiot:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29825
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | guess not
| Tests pass?   | yes/no
| Doc needed?   | no

For a customer I need to override some specific content edit form fragment depending of the content type of the edited content. (The customer will go in 2.4 when released)

This could be achieved easily with the ViewProvider on the content_edit_view passing a form_templates param containing the specific form fragments I want to display ( see example in jira ticket )

For this I need my custom form templates to be after the standard form templates in the array used by the form_theme.

Thank you.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/content/content_edit/content_edit_base.html.twig
+++ b/src/bundle/Resources/views/content/content_edit/content_edit_base.html.twig
@@ -1,7 +1,7 @@
 {% extends viewbaseLayout is defined ? viewbaseLayout : '@ezdesign/layout.html.twig' %}
 
 {% set default_form_templates = admin_ui_config.contentEditFormTemplates %}
-{% set form_templates = form_templates is defined ? form_templates|merge(default_form_templates) : default_form_templates %}
+{% set form_templates = form_templates is defined ? default_form_templates|merge(form_templates) : default_form_templates %}
 
 {% trans_default_domain 'content_edit' %}
 {% form_theme form with form_templates %}


andrerom:
@webhdx Is this a bug perhaps?

webhdx:
@StephaneDiot you should be able to alter the order via configuration:
https://github.com/ezsystems/ezplatform-admin-ui/blob/2a14ad6e5dea3f504c15673f2b381430e2bc23f6/src/bundle/Resources/config/admin_ui_forms.yml

This change might break UI in case of file loading order change and now block from different file might be rendered.

StephaneDiot:
@webhdx Thank you for your time.
I did notice that I could just override the content edit form fields fragment from here, but that wasn't very useful in my case because I must override those fragments only when editing specifics content types.

lserwatka:
@webhdx let's target this only for master (2.4 release) @DominikaK we need few bits of documentation here

lserwatka:
Note for QA: we need sanities here.

cc @micszo

StephaneDiot:
rebased

lserwatka:
@DominikaK we need a short memo about this in the eelease note for 2.4

DominikaK:
Mention in release notes already prepared in https://github.com/ezsystems/developer-documentation/pull/470/commits/438f8a7dfdc895a4763981300a5ffb6fb42715c3#diff-19de2e0707ca2fdf165e416531cb5309R104

PR #721, opened 2018-11-21
EZP-29827: Harmonized Preview Switcher in Content Section Editing view

inakijv:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-29827](https://jira.ez.no/browse/EZP-29827)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

#### Checklist:
- [X] Ready for Code Review

ez platform 70
ez platform 71
ez platform 72

### DIFF
--- a/src/bundle/Resources/public/scss/_preview.scss
+++ b/src/bundle/Resources/public/scss/_preview.scss
@@ -21,7 +21,7 @@
                 display: flex;
                 justify-content: space-around;
                 align-items: center;
-                width: calculateRem(118px);
+                max-width: calculateRem(118px);
                 height: calculateRem(44px);
                 border: calculateRem(2px) solid $ez-color-base-medium;
                 border-radius: calculateRem(4px);
@@ -59,8 +59,8 @@
                 display: flex;
                 justify-content: center;
                 align-items: center;
-                flex: 0 0 34px;
-                margin: 0;
+                flex: 0 0 calculateRem(34px);
+                margin: calculateRem(2px);
                 width: calculateRem(34px);
                 height: calculateRem(34px);
                 padding: 0;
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/public/scss/_preview.scss
```diff
@@ -54,14 +56,25 @@
...
             .ez-preview__action {
-                line-height: 0;
-                display: block;
-                padding: .5rem;
-                border: none;
-                border-radius: initial;
+                display: flex;
+                justify-content: center;
+                align-items: center;
+                flex: 0 0 34px;
```
sunpietro: Maybe `flex: 0 0 calculateRem(34px)`?

src/bundle/Resources/public/scss/_preview.scss
```diff
@@ -76,32 +89,31 @@
...
             margin: 0 auto;
-            transition: all .2s $ez-admin-transition;
+            transition: all 0.2s $ez-admin-transition;
         }
 
         &--tablet {
             background: url('/bundles/ezplatformadminui/img/preview-tablet.png') top center no-repeat;
-            height: 1250px;
-            margin-top: 1rem;
+            height: calculateRem(1250px);
```
sunpietro: I'm wondering, do we need such big value for `height`? I mean, it will cause the scroll.

inakijv: Hi @sunpietro, I'd say yes if we want to show the full `preview-tablet.png` img which is preview's objective. In any case given that you are pointing this out `preview-tablet.png` height is 1287px, hence maybe the correct height would be 1290px.

inakijv:
Updated based on review, and added fixup for responsiveness in order to avoid behavior of an enlarged switcher when resizing browser window (see screenshot of behavior detected and corrected)
screen shot 2018-11-26 at 11 07 56 am

inakijv:
Ping @sunpietro

PR #728, opened 2018-11-30
EZP-29848: Make existing AdminUI tabs extendable by allowing template path and parameters change

webhdx:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29848
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | yes

1. Introduced new abstract class `EventDispatchingAbstractTab` which dispatches `TabEvents::ON_RENDER` event before rendering tab. This gives a flexibility of changing template and accessing parameters to modify them.
2. I've added some useful twig blocks to Versions tab allowing to add custom columns and tables from other bundles. It's being used by upcoming Workflow bundle. 

#### TODO
- [x] Implement `EventDispatchingAbstractTab` in other tabs after validating idea on code review

### DIFF
--- a/src/lib/Tab/Event/TabEvents.php
+++ b/src/lib/Tab/Event/TabEvents.php
@@ -26,7 +26,7 @@ class TabEvents
     const TAB_PRE_RENDER = 'ezplatform.tab.pre_render';
 
     /**
-     * Is dispatched on tabs extending EventDispatchingAbstractTab.
+     * Is dispatched on tabs extending AbstractEventDispatchingTab.
      *
      * Allows to manipulate template path and parameters before rendering by Twig.
      */
--- a/src/lib/Tab/Event/TabEvents.php
+++ b/src/lib/Tab/Event/TabEvents.php
@@ -24,4 +24,11 @@ class TabEvents
      * Happens just before rendering tab.
      */
     const TAB_PRE_RENDER = 'ezplatform.tab.pre_render';
+
+    /**
+     * Is dispatched on tabs extending EventDispatchingAbstractTab.
+     *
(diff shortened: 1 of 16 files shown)


src/lib/Tab/Event/TabViewRenderEvent.php
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public function getTabIdentifier(): string
+    {
+        return $this->tabIdentifier;
+    }
+
+    /**
+     * @param string $tabIdentifier
+     */
+    public function setTabIdentifier(string $tabIdentifier): void
```
mikadamczyk: Should we change a tab identifier after creation?

src/lib/Tab/Event/TabEvents.php
```diff
@@ -24,4 +24,11 @@ class TabEvents
      * Happens just before rendering tab.
      */
     const TAB_PRE_RENDER = 'ezplatform.tab.pre_render';
+
+    /**
+     * Is dispatched on tabs extending EventDispatchingAbstractTab.
+     *
+     * Allows to manipulate template path and parameters before rendering by Twig.
```
mikadamczyk: If we allow changing the identifier than the comment should be changed.

src/lib/Tab/Event/TabViewRenderEvent.php
webhdx: Good catch, no we should not.

lserwatka:
Note for QA: we only need sanities here. cc @micszo

PR #731, opened 2018-12-02
EZP-29122: As a developer, I want to configure password policies in ezuser

adamwojs:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29122
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Tests pass?   | no
| Doc needed?   | no

:exclamation:  **Depends on https://github.com/ezsystems/ezpublish-kernel/pull/2482 and https://github.com/ezsystems/repository-forms/pull/262** 

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/admin/content_type/field_types.html.twig
+++ b/src/bundle/Resources/views/admin/content_type/field_types.html.twig
@@ -256,23 +256,25 @@
 {% endblock %}
 
 {% block ezuser_field_definition_edit %}
-    <div class="ezuser-settings {% if group_class is not empty %} {{ group_class }}{% endif %}">
-        {{- form_row(form.requireAtLeastOneUpperCaseCharacter, {'label_attr': {'class': 'checkbox-inline ez-label'}}) -}}
-    </div>
+    <div class="ez-field ez-field--user ez-user">
+        <div class="ez-user__settings {% if group_class is not empty %} {{ group_class }}{% endif %}">
+            {{- form_row(form.requireAtLeastOneUpperCaseCharacter, {'label_attr': {'class': 'checkbox-inline ez-label'}}) -}}
+        </div>
 
-    <div class="ezuser-settings {% if group_class is not empty %} {{ group_class }}{% endif %}">
-        {{- form_row(form.requireAtLeastOneLowerCaseCharacter, {'label_attr': {'class': 'checkbox-inline ez-label'}}) -}}
-    </div>
+        <div class="ez-user__settings {% if group_class is not empty %} {{ group_class }}{% endif %}">
+            {{- form_row(form.requireAtLeastOneLowerCaseCharacter, {'label_attr': {'class': 'checkbox-inline ez-label'}}) -}}
+        </div>
 
-    <div class="ezuser-settings {% if group_class is not empty %} {{ group_class }}{% endif %}">
-        {{- form_row(form.requireAtLeastOneNumericCharacter, {'label_attr': {'class': 'checkbox-inline ez-label'}}) -}}
-    </div>
+        <div class="ez-user__settings {% if group_class is not empty %} {{ group_class }}{% endif %}">
+            {{- form_row(form.requireAtLeastOneNumericCharacter, {'label_attr': {'class': 'checkbox-inline ez-label'}}) -}}
+        </div>
 
-    <div class="ezuser-settings {% if group_class is not empty %} {{ group_class }}{% endif %}">
-        {{- form_row(form.requireAtLeastOneNonAlphanumericCharacter, {'label_attr': {'class': 'checkbox-inline ez-label'}}) -}}
-    </div>
+        <div class="ez-user__settings {% if group_class is not empty %} {{ group_class }}{% endif %}">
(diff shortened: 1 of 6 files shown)


src/bundle/Resources/views/admin/content_type/field_types.html.twig
```diff
@@ -254,3 +254,25 @@
         {{- form_row(form.textRows, {'label_attr': {'class': 'ez-label'}}) -}}
     </div>
 {% endblock %}
+
+{% block ezuser_field_definition_edit %}
+    <div class="ezuser-settings {% if group_class is not empty %} {{ group_class }}{% endif %}">
```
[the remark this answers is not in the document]
adamwojs: I introduced this class based on convention which follow other field types settings blocks defined in this file.

adamwojs:
PR is ready for code review.

src/bundle/Resources/views/admin/content_type/field_types.html.twig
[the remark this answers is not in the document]
sunpietro: My proposal is: `ez-field ez-field--user ez-user` for field type container. Then here: `ez-user__settings`, what do you think about that @adamwojs @dew326 ?

src/lib/Form/Type/User/UserPasswordChangeType.php
```diff
@@ -8,18 +8,40 @@
...
+    public function __construct(ContentTypeService $contentTypeService)
+    {
+        $this->contentTypeService = $contentTypeService;
+    }
+
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
+        $contentType = null;
+        if ($options['user'] instanceof User) {
+            // TODO: Refactor to use $user->getContentType() when https://jira.ez.no/browse/EZP-29613 will be merged
```
ViniTou: It is already merged, so ;)

src/lib/Form/Factory/FormFactory.php
```diff
@@ -1018,18 +1023,22 @@ public function forgotUserPasswordWithLogin(
...
      *
      * @return \Symfony\Component\Form\FormInterface
      *
      * @throws \Symfony\Component\OptionsResolver\Exception\InvalidOptionsException
      */
     public function resetUserPassword(
         UserPasswordResetData $data = null,
-        ?string $name = null
+        ?string $name = null,
+        ?User $user = null
```
ViniTou: I think we are passing additional arguments as options, but I am not going to force this one.

src/lib/Form/Factory/FormFactory.php
```diff
@@ -1018,18 +1023,22 @@ public function forgotUserPasswordWithLogin(
     /**
      * @param \EzSystems\EzPlatformAdminUi\Form\Data\User\UserPasswordResetData $data
      * @param null|string $name
+     * @param null|User $user
```
ViniTou: User|null

webhdx: FQCN

src/lib/Form/Factory/FormFactory.php
```diff
@@ -1018,18 +1023,22 @@ public function forgotUserPasswordWithLogin(
     /**
      * @param \EzSystems\EzPlatformAdminUi\Form\Data\User\UserPasswordResetData $data
      * @param null|string $name
```
webhdx: You might as well fix this: 
```suggestion
      * @param string|null $name
```

adamwojs:
PR rebased and updated according to suggestions.

src/bundle/Resources/views/admin/content_type/field_types.html.twig
```diff
@@ -254,3 +254,27 @@
...
     </div>
 {% endblock %}
+
+{% block ezuser_field_definition_edit %}
+    <div class="ez-field ez-field--user ez-user">
+        <div class="ez-user__settings {% if group_class is not empty %} {{ group_class }}{% endif %}">
+            {{- form_row(form.requireAtLeastOneUpperCaseCharacter, {'label_attr': {'class': 'checkbox-inline ez-label'}}) -}}
+        </div>
+
+        <div class="ez-user__settings {% if group_class is not empty %} {{ group_class }}{% endif %}">
```
sunpietro: `ez-user__constraint` here and in all other places.

barbaragr:
**Password rules checkboxes are always checked**
I've created new Content Type and I've unchecked all the checkboxes, but after saving they are still checked. The same for User Content type  - they are checked by default, but it doesn't work - you can set any password. If you uncheck them, after saving they are still checked.

Some tests are blocked by: https://jira.ez.no/browse/EZP-29871

barbaragr:
1. I've created a user when User content type has all the rules unchecked (password: `p`)
2. I edited User content type by marking as checked `upper case letter` and setting minimum length to `2`
3. I went back to my user from step 1 and clicked `Edit` - password fields are empty, I changed nothin, just clicked `Update`
**Question:** It was saved without any validation. Should it be validated now?
4. I clicked `Edit` once again and typed `a` in password fields. 
5. I clicked `Update` and then Username and passwords disappeared and password field is validated. But I'm not convinced that username should also be empty.

PR #732, opened 2018-12-03
EZP-29854: Cannot remove a selected item from custom dropdown

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29854
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

## FOR QA
It's hardly testable in eZ Platform. Requires customizing the template of a field that implements custom dropdown feature.

### DIFF
--- a/src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
+++ b/src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
@@ -1,12 +1,14 @@
 (function(global, doc) {
     const CLASS_CUSTOM_DROPDOWN = 'ez-custom-dropdown';
+    const CLASS_CUSTOM_DROPDOWN_ITEM = 'ez-custom-dropdown__item';
     const CLASS_ITEMS_HIDDEN = 'ez-custom-dropdown__items--hidden';
     const CLASS_REMOVE_SELECTION = 'ez-custom-dropdown__remove-selection';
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
```diff
@@ -52,22 +55,14 @@
...
-        deselectPlaceholderOption() {
-            const placeholderItem = this.itemsContainer.querySelector('[data-value=""]');
-
-            if (!placeholderItem) {
-                return;
-            }
+            doc.body.removeEventListener('click', this.onClickOutside);
 
-            placeholderItem.selected = false;
+            return this.itemsContainer.classList.add(CLASS_ITEMS_HIDDEN);
```
tischsoic: Is there a reason behind returning this? I checked what value is returned by `classList.add` and it returns `undefined`.

lserwatka: @sunpietro is this something you will address before merge?

sunpietro: One way or another, the function will return `undefined`. I see no risk here.

PR #735, opened 2018-12-04
EZP-29857: Unable to upload audio to Media field type

janit:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29857
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass? | yes
| Doc needed?   | no

Adds conditional selecting of MIME type for Media field type in Twig template. When HTML5 Audio (html5_audio) is selected, accepted MIME type is set to `audio/*`, in all other cases it is set to `video/*`. Note that this does not change behaviour for Flash, SilverLight, and QuickTime types.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/fieldtypes/edit/ezmedia.html.twig
+++ b/src/bundle/Resources/views/fieldtypes/edit/ezmedia.html.twig
@@ -5,7 +5,11 @@
 {%- block ezplatform_fieldtype_ezmedia_row -%}
     {% set preview_block_name = 'ezmedia_preview' %}
     {% set max_file_size = min(form.parent.vars.value.fieldDefinition.validatorConfiguration.FileSizeValidator.maxFileSize * 1024 * 1024, max_upload_size)|round %}
-    {% set attr = attr|merge({'accept': 'video/*'}) %}
+    {% if form.parent.vars.value.fieldDefinition.fieldSettings.mediaType == "html5_audio" %}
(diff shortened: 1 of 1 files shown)


barbaragr:
@janit rebase is needed here, please.

janit:
@barbaragr This should be done now. Please verify.

janit:
@barbaragr: 

1. I think MIME type `video/mp4` should not be accepted to be uploaded at all when type is set to HTML 5 audio. MP4 audio (`audio/mp4`) should be accepted. If there is exceptions between the two. Which browser did you use?

2. The template used does not seem to result in any preview for audio files. However, the files is stored as expected. Please observe video here: https://www.youtube.com/watch?v=NixHzM_RHVU

ViniTou:
@janit 
Is there anyway this could be solved in corresponding MediaType and mediaFormMapper instead of making this in a view?

janit:
@ViniTou The view is definitely not an optimal location for this. This fixed an immediate issue where audio uploads did not work.

In the long term MIME types should be mapped somewhere else, but I'm not 100% where this would be best done. I can look into it given some insight from someone with a broader understanding of where this is best done.

andrerom:
Seems we can close this then, thanks @ViniTou! 👍

PR #747, opened 2018-12-13
EZP-29892: Error messages keep adding next to existing one after field validation

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29892
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes/no
| Doc needed?   | yes/no

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/base/base-field.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/base/base-field.js
@@ -1,5 +1,6 @@
 (function(global, doc) {
     const eZ = (global.eZ = global.eZ || {});
+    const SELECTOR_ERROR_CONTAINER = '.ez-field-edit__error';
 
     eZ.BaseFieldValidator = class BaseFieldValidator {
         constructor(config) {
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/scripts/fieldType/base/base-field.js
```diff
@@ -164,6 +165,7 @@
                 nodes.forEach((el) => {
                     const errorNode = this.createErrorNode(validationResult.errorMessage);
 
+                    el.querySelectorAll(SELECTOR_ERROR_CONTAINER).forEach((el) => el.remove());
```
dew326: This feels like redundant with `existingErrorNodes.forEach((el) => el.remove());`.

In this author field, the problem is that method `findExistingErrorNodes` is not returning all of the error nodes.

lserwatka:
@sunpietro are you still on this one?

sunpietro:
It was fixed here: https://github.com/ezsystems/ezplatform-admin-ui/pull/753

PR #739, opened 2018-12-07
EZP-29862: Trash button is disabled after using browser back button

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29862
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no
| License       | [GPL-2.0](https://github.com/ezsystems/ezplatform-admin-ui/blob/master/LICENSE)

After using "back button" in a browser, browser sets inputs values to values entered by a user before. 
Thus, we cannot assume that all checkboxes/radios are unchecked at the beginning, therefore we need to check at the start their values and set disabled state accordingly.

**For QA**: fix will affect all tables with checkboxes and Trash buttons + it will affect trash modal for content item with Immage Asset field.

#### Checklist:
- [x] Coding standards (`$ composer fix-cs`)
- [x] Ready for Code Review

PR #748, opened 2018-12-13
EZP-29889: Cannot set property 'disabled' of null error in console in Trash

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29889
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

In this fix: https://github.com/ezsystems/ezplatform-admin-ui/pull/739 I added functions executions, which set _disabled_ status of buttons just after page has been loaded. Previous code assumed that buttons are always present, but in case of Trash, when there are no items, there are no buttons:

screen shot 2018-12-13 at 15 25 30

it seems that the problem does not exist in other places because buttons are always present, e.g. in Bookmarks or always at least one item is present or table is not displayed when there are no items (e.g. _Draft under edit_ table):

screen shot 2018-12-13 at 14 48 32

but I added `if`s also in other places to make code more bullet-proof. 

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.trash.list.js
+++ b/src/bundle/Resources/public/js/scripts/admin.trash.list.js
@@ -55,9 +55,17 @@
         const isEmptySelection = checkboxes.every((el) => !el.checked);
         const isMissingParent = checkboxes.some((el) => el.checked && parseInt(el.dataset.isParentInTrash, 10) === 1);
 
-        buttonRestore.disabled = isEmptySelection || isMissingParent;
-        buttonDelete.disabled = isEmptySelection;
-        buttonRestoreUnderNewParent.disabled = isEmptySelection;
+        if (buttonRestore) {
+            buttonRestore.disabled = isEmptySelection || isMissingParent;
+        }
+
+        if (buttonDelete) {
+            buttonDelete.disabled = isEmptySelection;
+        }
+
+        if (buttonRestoreUnderNewParent) {
+            buttonRestoreUnderNewParent.disabled = isEmptySelection;
+        }
     };
     const updateTrashForm = (checkboxes) => {
         checkboxes.forEach((checkbox) => {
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/public/js/scripts/admin.trash.list.js
```diff
@@ -55,9 +55,17 @@
...
 
-        buttonRestore.disabled = isEmptySelection || isMissingParent;
-        buttonDelete.disabled = isEmptySelection;
-        buttonRestoreUnderNewParent.disabled = isEmptySelection;
+        if (buttonRestore) {
+            buttonRestore.disabled = isEmptySelection || isMissingParent;
+        }
+
+        if (buttonRestore) {
+            buttonDelete.disabled = isEmptySelection;
```
sunpietro: Can't it be in one `if`?

tischsoic: Actually, I wanted to ask about it. I made 3 `if`s because I didn't want to assume that all buttons are present or none. If I change it to `if (buttonRestore && buttonDelete && buttonRestoreUnderNewParent) {` then if only one is not present, the other 2 wont work properly. 

What do you thing about it?

tischsoic:
I was also wondering, maybe we should display all 3 buttons in _Trash_, when no items are present, but in inactive state, in order to be consisten with the rest of the UI, what do you think - ping @sunpietro @dew326 @micszo ?

micszo:
Personally I like the idea.

src/bundle/Resources/public/js/scripts/admin.trash.list.js
sunpietro: When I opened it for review, there was: `if (buttonRestore) {` three times. Now it's fixed and it makes sense.

tischsoic: Oh, you're right, I have realised that I made slip and I removed `Ready for review` label but I was too slow. 😄

micszo:
@webhdx when trying to reload cache on EE clean master with this branch I get this error:
screenshot 2018-12-14 at 08 49 19
Please advise.

webhdx:
@tischsoic Is this branch rebased on current `master`?

tischsoic:
@webhdx But it's bug fix - AFAIK it must go to 1.3.

micszo:
OK, I will recheck it later on master.

lserwatka:
@tischsoic you can merge it up

tischsoic:
merged up: https://github.com/ezsystems/ezplatform-admin-ui/commit/b4c645ef99067abd18b9591b5a3c47c75f78f24e
ping @micszo - you can check it on _master_ :)

PR #754, opened 2018-12-17
EZP-29871: Error 500 after creating Content Item based on Content Type with ezuser

adamwojs:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29871
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

IMHO we can assume that if a content type contains `ezuser` field then it is "User" content type (and skip adding it to YAML configuration).

#### Checklist:
- [X] Ready for Code Review


lserwatka:
@adamwojs same here you will have to re-base with 1.3 now to make PHPUnit pass again.

PR #758, opened 2018-12-17
EZP-29875: Facebook tag set focus issue

dew326:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29875
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review


src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-customtag-update.js
```diff
@@ -182,7 +184,7 @@ export default class EzBtnCustomTagUpdate extends EzWidgetButton {
             this.execCommand();
         }
 
-        const widget = this.getWidget();
+        const widget = this.getWidget() || this.widget;
```
sunpietro: Can you tell me more about it? You're setting `this.widget` value in the constructor. When the output of `this.getWidget()` will change its value?

dew326: The output of `this.getWidget()` will be null when we will lose focus in edit custom tag. This happens when we are editing custom tag and use cmd+a to select text in the input.

lserwatka:
@micszo I merged it by mistake. Please report followups.

micszo:
It works fine. 🎉

PR #761, opened 2018-12-18
EZP-29919: Add 'approved' and 'rejected' icons

webhdx:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29919
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

image

#### Checklist:
- [x] Ready for Code Review


src/bundle/Resources/public/img/ez-icons.svg
```diff
@@ -638,4 +638,10 @@
...
 </symbol>
+<symbol id="approved" viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg">
+    <path d="M16.177 22c-.357 0-.681-.141-.92-.37L7.08 13.817a1.333 1.333 0 1 1 1.84-1.926l7.23 6.907L29.717 5.065a1.334 1.334 0 1 1 1.9 1.874l-14.49 14.667a1.33 1.33 0 0 1-.949.397h-.001zM32 16c0-1.289-.151-2.542-.435-3.743l.022.11a1.333 1.333 0 1 0-2.585.61l-.001-.007c.224.915.353 1.966.353 3.047 0 7.364-5.97 13.333-13.333 13.333S2.688 23.38 2.688 16.017c0-7.364 5.97-13.333 13.333-13.333a13.28 13.28 0 0 1 8.493 3.054l-.022-.018a1.333 1.333 0 0 0 1.702-2.051l-.003-.002A15.907 15.907 0 0 0 16.001.002c-8.837 0-16 7.163-16 16s7.163 16 16 16 16-7.163 16-16V16z"></path>
+</symbol>
+<symbol id="rejected" viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg">
```
sunpietro: I would rename it to `workflow-rejected`

webhdx: It was named `rejected` on purpose because AdminUI should not have dependency on Workflow at all. It's weird to me that we store icons used by other packages in AdminUI repository.

Naming it `rejected` and `approved` makes them general purpose icons.

sunpietro: Yes, but if you look closer to the icon itself, then you'll notice that it says nothing about being rejected. It's more like revert

PR #763, opened 2018-12-18
EZP-29921: Add possibility to use icons with custom path in Sidebar menu

webhdx:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29921
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | yes

1. Introduces new `icon_path` parameter in menu item extras to be used instead of existing `icon` parameter if menu items need custom svg icons.
2. `icon_class` was added for adding custom classes to `<svg></svg>` element.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/parts/menu/sidebar_base.html.twig
+++ b/src/bundle/Resources/views/parts/menu/sidebar_base.html.twig
@@ -36,7 +36,7 @@
 {% block label %}
     {% set icon_path = null %}
     {% set icon_class = (item_class|default('ez-icon ') ~ item.extras.icon_class|default(''))|trim %}
-    {% if item.extras.icon_path is defined and item.extras.icon_path is not empty %}
+    {% if item.extras.icon_path|default %}
         {% set icon_path = item.extras.icon_path %}
     {% elseif item.extras.icon is defined and item.extras.icon is not empty %}
         {% set icon_path = asset('bundles/ezplatformadminui/img/ez-icons.svg') ~ '#' ~ item.extras.icon %}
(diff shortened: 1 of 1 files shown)


webhdx:
Ping @DominikaK. We need to mention it in the doc.

src/bundle/Resources/views/parts/menu/sidebar_base.html.twig
```diff
@@ -34,9 +34,18 @@
 {% endblock %}
 
 {% block label %}
-    {% if item.extras.icon is defined and item.extras.icon != '' %}
-        <svg class="ez-icon ez-icon-{{ item.extras.icon }}">
-            <use xlink:href="{{ asset('bundles/ezplatformadminui/img/ez-icons.svg') }}#{{ item.extras.icon }}"></use>
+    {% set icon_path = null %}
+    {% set icon_class = (item_class|default('ez-icon ') ~ item.extras.icon_class|default(''))|trim %}
+    {% if item.extras.icon_path is defined and item.extras.icon_path is not empty %}
```
mikadamczyk:
```suggestion
    {% if item.extras.icon_path|default %}
```

webhdx: @mikadamczyk Are you sure it will work? Doesn't seem that obvious to me.

mikadamczyk: I tested this in twig and it works for me

PR #778, opened 2018-12-21
EZP-29821: Reduce API/SPI lookups in Admin UI

andrerom:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29821
| Improvment     | yes
| BC breaks?    | no
| Tests pass?   | ??
| Doc needed?   | no

##### Before:
dashboard_before

##### After:
dashboard_after

##### Done:
- Take advantage of `isUser[Group]()` API's added in 2.4
- Avoid loading language map twice
- Deprecate content type config to just use one
  - Possible todo or followup: Find a way to share the data between the two, could be done on JS side too, once that is done ContentTypeNames can be removed

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/UI/Config/Provider/ContentTypes.php
+++ b/src/lib/UI/Config/Provider/ContentTypes.php
@@ -33,7 +33,7 @@ class ContentTypes implements ProviderInterface
      */
     public function getConfig()
     {
-        $contentTypeGroups = ['_types' => []];
+        $contentTypeGroups = [];
 
         $preferredLanguages = $this->userLanguagePreferenceProvider->getPreferredLanguages();
         $loadedContentTypeGroups = $this->contentTypeService->loadContentTypeGroups(
@@ -49,7 +49,6 @@ class ContentTypes implements ProviderInterface
                     'identifier' => $contentType->identifier,
                     'name' => $contentType->getName(),
                 ];
-                $contentTypeGroups['_types'][$contentType->identifier] = $contentType->getName();
             }
         }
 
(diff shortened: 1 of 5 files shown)


src/lib/UI/Config/Provider/ContentTypes.php
```diff
@@ -27,14 +27,15 @@ public function __construct(ContentTypeService $contentTypeService)
...
-        $contentTypeGroups = [];
+        $contentTypeGroups = ['_types' => []];
 
         foreach ($this->contentTypeService->loadContentTypeGroups() as $contentTypeGroup) {
             foreach ($this->contentTypeService->loadContentTypes($contentTypeGroup) as $contentType) {
                 $contentTypeGroups[$contentTypeGroup->identifier][] = [
                     'identifier' => $contentType->identifier,
                     'name' => $contentType->getName(),
                 ];
+                $contentTypeGroups['_types'][$contentType->identifier] = $contentType->getName();
```
andrerom: Note: Not strictly needed, it could be extracted on JS side instead if we wanted to, depends on where you want to take this later.

andrerom:
Thanks @katarzynazawada, reproduced, and fixed in 33056e0

andrerom:
@mnocon Failures seems unrelated and known, ok for merge?
https://res.cloudinary.com/ezplatformtravis/image/upload/v1547825242/screenshots/vendor_ezsystems_ezplatform-admin-ui_features_contenttypefields_feature_26_esyahm.png

mnocon:
@andrerom yes, this error should be fixed already

I've restarted the jobs - can we wait a bit to make sure they're passing?

andrerom:
@mnocon Sure

lserwatka:
All green, so merging.

PR #786, opened 2019-01-03
ezdatetime displays selected date and timeon calendar after clearing the field

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29783
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Fix clearing of date/datetime in ezdate & ezdatetime. Clear should clear both input element and date selected in calendar view.

Most of the changes were caused by formatting files with Prettier + changes in IIFE parameters.
The Problem was fixed by changing:
```js
flatPickrInput.value = '';	                 
sourceInput.value = '';
```
to
```js
flatPickrInput._flatpickr.clear();
```
in both files.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
@@ -1,11 +1,11 @@
-(function (global) {
+(function(global, doc, eZ, flatpickr) {
     const SELECTOR_FIELD = '.ez-field-edit--ezdate';
     const SELECTOR_INPUT = '.ez-data-source__input:not(.flatpickr-input)';
     const SELECTOR_LABEL_WRAPPER = '.ez-field-edit__label-wrapper';
     const SELECTOR_FLATPICKR_INPUT = '.flatpickr-input';
     const EVENT_VALUE_CHANGED = 'valueChanged';
 
-    class EzDateValidator extends global.eZ.BaseFieldValidator {
+    class EzDateValidator extends eZ.BaseFieldValidator {
         /**
          * Validates the input
          *
@@ -24,12 +24,12 @@
 
             if (isRequired && isEmpty) {
                 isError = true;
-                errorMessage = window.eZ.errors.emptyField.replace('{fieldName}', label);
+                errorMessage = eZ.errors.emptyField.replace('{fieldName}', label);
             }
 
             return {
                 isError,
-                errorMessage
+                errorMessage,
             };
         }
     }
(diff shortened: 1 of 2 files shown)


tischsoic:
ping @sunpietro @dew326 :)

src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
```diff
@@ -85,30 +83,41 @@
...
 
         if (sourceInput.value) {
             defaultDate = new Date(sourceInput.value * 1000);
         }
 
-        btnClear.addEventListener('click', (event) => {
-            event.preventDefault();
+        btnClear.addEventListener(
+            'click',
+            (event) => {
```
dew326: Maybe it would be good to define callback in some variable?

src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
```diff
@@ -85,30 +83,41 @@
...
+                sourceInput.dispatchEvent(new CustomEvent(EVENT_VALUE_CHANGED));
+            },
+            false
+        );
 
-        window.flatpickr(flatPickrInput, Object.assign({}, dateConfig, {
-            onChange: updateInputValue.bind(null, sourceInput),
-            defaultDate
-        }));
+        flatpickrInstance = flatpickr(
```
dew326: Maybe you can move this above and use normal `const` instead of `let`?

src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
```diff
@@ -85,30 +83,41 @@
...
-        btnClear.addEventListener('click', (event) => {
-            event.preventDefault();
+        btnClear.addEventListener(
+            'click',
+            (event) => {
+                event.preventDefault();
 
-            flatPickrInput.value = '';
-            sourceInput.value = '';
+                if (!flatpickrInstance) {
```
dew326: I think with my changes the `flatpickrInstance` will never be null?

dew326:
The comments can be applied in both files.

src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
tischsoic: You're right :)

src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
tischsoic: I didn't want to change to much at first, but in last commit I refactored it. :)

tischsoic:
ping @sunpietro @dew326

PR #790, opened 2019-01-07
EZP-29811: Content field definitions is misaligned while creating Content Type

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29811
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

## Fixes button styling. After changes:
screen shot 2019-01-07 at 08 38 43

screen shot 2019-01-07 at 08 45 22

### DIFF
--- a/src/bundle/Resources/public/scss/_card.scss
+++ b/src/bundle/Resources/public/scss/_card.scss
@@ -34,15 +34,12 @@
             padding: .25rem 1.25rem;
             justify-content: space-between;
             z-index: 1040;
+        }
 
-            .form-inline {
-                .btn-danger {
-                    .ez-icon {
-                        width: 1.5rem;
-                        height: 1.5rem;
-                        margin-top: 5px;
-                    }
-                }
+        .form-inline {
+            .btn-danger {
+                display: flex;
+                height: calculateRem(48px);
             }
         }
     }
--- a/src/bundle/Resources/views/admin/content_type/create.html.twig
+++ b/src/bundle/Resources/views/admin/content_type/create.html.twig
@@ -6,7 +6,7 @@
 
 {% block _ezrepoforms_contenttype_update_removeFieldDefinition_widget %}
     <button type="{{ type|default('button') }}" {{ block('button_attributes') }}>
-        <svg class="ez-icon ez-icon--medium ez-icon--light ez-icon-trash">
+        <svg class="ez-icon ez-icon--medium ez-icon--light">
             <use xlink:href="{{ asset('bundles/ezplatformadminui/img/ez-icons.svg') }}#trash"></use>
         </svg>
     </button>
--- a/src/bundle/Resources/views/admin/content_type/create.html.twig
+++ b/src/bundle/Resources/views/admin/content_type/create.html.twig
@@ -6,7 +6,7 @@
 
 {% block _ezrepoforms_contenttype_update_removeFieldDefinition_widget %}
     <button type="{{ type|default('button') }}" {{ block('button_attributes') }}>
-        <svg class="ez-icon ez-icon-trash">
+        <svg class="ez-icon ez-icon--medium ez-icon--light ez-icon-trash">
             <use xlink:href="{{ asset('bundles/ezplatformadminui/img/ez-icons.svg') }}#trash"></use>
         </svg>
     </button>
@@ -53,7 +53,7 @@
     <section class="container mt-4 px-5">
         <div class="card ez-card">
             <div class="ez-card__header ez-card__header--secondary d-flex justify-content-between">
-                <div class="p-2">
+                <div class="p-3">
                     {{ 'content_type.view.create.content_field_definitions'|trans|desc('Content field definitions') }}
                 </div>
                 <div class="form-inline">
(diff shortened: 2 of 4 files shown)


tischsoic:
Styled using flex - ping @sunpietro @dew326

src/bundle/Resources/public/scss/_card.scss
```diff
@@ -34,14 +34,16 @@
...
+                height: calculateRem(48px);
 
-            .form-inline {
-                .btn-danger {
-                    .ez-icon {
-                        width: 1.5rem;
-                        height: 1.5rem;
-                        margin-top: 5px;
-                    }
+                .ez-icon-trash {
```
dew326: You can use class `ez-icon--medium` to set 24px height and width

src/bundle/Resources/public/scss/_card.scss
```diff
@@ -34,14 +34,16 @@
             padding: .25rem 1.25rem;
             justify-content: space-between;
             z-index: 1040;
+        }
+
+        .form-inline {
+            .btn-danger {
+                display: flex;
+                height: calculateRem(48px);
```
sunpietro: Aren't properties like `align-items: center;` and `justify-content: center` missing here, are they?

tischsoic: It works without these properties.

src/bundle/Resources/public/scss/_card.scss
tischsoic: Yes, but I have decided to add `.ez-icon-trash` instead, because in `create.html.twig` this statement:
https://github.com/ezsystems/ezplatform-admin-ui/blob/5b9dce453200d89961b0082c8ad0b747dd4f6e18/src/bundle/Resources/views/admin/content_type/create.html.twig#L64 
generates button with `.ez-icon-trash`. Also, other trash buttons in our system have `.ez-icon-trash` class. For consistency, I think we should use this class here as well.

dew326: But adding class to set width and height is redundant. For having icons with width and height we have special modifier class `ez-icon--medium`. I'm not against adding `.ez-icon-trash`, I'm against using this class to set 24px width and height.

tischsoic: You're right, better idea would be to use `.ez-icon--medium`. Also, I realised that the template for `{{ form_widget(form.removeFieldDefinition ...` is in the same file, so I can add `.ez-icon--medium` there. 
Unfortunately, after applying changes I realised that we are styling `.ez-card__header .form-inline .btn-danger .ez-icon` in `_icons.scss`:
https://github.com/ezsystems/ezplatform-admin-ui/blob/5b9dce453200d89961b0082c8ad0b747dd4f6e18/src/bundle/Resources/public/scss/_icons.scss#L76-L81
what overwrites `.ez-icon--medium` selector. Probably this is why someone before me styled it here using this selector: `.ez-card__header--sticky-top .form-inline .btn-danger .ez-icon` (I only changed `.ez-icon` -> `.ez-icon-trash` and moved out from `--sticky-top`).

So, I think that my current solution is valid in context of this task. Definitely, it would be great to refactor `.ez-card__header .form-inline .btn-danger .ez-icon` from `_icons.scss`, but I think that it should be done in another task/PR. 
What do you think, @dew326 ? 🙂

sunpietro: Maybe you should fix that selector as well and apply a proper CSS class the elements defined by `.ez-card__header .form-inline .btn-danger .ez-icon` selector?

lserwatka:
@dew326 could you make a final review here? So we can move this to QA?

lserwatka:
@tischsoic could you change target branch to 1.4?

katarzynazawada:
@tischsoic according to Affects Version base here should be 1.3. Please, do the rebase also :)

src/bundle/Resources/views/admin/content_type/create.html.twig
```diff
@@ -6,7 +6,7 @@
 
 {% block _ezrepoforms_contenttype_update_removeFieldDefinition_widget %}
     <button type="{{ type|default('button') }}" {{ block('button_attributes') }}>
-        <svg class="ez-icon ez-icon-trash">
+        <svg class="ez-icon ez-icon--medium ez-icon--light ez-icon-trash">
```
sunpietro: Is this class: `ez-icon-trash` used anywhere in CSS or JS?

tischsoic: Actually it is not used, I will remove it.

lserwatka:
@tischsoic could you merge it up?

lserwatka:
1.4 -> master

tischsoic:
Merged up:
- 1.4: https://github.com/ezsystems/ezplatform-admin-ui/commit/e32cf7ae968b6a70fe3fb363ef441c915f43c390
- master: https://github.com/ezsystems/ezplatform-admin-ui/commit/ba94b30c0b4ba86ad96d7a674092d7ba9612cede

PR #793, opened 2019-01-09
EZP-29779: [Safari] Nothing responds after using web browser back button

konradoboza:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29779
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

When the back button is used in Safari, the `onload` event isn't called and the whole UI freezes. This PR contains workaround which is just about removing `ez-prevent-click` class from the `<body>` element.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/layout.html.twig
+++ b/src/bundle/Resources/views/layout.html.twig
@@ -5,9 +5,9 @@
         <meta name="CSRF-Token" content="{{ csrf_token('authenticate') }}" />
         <meta name="SiteAccess" content="{{ app.request.get('siteaccess').name }}" />
         <script>
-            const ua = navigator.userAgent.toLowerCase();
+            const userAgent = navigator.userAgent.toLowerCase();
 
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/views/layout.html.twig
```diff
@@ -4,6 +4,17 @@
         <meta charset="UTF-8" />
         <meta name="CSRF-Token" content="{{ csrf_token('authenticate') }}" />
         <meta name="SiteAccess" content="{{ app.request.get('siteaccess').name }}" />
+        <script>
+            const ua = navigator.userAgent.toLowerCase();
```
tischsoic: In JS we usually use full variable names - `userAgent`. 😝

konradoboza:
ping @lserwatka

PR #795, opened 2019-01-11
EZP-29993: As the eZ Platform developer I would like to use Sass compiler in the Admin UI modules styles

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29993
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

### DIFF
--- a/src/bundle/Resources/public/scss/_icons.scss
+++ b/src/bundle/Resources/public/scss/_icons.scss
@@ -1,6 +1,7 @@
 .ez-icon {
     width: calculateRem(32px);
     height: calculateRem(32px);
+    transition: fill 0.3s $ez-admin-transition;
 
     &--small {
         width: calculateRem(16px);
--- a/src/bundle/Resources/public/scss/_icons.scss
+++ b/src/bundle/Resources/public/scss/_icons.scss
@@ -17,6 +17,11 @@
         height: calculateRem(24px);
     }
 
+    &--large {
+        width: calculateRem(64px);
+        height: calculateRem(64px);
+    }
+
     &--light {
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/public/scss/_icons.scss
```diff
@@ -7,11 +7,26 @@
         height: calculateRem(16px);
     }
 
+    &--small-medium {
```
tischsoic: I think that `medium-small` would be more pronounceable.

src/bundle/Resources/public/scss/_icons.scss
```diff
@@ -7,11 +7,26 @@
...
+        width: calculateRem(20px);
+        height: calculateRem(20px);
+    }
+
     &--medium {
         width: calculateRem(24px);
         height: calculateRem(24px);
     }
 
+    &--large {
```
tischsoic: I wonder if `48px` and `64px` are based on elements from the interface?

src/bundle/Resources/public/scss/_icons.scss
sunpietro: Me and @dew326 agreed to use this name

src/bundle/Resources/public/scss/_icons.scss
sunpietro: Yes, these sizes come from AdminUI modules.

tischsoic: Great!

PR #798, opened 2019-01-15
EZP-29026: As an Editor, I want to be able to assign a section to a subtree

adamwojs:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29026
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

## Screenshot

screenshot_2019-01-15 ez platform

#### Checklist:
- [ ] Permissions should be taken into account in list of sections
- [ ] Ready for Code Review


src/lib/Form/Data/Location/LocationAssignSubtreeData.php
```diff
@@ -0,0 +1,59 @@
...
+
+class LocationAssignSubtreeData
+{
+    /** @var \eZ\Publish\API\Repository\Values\Content\Section|null */
+    protected $section;
+
+    /** @var \eZ\Publish\API\Repository\Values\Content\Location|null */
+    protected $location;
+
+    public function __construct(?Section $section = null, ?Location $location = null)
```
ViniTou: When those can be null?

src/lib/Tab/LocationView/DetailsTab.php
```diff
@@ -139,10 +138,22 @@ public function getTemplateParameters(array $contextParameters = []): array
         $lastContributor = (new UserExists($this->userService))->isSatisfiedBy($versionInfo->creatorId)
             ? $this->userService->loadUser($versionInfo->creatorId) : null;
 
-        $canSeeSection = $this->permissionResolver->hasAccess('section', 'view');
+        $section = null;
+        $canSeeSection = $this->permissionResolver->hasAccess('section', 'view') !== false;
```
mikadamczyk: array is returned only if limitations are on this function, so we can not expect empty array here

src/lib/Form/Data/Location/LocationAssignSubtreeData.php
adamwojs: https://github.com/adamwojs/ezplatform-admin-ui/blob/1db795bf325b8ed65176bf12206f49239ebb7942/src/bundle/Controller/LocationController.php#L616

adamwojs:
PR updated according to @ViniTou @mikadamczyk code review suggestions.

src/lib/Form/Data/Location/LocationAssignSubtreeData.php
ViniTou: I understand that at point of creating, but what would happend if you pass null explicite into this, and then handle form? Shouldn't this have also have some nonEmpty validation?

adamwojs:
PR updated with `EzSystems\EzPlatformAdminUi\Tab\LocationView\DetailsTab::getTemplateParameters` method refactoring (offline discussion with @mikadamczyk)

adamwojs:
PR updated according to Slack comment: replaced `array &parameters` with `ArrayObject $parameters` in `supply*` methods.

adamwojs:
PR updated according to QA results.

barbaragr:
Issues to think about: 
1. Do we want this feature in Details tab? Maybe it will be more consistent, if user can assign a subtree in Admin/Sections like it is with single Content Item? Or if it should be in Details tab maybe we can change the title or button's name. I was testing the feature and I knew that I will assign the whole subtree because of PR's title, but I'm not sure if it is obvious enough.  
2. Error message can be more general. Now it looks like this:
screen shot 2019-01-25 at 14 49 02
no matter if user has Content Type or Owner or other limitation. It can be confusing. Maybe something like: action is not permitted or user doesn't have permissions would be better?
ping: @SylvainGuittard @inakijv

adamwojs:
AD 2) @SylvainGuittard only binary information about lack of permissions and input parameters: section and location 😉

barbaragr:
I've reported https://jira.ez.no/browse/EZP-30056 - assigning the whole subtree to the section. I will report also ticket with notification issue when it will be known how it should look like.

PR #799, opened 2019-01-15
EZP-29849: Allow user to remove location in subtree/location limitation (content/create)

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29849
| Bug fix?      | yes/no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no
| License       | [GPL-2.0](https://github.com/ezsystems/ezplatform-admin-ui/blob/master/LICENSE)

## New UI

screen shot 2019-01-15 at 15 15 54

screen shot 2019-01-15 at 15 16 46

screen shot 2019-01-15 at 15 16 18

### With multiple items

screen shot 2019-01-15 at 15 18 48

## ez-tag

This PR introduces new CSS Block - _ez-tag_, which is shared between location subtree limitation and subtree from Search filters.
Refactoring of subtree from Search filters was made in separate commit.

PR #804, opened 2019-01-17
EZP-29998: As a developer, I want to define icons for content types

adamwojs:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29998
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | yes  

## Example configuration 

```yml
ezpublish:
    system:
       default:
           content_type:
              article:
                 thumbnail: '/assets/images/customarticle.svg'
              poll:
                 thumbnail: '/assets/images/poll.svg'
```

## Using content type icons in Twig templates

Content type icons are accessible in Twig templates via `ez_content_type_icon` function which requires content type or content type identifier as an argument and returns path to content type icon.   

```twig
<svg class="ez-icon ez-icon-{{ contentType.identifier }}">
    <use xlink:href="{{ ez_content_type_icon(contentType.identifier) }}"></use>
</svg>
```

If the icon for given content type wasn't specified in the configuration then path default icon will be returned.

## Default content type icon 

A configuration of the default icon for content types is possible via `default-config` key. For example:  

```yml
ezpublish:
    system:
       default:
           content_type:
              default-config:
                 thumbnail: '/assets/images/mydefaulticon.svg'
```

## Using content type icons in JS

Icons confgiguration is pushed from the backend as as part of the global configuration `eZ.adminUiConfig.contentTypes`.

However, more convenient way of retrieving icon would be to use new helper function `getContentTypeIcon`, which takes content type identifier as argument. It belongs to global object `eZ.helpers.contentType`.

Example with _getContentTypeIcon_:
```jsx
const contentTypeIcon = eZ.helpers.contentType.getContentTypeIcon(contentTypeIdentifier);

return (
    <svg className="ez-icon">
        <use xlinkHref={contentTypeIcon} />
    </svg>
);
```

#### Checklist:
- [x] Ready for Code Review


lserwatka:
We might need to rethink naming for this function `ezplatform_admin_ui_content_type_icon`

I woukd suggest something like `ez_render_icon`, `ez_ui_content_type_icon`, or `ez_content_type_icon`

src/lib/UI/Service/ContentTypeIconResolver.php
```diff
@@ -0,0 +1,56 @@
...
+    public function getContentTypeIcon($contentType): ?string
+    {
+        if ($contentType instanceof ContentType) {
+            $contentType = $contentType->identifier;
+        }
+
+        $config = $this->configResolver->getParameter('content_type');
+
+        if (isset($config[$contentType]) && !empty($config[$contentType]['thumbnail'])) {
+            return $config[$contentType]['thumbnail'];
```
adamwojs: Note to myself: this should be passed through asset component, otherwise URL will be incorrect when it contains  `#`. Ref. https://github.com/ezsystems/ezplatform-admin-ui/pull/799#discussion_r248653338

adamwojs:
PR updated with the following changes:
* Renamed twig helper function  to `ez_content_type_icon` according to @lserwatka suggestion
* `ez_content_type_icon` returns url decorated by assets component
* Fixed configuration overwrite bug

src/bundle/Resources/views/layout.html.twig
```diff
@@ -162,6 +162,7 @@
             '@EzPlatformAdminUiBundle/Resources/public/js/scripts/helpers/request.helper.js'
             '@EzPlatformAdminUiBundle/Resources/public/js/scripts/helpers/notification.helper.js'
             '@EzPlatformAdminUiBundle/Resources/public/js/scripts/helpers/timezone.helper.js'
+            '@EzPlatformAdminUiBundle/Resources/public/js/scripts/helpers/content.type.icon.helper.js'
```
dew326: I would name this just `content.type.helper.js` because in the `addConfig()` the group for those methods is `helpers.contentType`, which implies there will be different methods for content types not only of the icon.

src/lib/UI/Service/ContentTypeIconResolver.php
```diff
@@ -0,0 +1,84 @@
...
+    /** @var \Symfony\Component\Asset\Packages */
+    private $packages;
+
+    /** @var string|null */
+    private $defaultThumbnail;
+
+    /**
+     * @param \eZ\Publish\Core\MVC\ConfigResolverInterface $configResolver
+     * @param \Symfony\Component\Asset\Packages $packages
+     * @param string|null $defaultThumbnail
```
ViniTou: doc does not match constructor parameter.

Similiar tihing seems to be in `getContentTypeIcon` method

src/lib/UI/Service/ContentTypeIconResolver.php
```diff
@@ -0,0 +1,84 @@
...
+     *
+     * @return string|null
+     */
+    public function getContentTypeIcon($contentType): ?string
+    {
+        if ($contentType instanceof ContentType) {
+            $contentType = $contentType->identifier;
+        }
+
+        $thumbnail = null;
```
dew326: Here could be `$thumbnail = $this->defaultThumbnail;` and the `if empty($thumbnail)` is not needed.

src/lib/UI/Service/ContentTypeIconResolver.php
```diff
@@ -0,0 +1,84 @@
...
+     * entry for given content type, then path to default icon will be returned.
+     *
+     * @param \eZ\Publish\API\Repository\Values\ContentType\ContentType|string $contentType
+     *
+     * @return string|null
+     */
+    public function getContentTypeIcon($contentType): ?string
+    {
+        if ($contentType instanceof ContentType) {
+            $contentType = $contentType->identifier;
```
dew326: This looks odd one time under the `$contentType` is ContentType object and once is the identifier. The ContentType seems to be not needed here so `getContentTypeIcon()` shouldn't take just string with an identifier?

src/bundle/Resources/views/layout.html.twig
tischsoic: I was thinking about the same and I agree.

src/bundle/Resources/config/services.yml
```diff
@@ -21,6 +21,10 @@ imports:
     - { resource: services/permissions.yml }
     - { resource: services/forms.yml }
 
+parameters:
+    # default content type icon
+    ezsystems.ezplatform_admin_ui.default_content_type_thumbnail: '/bundles/ezplatformadminui/img/ez-icons.svg#file'
```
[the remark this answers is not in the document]
tischsoic: It was specified in the task (https://jira.ez.no/browse/EZP-29998) description:
> In case of a custom content type without defined custom icons, then the file icon from admin-ui will be used.

adamwojs:
PR updated according to @ViniTou and @dew326 suggestions.

src/lib/UI/Service/ContentTypeIconResolver.php
adamwojs: Not really. If you configure null or empty string as the content type icon then default icon should be used.

src/lib/UI/Service/ContentTypeIconResolver.php
adamwojs: Fixed in https://github.com/ezsystems/ezplatform-admin-ui/pull/804/commits/c10c434f626e0ad965d58656ad8f971e051ed219

src/lib/UI/Service/ContentTypeIconResolver.php
adamwojs: Fixed in https://github.com/ezsystems/ezplatform-admin-ui/pull/804/commits/c10c434f626e0ad965d58656ad8f971e051ed219

src/lib/UI/Service/ContentTypeIconResolver.php
```diff
@@ -0,0 +1,79 @@
...
+     * Returns path to content type icon.
+     *
+     * Path is resolved based on configuration (ezpublish.system.<SCOPE>.content_type). If there isn't coresponding
+     * entry for given content type, then path to default icon will be returned.
+     *
+     * @param string $contentType
+     *
+     * @return string|null
+     */
+    public function getContentTypeIcon(string $contentType): ?string
```
dew326: `$contentTypeIdentifier` or just `$identifier`?

mikadamczyk:
As we talked, we can add the configuration for the default icon as part of the `content_type` configuration.

adamwojs:
> As we talked, we can add the configuration for the default icon as part of the content_type configuration.

Done in https://github.com/ezsystems/ezplatform-admin-ui/pull/804/commits/3b911721f6d0e051b4ba67a311f0987a4840bb6a: default configuration is available under `default-config` key. 

I also added unit tests for `EzSystems\EzPlatformAdminUi\UI\Service\ContentTypeIconResolver`.

src/lib/UI/Service/ContentTypeIconResolver.php
```diff
@@ -0,0 +1,93 @@
...
+namespace EzSystems\EzPlatformAdminUi\UI\Service;
+
+use eZ\Publish\Core\MVC\ConfigResolverInterface;
+use Symfony\Component\Asset\Packages;
+
+final class ContentTypeIconResolver
+{
+    private const DEFAULT_IDENTIFIER = 'default-config';
+    private const PARAM_NAME_FORMAT = 'content_type.%s';
+
```
mikadamczyk: do we need this empty line?

lserwatka:
@DominikaK this should be documented.

PR #805, opened 2019-01-17
EZP-29999: As an editor, I want to see custom icons for content types

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29999
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Requires: https://github.com/ezsystems/ezplatform-admin-ui/pull/804
UDW part in: https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/132
date-based-publisher: https://github.com/ezsystems/date-based-publisher/pull/89
Workflow: https://github.com/ezsystems/ezplatform-workflow/pull/46

## Views with icons

screen shot 2019-01-21 at 13 00 55

### Content item view & edit view
screen shot 2019-01-21 at 13 01 30

screen shot 2019-01-21 at 13 01 42

### Bookmarks
screen shot 2019-01-28 at 09 22 08

### Trash
screen shot 2019-01-28 at 09 22 20

### Drafts
screen shot 2019-01-30 at 13 09 17

### Content type list; content type view; content type edit
screen shot 2019-01-21 at 13 02 51
screen shot 2019-01-21 at 13 14 23
screen shot 2019-01-21 at 13 14 39

### Dashboard
screen shot 2019-01-30 at 13 08 45

#### Checklist:
- [x] pass content types data to twig templates, which lack them
- [x] add `ez_content_type_icon` where necessary
- [x] ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.location.view.js
+++ b/src/bundle/Resources/public/js/scripts/admin.location.view.js
@@ -127,6 +127,7 @@
                         attrs: Object.assign({}, mfuAttrs, {
                             onPopupClose: (itemsUploaded) => itemsUploaded.length && global.location.reload(true),
                             contentCreatePermissionsConfig: JSON.parse(container.dataset.mfuCreatePermissionsConfig),
+                            contentTypesMap,
                         }),
                     },
                 ],
--- a/src/bundle/Resources/public/js/scripts/admin.location.view.js
+++ b/src/bundle/Resources/public/js/scripts/admin.location.view.js
@@ -113,6 +113,13 @@
             return total;
         }, {});
         const udwConfigBulkMoveItems = JSON.parse(container.dataset.udwConfigBulkMoveItems);
+        const mfuContentTypesMap = Object.values(eZ.adminUiConfig.contentTypes).reduce((contentTypeDataMap, contentTypeGroup) => {
+            for (const contentTypeData of contentTypeGroup) {
+                contentTypeDataMap[contentTypeData.href] = contentTypeData;
+            }
+
+            return contentTypeDataMap;
+        }, {});
 
         ReactDOM.render(
             React.createElement(eZ.modules.SubItems, {
@@ -127,7 +134,7 @@
                         attrs: Object.assign({}, mfuAttrs, {
                             onPopupClose: (itemsUploaded) => itemsUploaded.length && global.location.reload(true),
                             contentCreatePermissionsConfig: JSON.parse(container.dataset.mfuCreatePermissionsConfig),
-                            contentTypesMap,
+                            contentTypesMap: mfuContentTypesMap,
                         }),
                     },
                 ],
--- a/src/bundle/Resources/public/scss/_instant-filter.scss
+++ b/src/bundle/Resources/public/scss/_instant-filter.scss
@@ -8,21 +8,33 @@
 .ez-extra-actions--create {
     .ez-instant-filter {
         &__group-name {
-            padding-left: .5rem;
+            padding-left: calculateRem(8px);
         }
 
         &__input-wrapper {
-            padding: 1rem .5rem;
+            padding: calculateRem(16px) calculateRem(8px);
         }
 
         &__group-item {
+            position: relative;
+
             label {
                 cursor: pointer;
+            }
+
+            .ez-icon {
+                fill: $ez-black;
+                position: absolute;
+                top: 0;
+                left: calculateRem(20px);
+                width: calculateRem(16px);
+                height: calculateRem(16px);
+                margin-top: calculateRem(3px);
             }
         }
 
         &__items {
-            max-height: 200px;
+            max-height: calculateRem(200px);
             overflow: auto;
         }
     }
--- a/src/bundle/Resources/views/admin/content_type/edit.html.twig
+++ b/src/bundle/Resources/views/admin/content_type/edit.html.twig
@@ -17,7 +17,7 @@
 {% block page_title_admin %}
     {% include '@ezdesign/parts/page_title.html.twig' with {
         title: 'content_type.view.edit.title'|trans({ '%name%': content_type.name })|desc('Editing Content Type: %name%'),
-        contentType: content_type.identifier
+        content_type_identifier: content_type.identifier
     } %}
 {% endblock %}
 
--- a/src/bundle/Resources/views/admin/content_type/edit.html.twig
+++ b/src/bundle/Resources/views/admin/content_type/edit.html.twig
@@ -17,7 +17,7 @@
 {% block page_title_admin %}
     {% include '@ezdesign/parts/page_title.html.twig' with {
         title: 'content_type.view.edit.title'|trans({ '%name%': content_type.name })|desc('Editing Content Type: %name%'),
-        iconName: 'content-type'
+        contentType: content_type.identifier
     } %}
 {% endblock %}
 
--- a/src/bundle/Resources/views/parts/page_title.html.twig
+++ b/src/bundle/Resources/views/parts/page_title.html.twig
@@ -5,6 +5,10 @@
             <svg class="ez-icon ez-icon-{{ iconName }}">
                 <use xlink:href="{{ asset('bundles/ezplatformadminui/img/ez-icons.svg') }}#{{ iconName }}"></use>
             </svg>
+        {% elseif contentType is defined %}
+            <svg class="ez-icon ez-icon-{{ contentType }}">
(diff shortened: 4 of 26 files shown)


src/bundle/Templating/Twig/ContentTypeIconExtension.php
```diff
@@ -0,0 +1,43 @@
...
+     */
+    public function __construct(ContentTypeIconResolver $contentTypeIconResolver)
+    {
+        $this->contentTypeIconResolver = $contentTypeIconResolver;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getFunctions(): array
```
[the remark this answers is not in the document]
tischsoic: This PR containes code from: https://github.com/ezsystems/ezplatform-admin-ui/pull/804 as it depends on it.
 
Please comment PHP in https://github.com/ezsystems/ezplatform-admin-ui/pull/804. 
Corresponding line: https://github.com/ezsystems/ezplatform-admin-ui/pull/804/files#diff-a7ba97e75601073fc38f8daa145f316dR31

BTW It's Adam's part - ping @adamwojs.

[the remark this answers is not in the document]
tischsoic: But it seems to be inherited from `Twig_Extension` and it "Returns a list of functions to add to the existing list." (https://twig.symfony.com/doc/2.x/advanced.html)

src/bundle/DependencyInjection/Configuration/Parser/ContentType.php
```diff
@@ -0,0 +1,60 @@
...
+use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\SiteAccessAware\ContextualizerInterface;
+use Symfony\Component\Config\Definition\Builder\NodeBuilder;
+
+/**
+ * Configuration parser for subtree related operations.
+ *
+ * Example configuration:
+ *
+ * ```yaml
+ * ezpublish:
```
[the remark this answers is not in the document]
adamwojs: No.

src/bundle/Templating/Twig/ContentTypeIconExtension.php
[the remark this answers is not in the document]
adamwojs: In this way you are defining custom Twig functions provided by extension. As you me notice this class implement interface from the 3rd party library so "naming" is not subject to discussion.

tischsoic:
Branch was rebased onto master and is ready for review.

src/bundle/Resources/public/scss/_instant-filter.scss
```diff
@@ -8,21 +8,33 @@
...
                 cursor: pointer;
             }
+
+            .ez-icon {
+                fill: $ez-black;
+                position: absolute;
+                top: 0;
+                left: calculateRem(20px);
+                width: calculateRem(16px);
+                height: calculateRem(16px);
```
dew326: You should use `ez-icon--small`

src/bundle/Resources/views/admin/content_type/edit.html.twig
```diff
@@ -17,7 +17,7 @@
 {% block page_title_admin %}
     {% include '@ezdesign/parts/page_title.html.twig' with {
         title: 'content_type.view.edit.title'|trans({ '%name%': content_type.name })|desc('Editing Content Type: %name%'),
-        iconName: 'content-type'
+        contentType: content_type.identifier
```
dew326: It's not `contentType`. Variables in twig are with `_`.

src/bundle/Resources/views/parts/page_title.html.twig
```diff
@@ -5,6 +5,10 @@
             <svg class="ez-icon ez-icon-{{ iconName }}">
                 <use xlink:href="{{ asset('bundles/ezplatformadminui/img/ez-icons.svg') }}#{{ iconName }}"></use>
             </svg>
+        {% elseif contentType is defined %}
+            <svg class="ez-icon ez-icon-{{ contentType }}">
```
dew326: Something is wrong here. From a line below `contentType` is an object not string. In the other template you pass `identifier` as a `contentType`.

src/bundle/Resources/views/admin/content_type/edit.html.twig
tischsoic: There was already `iconName` in`/views/parts/page_title.html.twig`, this is why I opted for `contentType` to make it consistent within template file.

dew326: In this case consistent is not a good way.

src/bundle/Resources/public/scss/_instant-filter.scss
tischsoic: I defined it here in order to overwrite rule: `.ez-context-menu .ez-sticky-container .ez-icon`.

src/bundle/Resources/views/admin/content_type/edit.html.twig
tischsoic: OK, changed. :)

src/bundle/Resources/views/parts/page_title.html.twig
tischsoic: You was right, very good catch, I'm impressed! 😄 
Fixed in: https://github.com/ezsystems/ezplatform-admin-ui/pull/805/commits/92412c2a9ae2715780dd70178e11f1f96c29dee8

src/bundle/Resources/views/parts/tab/content_type.html.twig
```diff
@@ -17,7 +17,7 @@
                 {% block page_title %}
                     {% include '@ezdesign/parts/page_title.html.twig' with {
                         title: 'content_type.view.view.title'|trans({ '%name%': content_type.name })|desc('%name%'),
-                        iconName: 'content-type'
+                        contentType: content_type.identifier
```
dew326: `content_type_identifier`?

src/lib/Tab/Dashboard/PagerContentToDataMapper.php
```diff
@@ -76,6 +76,8 @@ public function map(Pagerfanta $pager): array
                 'language' => $contentInfo->mainLanguageCode,
                 'contributor' => $contributor,
                 'version' => $content->versionInfo->versionNo,
+                'type' => $content->getContentType()->getName(),
+                'content_type' => $content->getContentType(),
```
webhdx: What's the point of having these two? `content_type` is sufficient if this goes to the twig.

tischsoic: It is @adamwojs part, but I think that it is my error made during rebasing.

barbaragr:
Everything was fixed, just MFU left. It occurs when user tries to upload items when folder is empty - uploading for the first time. Later everything works fine.

src/bundle/Resources/public/js/scripts/admin.location.view.js
```diff
@@ -113,6 +113,13 @@
             return total;
         }, {});
         const udwConfigBulkMoveItems = JSON.parse(container.dataset.udwConfigBulkMoveItems);
+        const mfuContentTypesMap = Object.values(eZ.adminUiConfig.contentTypes).reduce((contentTypeDataMap, contentTypeGroup) => {
```
tischsoic: We need to pass different config to MFU because the one based on: `container.dataset.contentTypes` contains only CTs of elements displayed in Sub-items. 

Given that `mfuContentTypesMap` in contrast to  `contentTypesMap` (which is passed to Sub-items) does not contain whole CT objects but rather data stored in `eZ.adminUiConfig.contentTypes`, are we OK with passing `mfuContentTypesMap` to MFU as `contentTypesMap` - ping @dew326  @sunpietro ?

dew326: I think the approach is ok.

sunpietro: Is a difference only in missing data or in the data structure as well?

sunpietro: Thanks for explanation. Can't the backend provide this structure instead?

tischsoic: We opted for this solution mostly because this is more efficient than extracting objects for all Content Types. But maybe it is way to go. This question is more to @adamwojs .

PR #809, opened 2019-01-25
[WIP] Wait until DraftConflict dialog is visible

mnocon:
Trying to fix an error from build: https://travis-ci.com/ezsystems/ezplatform-form-builder/jobs/172979252

```
   And I start creating new draft from draft conflict modal                           # EzSystems\EzPlatformAdminUi\Behat\BusinessContext\ContentViewContext::startCreatingNewDraftFromDraftConflictModal()
      WebDriver\Exception\ElementNotVisible: element not interactable
        (Session info: chrome=71.0.3578.98)
        (Driver info: chromedriver=2.45.615279 (12b89733300bd268cff3b78fc76cb8f3a7cc44e5),platform=Linux 4.4.0-101-generic x86_64) in vendor/instaclick/php-webdriver/lib/WebDriver/Exception.php:155
```

Screen: https://res.cloudinary.com/ezplatformtravis/image/upload/v1548415191/screenshots/vendor_ezsystems_ezplatform-form-builder_features_formcontentitemadministration_feature_123_zavl4r.png

Looks like we're trying to interact with the popup before it's ready.

### DIFF
--- a/src/lib/Behat/BusinessContext/ContentViewContext.php
+++ b/src/lib/Behat/BusinessContext/ContentViewContext.php
@@ -176,6 +176,7 @@ class ContentViewContext extends BusinessContext
     public function startCreatingNewDraftFromDraftConflictModal(): void
     {
         $draftConflictModal = ElementFactory::createElement($this->utilityContext, DraftConflictDialog::ELEMENT_NAME);
+        $draftConflictModal->verifyVisibility();
         $draftConflictModal->createNewDraft();
     }
 
@@ -185,6 +186,7 @@ class ContentViewContext extends BusinessContext
     public function startEditingDraftFromDraftConflictModal(string $draftID): void
     {
         $draftConflictModal = ElementFactory::createElement($this->utilityContext, DraftConflictDialog::ELEMENT_NAME);
+        $draftConflictModal->verifyVisibility();
         $draftConflictModal->dashboardTable->clickEditButton($draftID);
     }
 


src/lib/Behat/Helper/UtilityContext.php
```diff
@@ -265,7 +265,7 @@ public function waitUntilElementDisappears(string $cssSelector, int $timeout): v
     {
         try {
             $this->waitUntil($timeout, function () use ($cssSelector, $timeout) {
-                return !$this->isElementVisible($cssSelector, $timeout);
+                return !$this->isElementVisible($cssSelector, 1);
```
micszo: Could you explain this change? :) Looks like hardcoded value.

mnocon: Sure. The problem is that `isElementVisible` also has a (blocking) waitUntil inside it - when `$timeout` is passed as an argument then the `isElementVisible` will wait the whole `$timeout` for the Element to become visible - making the outer loop not working as intended.

Example (current behaviour)

```
$timeout = 5;
$this->waitUntil($timeout, function () use ($cssSelector, $timeout) {
                return !$this->isElementVisible($cssSelector, $timeout);
}
```

The first time `isElementVisible` is invoked there are two possibilities:
- Element is visible - function returns true immediately
- Element is not visible - function returns false after 5 seconds

Only then the (outer) waitUntil is taken into account:
- Element is visible - isElementVisible is run again
- Element is not visible - function passes

This means that in both scenarios (Element is visible all the time and Element not visible at all) the function will take $timeout seconds to execute it. Turning it into a more complex sleep 😛 

Example after changes:

```
$timeout = 5;
$this->waitUntil($timeout, function () use ($cssSelector, $timeout) {
                return !$this->isElementVisible($cssSelector, 1);
}
```

The first time `isElementVisible` is invoked there are two possibilities:
- Element is visible - function returns true immediately
- Element is not visible - function returns false after 1 second

Only then the (outer) waitUntil is taken into account:
- Element is visible - isElementVisible is run again
- Element is not visible - function passes

Result:
in one scenario (Element is visible all the time) the function takes $timeout seconds, in the other (Element not visible at all) the function takes 1 second to execute.

The current behaviour becomes a problem when $timeout values are getting bigger, for example 20-30 seconds.

PR #811, opened 2019-01-25
EZP-29320: As an Editor, I want to be able to switch content main tra…

pawbuj:
…nslation in PlatformUI

| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-29320](https://jira.ez.no/browse/EZP-29320)
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Allows users of the admin panel to change the main language of the content.

#### Checklist:
- [x] Ready for Code Review


src/bundle/Controller/ContentController.php
```diff
@@ -331,4 +333,57 @@ public function previewAction(
             'versionNo' => $versionNo ?? $content->getVersionInfo()->versionNo,
         ]);
     }
+
+    /**
+     * @param Request $request
```
webhdx: Please use FQCN in doc blocks. Please fix other cases as well as I notice it in other files too.

src/bundle/Controller/ContentController.php
```diff
@@ -331,4 +333,57 @@ public function previewAction(
...
+     * @throws TranslationInvalidArgumentException
+     * @throws APIRepositoryInvalidArgumentException
+     * @throws UnauthorizedException
+     * @throws InvalidOptionsException
+     */
+    public function updateMainTranslationAction(Request $request): Response
+    {
+        $form = $this->formFactory->updateMainTranslation();
+        $form->handleRequest($request);
+        if ($form->isSubmitted() && $form->isValid()) {
```
webhdx: If I remember correctly using `$form->isValid()` here will disable form errors to notifications mapping. It should be done by `SubmitHandler`. Please confirm I'm correct and fix if needed.

src/bundle/Controller/ContentController.php
```diff
@@ -331,4 +333,57 @@ public function previewAction(
...
+            $result = $this->submitHandler->handle($form, function (MainTranslationUpdateData $data) {
+                $contentInfo = $data->getContentInfo();
+                $mapper = new MainTranslationUpdateMapper();
```
webhdx: In kernel it's called `main language` and I would use this name instead of `main translation` to avoid confusion.

src/bundle/Controller/ContentController.php
```diff
@@ -331,4 +333,57 @@ public function previewAction(
...
+                return new RedirectResponse($this->generateUrl('_ezpublishLocation', [
+                    'locationId' => $contentInfo->mainLocationId,
+                    '_fragment' => 'ez-tab-location-view-translations',
+                ]));
+            });
+            if ($result instanceof Response) {
+                return $result;
+            }
+        }
+        /** @var ContentDraftCreateData $data */
```
webhdx: Should be `\EzSystems\EzPlatformAdminUi\Form\Data\Content\Translation\MainTranslationUpdateData`

src/bundle/Resources/config/routing.yml
```diff
@@ -459,6 +459,12 @@ ezplatform.translation.remove:
     defaults:
         _controller: 'EzPlatformAdminUiBundle:Translation:remove'
 
+ezplatform.content.update_main_translation:
```
webhdx: Note my above comment about naming - if you (and others) agree you have to fix route name and path as well.

src/bundle/Resources/translations/forms.en.xliff
```diff
@@ -491,21 +491,6 @@
         <target>User / Selfedit</target>
         <note>key: role.policy.user.selfedit</note>
       </trans-unit>
-      <trans-unit id="30c5e1c11a520e499ad68130a2c7b844c2243888" resname="role.policy.workflow">
```
webhdx: Even though this should not be a part of AdminUI's translations I wouldn't remove it in this PR because it will require moving these translation strings to `ezsystems/ezplatform-workflow`.

src/bundle/Resources/views/content/tab/translations/tab.html.twig
```diff
@@ -21,6 +21,9 @@
             <th></th>
             <th>{{ 'tab.translations.language_name'|trans|desc('Language name') }}</th>
             <th>{{ 'tab.translations.language_code'|trans|desc('Language code') }}</th>
+            {% if main_translation_switch %}
```
webhdx: You could just use the fact that form is passed to the template to display this column, you don't need additional flag for this.

src/lib/Form/Data/Content/Translation/MainTranslationUpdateData.php
```diff
@@ -0,0 +1,76 @@
...
+     * @var string
+     */
+    public $languageCode;
+
+    /**
+     * @param ContentInfo|null $contentInfo
+     * @param string|null $languageCode
+     */
+    public function __construct(
+        ContentInfo $contentInfo = null,
```
webhdx: Since `$contentInfo` property is nullable it should allowing passing nulls directly too: `?ContentInfo $contentInfo = null`

src/lib/Form/DataMapper/MainTranslationUpdateMapper.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    /**
+     * @param ContentMetadataUpdateStruct|ValueObject $value
+     *
+     * @return ContentMainLocationUpdateData
+     */
+    public function map(ValueObject $value)
+    {
+        if (!$value instanceof ContentMetadataUpdateStruct) {
+            throw new InvalidArgumentException('value', 'must be an instance of ' . ContentMetadataUpdateStruct::class);
```
webhdx: Consider using `sprintf()` instead of string concatenation.

src/lib/Tab/LocationView/TranslationsTab.php
```diff
@@ -143,4 +153,19 @@ private function createTranslationDeleteForm(Location $location, array $language
...
+
+    /**
+     * @param ContentInfo $contentInfo
+     * @param string $languageCode
+     *
+     * @return FormInterface
+     *
+     * @throws InvalidOptionsException
+     */
+    private function createMainLanguageUpdateForm(ContentInfo $contentInfo, string $languageCode): FormInterface
```
webhdx: This private method provides little value.

src/lib/Form/Factory/FormFactory.php
```diff
@@ -1330,4 +1332,26 @@ public function trashContainerLocation(
...
+
+    /**
+     * @param MainTranslationUpdateData|null $data
+     * @param string|null $name
+     *
+     * @return FormInterface
+     *
+     * @throws InvalidOptionsException
+     */
+    public function updateMainTranslation(
```
webhdx: It has been decided recently that we are no longer adding new methods to the `FormFactory` and we will use Symfony's FormFactory for this.

src/bundle/Resources/translations/ezrepoforms_content.en.xliff
```diff
@@ -111,6 +111,11 @@
         <target state="new">Upload image</target>
         <note>key: fieldtype.ezimageasset.upload.label</note>
       </trans-unit>
+      <trans-unit id="d773dd487e4e5467fd20fc4fc4987eba33eaf1b7" resname="content.main_language_update.success">
+        <source>Main language for '%name%' updated.</source>
+        <target state="new">Main language for '%name%' updated.</target>
+        <note>key: content.main_language_update.success</note>
```
dew326: This key is in `content.en.xliff` and `ezrepoforms_content.en.xliff` but used is only once. Is it correct?

src/bundle/Resources/views/content/locationview.html.twig
```diff
@@ -170,6 +170,7 @@
         '@EzPlatformAdminUiBundle/Resources/public/js/scripts/button.state.toggle.js'
         '@EzPlatformAdminUiBundle/Resources/public/js/scripts/admin.version.edit.conflict.js'
         '@EzPlatformAdminUiBundle/Resources/public/js/scripts/admin.location.bookmark.js'
+        '@EzPlatformAdminUiBundle/Resources/public/js/scripts/admin.main_translation.update.js'
```
dew326: In JS files we do not use the convention with `_`. Should be `admin.main.translation.update.js`.

src/bundle/Resources/public/js/scripts/admin.main.translation.update.js
```diff
@@ -0,0 +1,10 @@
+(function(global, doc) {
+    const updateMainTranslation = (event) => {
+        doc.querySelector('#main_translation_update_language_code').value = event.target.value;
```
tischsoic: Was this ID generated on the backend? If it was generated on the backend, I think it should be passed via data-attribute to JS.

tischsoic: Was this ID generated on the backend? If it was generated on the backend, I think it should be passed via data-attribute to JS.

src/bundle/Resources/public/js/scripts/admin.main.translation.update.js
```diff
@@ -0,0 +1,10 @@
...
+        doc.querySelector('form[name="main_translation_update"]').submit();
+    };
+
+    [...doc.querySelectorAll('input[name="updateMainTranslation"]')].forEach((input) =>
```
tischsoic: There is no longer need to use `[...` with `querySelectorAll`. We used to add this because in older versions of Edge forEach was not supported on list returned by `querySelectorAll`.

src/bundle/Resources/public/js/scripts/admin.main.translation.update.js
```diff
@@ -0,0 +1,10 @@
...
+        input.addEventListener('change', updateMainTranslation, false)
```
tischsoic: Nitpick: no `;` at the end of expression, Prettier.

src/bundle/Resources/public/js/scripts/admin.main.translation.update.js
tischsoic: But I'm not 100% sure if we have such rule, ping @dew326 ?

src/bundle/Resources/public/js/scripts/admin.main.translation.update.js
tischsoic: I think I'm might be wrong here. 🤔

src/bundle/Resources/public/js/scripts/admin.main.translation.update.js
dew326: We do this mostly in generic and reusable classes. Here it's not a problem to be hardcoded.

src/bundle/Resources/public/js/scripts/admin.main.translation.update.js
sunpietro: I see there are multiple naming conventions for `name` attribute: `main_translation_update`, `updateMainTranslation`. Can you choose one convention and follow it?

sunpietro: Another thing is, that for better code readability you could cache there result of `doc.querySelectorAll('input[name="updateMainTranslation"]')` in a variable like: `const inputs`. Then you won't have to move the `input.addEventListener...` part to the next line.

src/bundle/Resources/translations/content.en.xliff
```diff
@@ -11,6 +11,11 @@
         <target state="new">New Version Draft for '%name%' created.</target>
         <note>key: content.create_draft.success</note>
       </trans-unit>
+      <trans-unit id="d773dd487e4e5467fd20fc4fc4987eba33eaf1b7" resname="content.main_language_update.success">
+        <source>Main language for '%name%' updated.</source>
+        <target state="new">Main language for '%name%' updated.</target>
```
sunpietro: Please, remove `state="new"` from here.

src/bundle/Resources/translations/locationview.en.xliff
```diff
@@ -386,6 +386,11 @@
         <target state="new">Language name</target>
         <note>key: tab.translations.language_name</note>
       </trans-unit>
+      <trans-unit id="c5d8ba580b9dd86b38986590b338816e3193c857" resname="tab.translations.main_language">
+        <source>Main language</source>
+        <target state="new">Main language</target>
```
sunpietro: Please, remove `state="new"` from here

src/bundle/Resources/translations/content.en.xliff
ViniTou: I think we had this disscusion and that info is useful for some people.

mikadamczyk: we talked about it and AFAIK we decided to not remove this information

PR #814, opened 2019-01-29
EZP-30064: As an editor I would like to view a content tree in location view

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30064
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes/no
| Doc needed?   | yes/no

zrzut ekranu 2019-01-29 o 16 37 17

## FOR QA AND REVIEWERS:

The feature will be implemented incrementally. That means, that the design will be changing in the forthcoming pull requests. In this PR I focused on preparing the content tree module base and implementing the basic file structure.

Requires: https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/136


src/bundle/Resources/public/scss/_view.content.scss
```diff
@@ -0,0 +1,14 @@
+.ez-content-view {
+    .ez-content-container {
+        display: flex;
+    }
+
+    .ez-content-tree-container {
+        min-width: 250px;
```
tischsoic: Is there any reason not to use `calculateRem(250px)` here?

sunpietro: Good catch.

src/bundle/Resources/views/layout.html.twig
```diff
@@ -164,6 +164,7 @@
             '@EzPlatformAdminUiBundle/Resources/public/js/scripts/helpers/timezone.helper.js'
             '@EzPlatformAdminUiBundle/Resources/public/js/scripts/admin.notifications.js'
             '@EzPlatformAdminUiModulesBundle/Resources/public/js/UniversalDiscovery.module.js'
+            '@EzPlatformAdminUiModulesBundle/Resources/public/js/ContentTree.module.js'
```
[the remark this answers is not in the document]
sunpietro: Good catch. Fixed.

sunpietro:
ping pong @dew326 @tischsoic

src/bundle/Resources/public/scss/_view.content.scss
```diff
@@ -0,0 +1,14 @@
+.ez-content-view {
```
dew326: The files should be named `_content-view.scss`. Here the pattern is with `-`.

src/bundle/Resources/views/content/locationview.html.twig
```diff
@@ -170,6 +174,7 @@
         '@EzPlatformAdminUiBundle/Resources/public/js/scripts/button.state.toggle.js'
         '@EzPlatformAdminUiBundle/Resources/public/js/scripts/admin.version.edit.conflict.js'
         '@EzPlatformAdminUiBundle/Resources/public/js/scripts/admin.location.bookmark.js'
+        '@EzPlatformAdminUiModulesBundle/Resources/public/js/ContentTree.module.js'
```
dew326: Is it working? The JS which needs this file is added before. This should be moved to the `react_modules` block.

src/bundle/Resources/public/scss/_view.content.scss
sunpietro: I disagree. In this repo we have a convention with dots.

sunpietro:
ping @dew326

PR #820, opened 2019-01-31
EZP-30069: As an editor I would like to change manually the width of content tree panel

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30069
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | yes/no
| Tests pass?   | yes/no
| Doc needed?   | yes/no

### DIFF
--- a/src/bundle/Resources/public/scss/_view-content.scss
+++ b/src/bundle/Resources/public/scss/_view-content.scss
@@ -5,10 +5,18 @@
 
     .ez-content-tree-container {
         min-width: calculateRem(250px);
-        color: inherit;
+        max-width: 33vw;
+        position: relative;
+        background: $ez-white;
     }
 
     .ez-location-view-container {
         flex: 1 1 auto;
+    }
+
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/scss/_view-content.scss
```diff
@@ -5,10 +5,18 @@
...
+        background: $ez-white;
     }
 
     .ez-location-view-container {
         flex: 1 1 auto;
     }
+
+    &.ez-is-tree-resizing {
+        .ez-location-view-container {
+            pointer-events: none;
```
tischsoic: We do this so that when dragging user is not able to accidentally select some text etc? Asking just out of sheer curiosity. 🙂

sunpietro: Yes, that's the goal of this change :)

PR #825, opened 2019-01-31
EZP-30096: Solr scoring is not taken into account in the backend search results list order

kmadejski:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30096](https://jira.ez.no/browse/EZP-30096)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   |no

As mentioned in the JIRA ticket, Solr scoring is not taken in the search results list order. 

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/SearchController.php
+++ b/src/bundle/Controller/SearchController.php
@@ -8,6 +8,7 @@ namespace EzSystems\EzPlatformAdminUiBundle\Controller;
 
 use eZ\Publish\API\Repository\Values\Content\Query;
 use eZ\Publish\API\Repository\Values\Content\Query\Criterion;
+use eZ\Publish\API\Repository\Values\Content\Query\SortClause;
 use eZ\Publish\API\Repository\Values\User\User;
 use eZ\Publish\Core\Pagination\Pagerfanta\ContentSearchAdapter;
 use eZ\Publish\API\Repository\SearchService;
@@ -142,9 +143,8 @@ class SearchController extends Controller
                 $fullText = null;
                 $criteria = [];
 
-                if (null !== $queryString) {
-                    $fullText = new Criterion\FullText($queryString);
-                }
+                $query->query = new Criterion\FullText($queryString);
+
                 if (null !== $section) {
                     $criteria[] = new Criterion\SectionId($section->id);
                 }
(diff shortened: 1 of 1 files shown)


src/bundle/Controller/SearchController.php
```diff
@@ -175,8 +175,9 @@ public function searchAction(Request $request): Response
                 if (!empty($criteria)) {
                     $query->filter = new Criterion\LogicalAnd($criteria);
                 }
-
-                $query->sortClauses[] = new SortClause\DateModified(Query::SORT_ASC);
```
andrerom: It might be cleaner to check for scoring capability, if not present, the sorting by date can be a fallback.
ref: https://github.com/ezsystems/ezplatform-solr-search-engine/blob/master/lib/Handler.php#L451

src/bundle/Controller/SearchController.php
```diff
@@ -140,10 +139,11 @@ public function searchAction(Request $request): Response
                 $created = $data->getCreated();
                 $creator = $data->getCreator();
                 $query = new Query();
+                $fullText = null;
                 $criteria = [];
 
                 if (null !== $queryString) {
-                    $criteria[] = new Criterion\FullText($queryString);
+                    $fullText = new Criterion\FullText($queryString);
```
andrerom: Unless you plan on taking this into account for sorting, this can be simplified to:
```php
$query->query = new Criterion\FullText($queryString);
```

src/bundle/Controller/SearchController.php
kmadejski: Good catch, I wasn't aware that this is possible to check whether `SearchService` supports some features. Thanks!

src/bundle/Controller/SearchController.php
kmadejski: Changed in https://github.com/ezsystems/ezplatform-admin-ui/pull/825/commits/5e8832e655b513b0f439091272df69a70f3b4eda.

src/bundle/Controller/SearchController.php
kmadejski: Good catch, I wasn't aware that this is possible to check whether SearchService supports some features. Thanks!

micszo:
So far retest ok in Content / Search on v2.2.3.
But passed a concern about UDW on priv.

lserwatka:
Could you merge it up?

kmadejski:
@lserwatka done in https://github.com/ezsystems/ezplatform-admin-ui/commit/f60970f924603451d16ac41b6b5cd5ca34a89ff5 / https://github.com/ezsystems/ezplatform-admin-ui/commit/1398555402ebb68e8f712aa0120ddec688a9c9e8 / https://github.com/ezsystems/ezplatform-admin-ui/commit/fead9059f07c56504cf24475351461b0c6b981d6.

PR #831, opened 2019-02-05
EZP-29280: As an editor, I want to add anchors to RichText content

dew326:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29280
| **Depends on** | https://github.com/ezsystems/ezplatform-richtext/pull/28 |
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

This PR introduces new svg icon `link-anchor`:
screen shot 2019-02-12 at 13 48 07

#### Checklist:
- [ ] Ready for Code Review


src/bundle/Resources/public/scss/_alloyeditor-anchor-edit.scss
```diff
@@ -0,0 +1,50 @@
+.ez-ae-anchor-edit {
+    display: flex;
+    padding: calculateRem(8px) 0 calculateRem(8px) calculateRem(8px);
+
+    &__input-wrapper {
```
sunpietro: What about moving it inside the `&__input {` selector? In fact, it looks like it should be a separate component, but it's acceptable use case here.

src/bundle/Resources/public/scss/_alloyeditor-anchor-edit.scss
```diff
@@ -0,0 +1,50 @@
...
+
+        &--trash {
+            background-color: $ez-color-base-dark;
+        }
+
+        &--save {
+            background-color: $ez-color-positive;
+        }
+
+        &[disabled] {
```
sunpietro: I'm still thinking about unifying this. Because I found we're adding the similar rule here and there in our codebase. We have to check whether we could unify it and make it more generic.

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-anchor.js
```diff
@@ -0,0 +1,45 @@
...
+     * @method render
+     * @return {Object} The content which should be rendered.
+     */
+    render() {
+        if (this.props.renderExclusive) {
+            return <AlloyEditor.EzBtnAnchorEdit {...this.props} />;
+        }
+
+        const cssClass = 'ae-button ez-btn-ae--anchor ez-btn-ae ' + this.getStateClasses();
+        const label = Translator.trans(/*@Desc("Anchor")*/ 'anchor_btn.label', {}, 'alloy_editor');
```
tischsoic: Not sure, are we missing generated translation here and in `ez-btn-anchoredit.js`?

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-anchor.js
```diff
@@ -0,0 +1,45 @@
...
+     *
+     * @method render
+     * @return {Object} The content which should be rendered.
+     */
+    render() {
+        if (this.props.renderExclusive) {
+            return <AlloyEditor.EzBtnAnchorEdit {...this.props} />;
+        }
+
+        const cssClass = 'ae-button ez-btn-ae--anchor ez-btn-ae ' + this.getStateClasses();
```
tischsoic: Nitpick: we usually use `` `${}` `` to concatenate strings. Do we have any in-house rule on this? There is eslint rule for this: https://eslint.org/docs/rules/prefer-template.

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-anchoredit.js
```diff
@@ -0,0 +1,164 @@
...
+
+        this.block = block;
+
+        return block;
+    }
+
+    findIcon() {
+        const block = this.findBlock();
+
+        return [...block.getChildren().$].find((child) => child.classList && child.classList.contains('ez-icon--anchor'));
```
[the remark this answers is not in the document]
tischsoic: I would extract arrow function to method to improve readability.

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-anchoredit.js
```diff
@@ -0,0 +1,164 @@
...
+
+    updateValue({ nativeEvent }) {
+        const value = nativeEvent.target.value;
+
+        this.setState(() => ({ value }));
+    }
+
+    removeAnchor() {
+        const block = this.findBlock();
+        const icon = this.findIcon();
```
tischsoic: Nitpick: we can pass `block` to  `findIcon` instead of finding it twice.

src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
```diff
@@ -130,6 +130,10 @@
             headers.forEach((header) => header.remove());
         }
 
+        clearAnchorIcons(icon) {
```
tischsoic: Nitpick: `clearAnchorIcon`, without `s` as we clear only one icon.

src/bundle/Resources/public/scss/_alloyeditor-anchor-edit.scss
tischsoic: Are you suggesting using this structure:
```scss
 &__input {
    /*...*/
    &-wrapper {/*...*/}
    &-label {/*...*/}
}
```
? I have bad experience with splitting selectors in this way. I have once searched for a selector and it took me a while, because _element_ name was split in halve. 
For me the most natural and easy to comprehend approach is to split class name in 3: _block_, _element_, _modifier_, like it was done here. 🙂

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-anchor.js
```diff
@@ -0,0 +1,45 @@
...
+        const label = Translator.trans(/*@Desc("Anchor")*/ 'anchor_btn.label', {}, 'alloy_editor');
+
+        return (
+            <button
```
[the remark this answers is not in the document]
dew326: Why do we need the `type` in alloyeditor? None of the buttons from AE has this and I follow their convention.

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-anchor.js
```diff
@@ -0,0 +1,45 @@
...
+                aria-pressed={cssClass.indexOf('pressed') !== -1}
+                className={cssClass}
+                onClick={this.props.requestExclusive}
+                tabIndex={this.props.tabIndex}
```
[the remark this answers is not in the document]
dew326: I follow the convention from AE

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-anchor.js
dew326: Yes, most of the button in AE has the convention that I proposed. We should unify this in 3.0.

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-anchor.js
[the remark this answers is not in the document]
sunpietro: Ok, then it's fine.

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-anchoredit.js
dew326: But I wanted to have a more atomic method. And if you look at`findBlock()` you will find out that the method is calculating things only one time, the second one is returning block immediately.

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-anchor.js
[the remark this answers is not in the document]
sunpietro: Unless I'm mistaken the alloy editor app is rendered inside a form. Having `type="button"` added to a button element secures the app and us from a random form submission attempt.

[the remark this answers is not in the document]
dew326: AE is not rendered in the form. It has own container at the end of the page.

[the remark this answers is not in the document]
sunpietro: I see. Anyway, it's a good practice to add this attribute, but I'm fine if you want to follow the convention from AE.

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-anchoredit.js
tischsoic: You're right, I didn't notice `this.block = block;`. I like your approach! :)

PR #833, opened 2019-02-07
EZP-30032: CustomTags not saved correctly

mateuszbieniek:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30032
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

This PR adds a line to fire `editorInteraction` event at the end of CustomTag creation in RichText so AlloyEditor's UI is updated and aware of the changes. 

As a side effect this PR fixes also this issue: https://jira.ez.no/browse/EZP-30010


src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-customtag-update.js
```diff
@@ -202,6 +202,8 @@ export default class EzBtnCustomTagUpdate extends EzWidgetButton {
         Object.keys(this.attributes).forEach((key) => {
             widget.setConfig(key, configValues[key].value);
         });
+
+        widget.fireEditorInteraction('change');
```
dew326: While firing the `editorInteraction` event is the proper fix, I'm not sure if the update button should be responsible for that.
I think it would have more sense if the `widget` fires the event itself when something was changed like it does when the widget is moved. So I think this event should be fired after the `setName`, `setWidgetAttributes`, `setAlignment`, `unsetAlignment`, `clearConfig` and 
`setConfig` methods are invoked. Each event firing should be of course debounced to prevent firing the `setConfigEvent` too many times.

mateuszbieniek: Thanks @dew326 for the suggestion!

src/bundle/Resources/public/js/alloyeditor/src/plugins/ez-custom-tag.js
```diff
@@ -210,6 +210,8 @@
                  */
                 setName: function(name) {
                     this.element.data('ezname', name);
+                    window.clearTimeout(this.setNameFireEditorInteractionTimeout);
```
dew326: Each timeout handler should be defined at the top (after the `editables` object).
```js
setNameFireEditorInteractionTimeout: null
```

src/bundle/Resources/public/js/alloyeditor/src/plugins/ez-custom-tag.js
```diff
@@ -263,6 +265,8 @@
                 setAlignment: function(type) {
                     this.wrapper.data(DATA_ALIGNMENT_ATTR, type);
                     this.element.data(DATA_ALIGNMENT_ATTR, type);
+                    window.clearTimeout(this.setAlignmentFireEditorInteractionTimeout);
+                    this.setAlignmentFireEditorInteractionTimeout = window.setTimeout(this.fireEditorInteraction.bind(this,'aligmentUpdated'), 50);
```
dew326: Missing space after `,` in `bind(this,'aligmentUpdated')`, applies in other cases.

src/bundle/Resources/public/js/alloyeditor/src/plugins/ez-custom-tag.js
```diff
@@ -274,6 +278,8 @@
                 unsetAlignment: function() {
                     this.wrapper.data(DATA_ALIGNMENT_ATTR, false);
                     this.element.data(DATA_ALIGNMENT_ATTR, false);
+                    window.clearTimeout(this.unsetAlignmentFireEditorInteractionTimeout);
+                    this.unsetAlignmentFireEditorInteractionTimeout = window.setTimeout(this.fireEditorInteraction.bind(this,'aligmentRemoved'), 50);
```
dew326: Those lines seem to be too long. Do you use Prettier?

mateuszbieniek: Nope, I will in the future, thanks!

lserwatka:
Could you merge it up?

mateuszbieniek:
> Could you merge it up?

done

PR #835, opened 2019-02-07
EZP-29410: As an Editor I want to be informed that I can't edit CI because of missing policies.

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | [https://jira.ez.no/browse/EZP-29410](https://jira.ez.no/browse/EZP-29410)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

To perform a conflict check action content/versionread permission is needed. If the user has no policy, he will be informed in the notification.

screen shot 2019-02-07 at 3 59 49 pm

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/button.content.edit.js
+++ b/src/bundle/Resources/public/js/scripts/button.content.edit.js
@@ -45,7 +45,7 @@
             if (response.status === 409) {
                 response.text().then(showModal);
             } else if (response.status === 403) {
-                response.text().then((error) => showErrorNotification(error));
+                response.text().then(showErrorNotification);
             } else if (response.status === 200) {
                 submitVersionEditForm();
             }
--- a/src/bundle/Resources/public/js/scripts/button.content.edit.js
+++ b/src/bundle/Resources/public/js/scripts/button.content.edit.js
@@ -1,5 +1,6 @@
-(function (global, doc, $) {
+(function (global, doc, $, eZ) {
     const FORM_EDIT = 'form.ez-edit-content-form';
+    const showErrorNotification = eZ.helpers.notification.showErrorNotification;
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/public/js/scripts/button.content.edit.js
```diff
@@ -34,6 +35,10 @@
             $('#version-draft-conflict-modal').modal('show');
         };
 
+        const showError = (error) => {
```
dew326: No empty line between `const`

sunpietro: Please make it a one liner: `const showError = (error) => showErrorNotification(error);`

dew326: I have a better idea. Why not call `showErrorNotification` directly in `then()`?

mikadamczyk: @dew326 @sunpietro is it ok now?

src/bundle/Resources/public/js/scripts/button.content.edit.js
```diff
@@ -43,11 +44,13 @@
             // Otherwise we can go to Content Item edit page.
             if (response.status === 409) {
                 response.text().then(showModal);
+            } else if (response.status === 403) {
+                response.text().then((error) => showErrorNotification(error));
```
dew326:
```suggestion
                response.text().then(showErrorNotification);
```

lserwatka:
Could you merge it up?

mikadamczyk:
@lserwatka done

PR #836, opened 2019-02-08
EZP-30066: As an editor I would like to see current location marked as selected in the content tree

sunpietro:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30066
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | yes/no
| Tests pass?   | yes/no
| Doc needed?   | yes/no

### DIFF
--- a/src/bundle/Resources/views/content/locationview.html.twig
+++ b/src/bundle/Resources/views/content/locationview.html.twig
@@ -26,7 +26,7 @@
         {% endblock left_sidebar %}
 
         <div class="px-0 pb-4 ez-content-container">
-            <div class="ez-content-tree-container"></div>
+            <div class="ez-content-tree-container" data-current-location-id="{{ location.id }}"></div>
             <div class="ez-location-view-container">
                 <div class="ez-header">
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/views/content/locationview.html.twig
```diff
@@ -13,6 +13,10 @@
 
 {% block body_class %}ez-content-view ez-has-two-sidebars{% endblock %}
 
+{% block meta %}
+<meta name="CurrentLocationId" content="{{ location.id }}" />
```
sunpietro: @webhdx are you ok with this approach of exposing meta information for React-based scripts?

webhdx: Wouldn't passing such information to props upon component creation be a better solution than this? I mean component themselves shouldn't access their context directly.

webhdx: I don't mind this

PR #840, opened 2019-02-11
EZP-30071: As an editor I would like to see tables not breaking location view design while making its container small

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30071
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

## Scrolling tables

This PR introduces table wrappers, which enable scrolling tables:
6c2usawlyb
ua9ye5n4el

-----------------

### Optional columns (we rather don't want this)

I was also playing a little bit with other possibilities and one table also have optional columns, which disappear when view is narrow:
su9cbpjn2e
but I don't think that it is a good idea to implement this without some column toggler (which would allow a user to decide, which column they really want us to show). I started implementing column toggler (like in Sub-items: https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/138 ) but I abandoned this work, as I thought that probably we do not need such column toggler in standard tables.

When it comes to implementation details of "optional columns" I used `ResizeObserver`. `ResizeObserver` is only available in Chrome Browser, but there is polyfill for it, however, it should be possible to implement this without  `ResizeObserver`, based on two events: window resize + content tree resize event.

But as I have already said, I don't think we want this. I just added this feature here because I have already implemented it.


src/bundle/Resources/public/scss/_view-content.scss
```diff
@@ -1,6 +1,7 @@
 .ez-content-view {
     .ez-content-container {
         display: flex;
+        min-width: 0;
```
tischsoic: Note: this is needed because we do not want the content (tables) to dictate container width.
The problem is described here: https://css-tricks.com/flexbox-truncated-text/#article-header-id-3
and exact `min-width` behaviour in flexbox was specified here: https://www.w3.org/TR/css-flexbox-1/#min-size-auto

src/bundle/Resources/public/scss/_view-content.scss
```diff
@@ -19,6 +20,7 @@
 
     .ez-location-view-container {
         flex: 1 1 auto;
+        min-width: 0;
```
tischsoic: Note: _as described above_

lserwatka:
@dew326 @sunpietro could you start with code review?

dew326:
But do we want to use this `ResizeObserver`? It's available only in chrome.

tischsoic:
In commit: https://github.com/ezsystems/ezplatform-admin-ui/pull/840/commits/48388b49dbfe275c46643372b1b2957726c883d3
was improved responsiveness of fieldtypes previews, mainly: ezimage, ezimageasset, ezmedia, ezgmaplocation.

Effect:
- ezimage & ezimageasset
field-preview--ezimage 3
- ezmedia
field-preview--ezmedia 2
-ezgmaplocation
field-preview--ezgmaplocation 2

tischsoic:
I used `ResizeObserver` only in one place to demonstrate possibilities, I will delete this, and ping you afterwards @dew326 .

sunpietro:
@tischsoic please, remove all the code you don't want to be merged.

tischsoic:
I deleted `ResizeObserver`, ready for review. ping @sunpietro @dew326

src/bundle/Resources/public/scss/fieldType/preview/_ezimageasset.scss
```diff
@@ -3,17 +3,23 @@
 
     .ez-field-preview {
         &__image-wrapper {
-            flex-basis: 50%;
-            margin-right: 2rem;
+            flex: 1 1 auto;
```
sunpietro: I can see that `ezimage` and `ezimageasset` field type styles share the same CSS code. Maybe you should think of creating a special class for such situations?

tischsoic: Yes, they are nearly identical, except one, first line. It also got my attention. I think that this is not necessarily in the scope of this task, but of course, we can manage this here. Do you have any idea where extract this? Do we want to share this similar code with `@extend` or in another way? ping @sunpietro @dew326 🙂

tischsoic:
I moved CSSes from ezimage & ezimageasset filed previews to common mixin. ping @dew326 please have a look. Maybe you have idea for better name for mixin than `image-preview`? :)

src/bundle/Resources/public/scss/fieldType/preview/_image-preview.scss
```diff
@@ -0,0 +1,51 @@
+@mixin image-preview() {
```
dew326: For now, we were adding mixins into the `_mixins.scss` but this file might be big in the future.
What about creating a folder called `mixins` and put them there?
Also, I would suggest importing this mixin in the `ezplatform.scss`, thanks to this mixin are "globally"

lserwatka:
@tischsoic rebase is needed here.

lserwatka:
Rebase is needed.

tischsoic:
Rebased

PR #847, opened 2019-02-20
EZP-30100: As an Editor I want to change Object States with dropdowns instead of Set buttons

pawbuj:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30100](https://jira.ez.no/browse/EZP-30100)
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no
| License       | [GPL-2.0](https://github.com/ezsystems/ezplatform-admin-ui/blob/master/LICENSE)

- autosubmit forms mechanism
- autosubmit object state form

#### Checklist:
- [x] Coding standards (`$ composer fix-cs`)
- [x] Ready for Code Review

PR #849, opened 2019-02-22
EZP-30158: As an Editor I want to assign Section or set sorting order…

pawbuj:
… with dropdowns instead of buttons

| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30158](https://jira.ez.no/browse/EZP-30158)
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Autosubmit all forms in the details tab.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/translations/content_type.en.xliff
+++ b/src/bundle/Resources/translations/content_type.en.xliff
@@ -386,11 +386,6 @@
         <target state="new">Sort order</target>
         <note>key: location_update_form.sort_order</note>
       </trans-unit>
-      <trans-unit id="2bb23a852a2a16458d2dad6931debc447564dff2" resname="location_update_form.update">
-        <source>Update</source>
-        <target state="new">Update</target>
-        <note>key: location_update_form.update</note>
-      </trans-unit>
       <trans-unit id="7fe072f41348fb097b6bb018eec4f1b1bc9cffc3" resname="modal.cancel">
         <source>Cancel</source>
         <target state="new">Cancel</target>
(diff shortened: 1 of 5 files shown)


webhdx:
It's a BC break - someone could have overriden those template and simple call: `{{ form_widget(form.update) }}` will throw 500.

My suggestion is to leave it in the FormType and call in the template:
`{% do form.update.setRendered() %} 

This will make sure it isn't rendered by `render_rest()`.

pawbuj:
> It's a BC break - someone could have overridden those template and simple call: `{{ form_widget(form.update) }}` will throw 500.
> 
> My suggestion is to leave it in the FormType and call in the template:
> `{% do form.update.setRendered() %}`
> 
> This will make sure it isn't rendered by `render_rest()`.

You right, but we already agread with this BC break.  https://github.com/ezsystems/ezplatform-admin-ui/pull/847

lserwatka:
It was our mistake @pawbuj , we overlooked it in the CR process. If we can avoid BC here, let's do it more defensive way.

barbaragr:
One question for sub-items sorting:
- when there was Set button, user can change desc/asc sorting by clicking on column's name and then clicking once again Set button, order of sorting was changed
http://g.recordit.co/z293svNO9O.gif
- now if I change sorting order by clicking column's name and then select the same option in dropdown, nothing changes. I need do select another option and then the old one once again. 
http://g.recordit.co/2fCeCOgqmT.gif
Are we ok with that?

ping @SylvainGuittard

SylvainGuittard:
@barbaragr I am fine with this behavior. The user does not change the default sorting order often.
The first time you load the page, the default order will be used. Then the user has full control on the column headers in order to find the content item he/she is looking for.

PR #860, opened 2019-02-28
EZP-30184: As an Editor I want to collapse/expand content preview

adamwojs:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30184
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | yes

## Description

This PR introduces a feature which allows editors to collapse/expand content preview. Default state is configurable via User Setting. 

Front-end implementation based on  https://getbootstrap.com/docs/4.1/components/collapse/ with some CSS tweaks.

## Demo

collapse_content_preview

#### Checklist:
- [x] Move `EzSystems\EzPlatformAdminUi\UserSetting\Setting\CollapseContentPreview` to `ezplatform-user` package (https://github.com/ezsystems/ezplatform-user/pull/5)
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_field-group.scss
+++ b/src/bundle/Resources/public/scss/_field-group.scss
@@ -1,31 +1,76 @@
 .ez-view-rawcontentview {
-
     .ez-raw-content-title {
-        border-bottom: 1px solid $ez-color-secondary;
+        border-bottom: calculateRem(1px) solid $ez-color-secondary;
         color: $ez-color-secondary;
 
         h2 {
-            margin-top: 1rem;
+            margin-top: calculateRem(16px);
             margin-bottom: 0;
-            padding-bottom: .5rem;
-            font-size: 1.125rem;
-            font-weight: 400;  
+            padding-bottom: calculateRem(8px);
+            font-size: calculateRem(18px);
+            font-weight: 400;
+        }
+
+        .ez-content-preview-toggle {
+            color: $ez-color-secondary;
+            position: relative;
+            padding-right: calculateRem(24px);
+            display: inline-block;
+
+            &:hover {
+                text-decoration: none;
+            }
+
+            &:before,
+            &:after {
+                content: '';
+                width: calculateRem(1px);
+                height: calculateRem(7px);
+                background-color: $ez-color-secondary;
+                position: absolute;
+                top: 50%;
+                right: calculateRem(-16px);
+                transition: all 0.2s $ez-admin-transition;
+            }
+
+            &:before {
+                transform: translate(calculateRem(-27px), -50%) rotate(135deg);
+            }
+
+            &:after {
+                transform: translate(calculateRem(-22px), -50%) rotate(45deg);
+            }
+
+            &.collapsed {
+                &:before {
+                    transform: translate(calculateRem(-27px), -20%) rotate(45deg);
+                }
+
+                &:after {
+                    transform: translate(calculateRem(-27px), -80%) rotate(-45deg);
+                }
+            }
         }
 
         .form-inline {
-            padding-bottom: .25rem;
+            padding-bottom: calculateRem(4px);
+
+            &.collapsing {
+                -webkit-transition: none;
+                transition: none;
+                display: none;
+            }
         }
-    }    
+    }
 }
 
 .ez-fieldgroup {
-
     &__name {
         color: $ez-color-secondary;
-        border-bottom: 1px solid $ez-color-secondary;
-        padding-bottom: .5rem;
+        border-bottom: calculateRem(1px) solid $ez-color-secondary;
+        padding-bottom: calculateRem(8px);
         font-weight: 400;
-        font-size: 1.125rem;
+        font-size: calculateRem(18px);
     }
 
     .ez-content-field-name {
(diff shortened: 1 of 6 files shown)


src/bundle/Resources/public/scss/_field-group.scss
```diff
@@ -9,11 +9,47 @@
...
+                    display: inline-block;
+                }
+
+                .ez-icon--expanded {
+                    display: none;
+                }
+            }
+
+            .ez-icon {
+                fill: $ez-color-secondary;
```
dew326: You can use `ez-icon--secondary`

src/bundle/Resources/views/content/content_view_fields.html.twig
```diff
@@ -1,32 +1,44 @@
...
     <div class="ez-raw-content-title d-flex justify-content-between mb-3">
         <h2>
-            {{ 'tab.view.content'|trans()|desc('Content') }}
+            <a data-toggle="collapse" href=".ez-content-preview-collapse" {% if preview_is_collapsed %}class="collapsed"{% endif %}>
+                {{ 'tab.view.preview'|trans()|desc('Preview') }}
+
+                <svg class="ez-icon ez-icon--collapsed">
+                    <use xlink:href="{{ asset('bundles/ezplatformadminui/img/ez-icons.svg') }}#caret-next"></use>
+                </svg>
+                <svg class="ez-icon ez-icon--expanded">
```
dew326: You can use only one icon and do `rotate()` on it. We would be able to add transition animation on rotate and be consistent in whole application.

src/bundle/Resources/views/content/content_view_fields.html.twig
```diff
@@ -1,32 +1,44 @@
 <section class="ez-view-rawcontentview">
+    {% set preview_is_collapsed = ez_user_settings['collapse_content_preview'] %}
     <div class="ez-raw-content-title d-flex justify-content-between mb-3">
         <h2>
-            {{ 'tab.view.content'|trans()|desc('Content') }}
+            <a data-toggle="collapse" href=".ez-content-preview-collapse" {% if preview_is_collapsed %}class="collapsed"{% endif %}>
```
dew326: Please add some class to this `a` element and style it by the class.

src/lib/UserSetting/Setting/CollapseContentPreview.php
```diff
@@ -0,0 +1,124 @@
...
+        return $this->getTranslatedDescription();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDisplayValue(string $storageValue): string
+    {
+        return $this->translator->trans(
+            'settings.collapse_content_preview.value.' . $storageValue,
```
mikadamczyk: If it is possible it is good idea to avoid concatenation of translation key

SylvainGuittard:
@adamwojs Thanks for this feature 👍 
About the caret, could you please reuse the same approach we have for the filters on the search interface? The caret should be thinner.
image

src/bundle/Resources/translations/user_settings.en.xliff
```diff
@@ -76,6 +76,26 @@
         <target state="new">User setting '%identifier%' updated.</target>
         <note>key: user_setting.update.success</note>
       </trans-unit>
+      <trans-unit id="834a92c1b3716e359e63bf81dc433db8dc34ab6c" resname="settings.collapse_content_preview.value.title">
+        <source>Content preview</source>
+        <target state="new">Content preview</target>
+        <note>key: settings.collapse_content_preview.value.title</note>
+      </trans-unit>
+      <trans-unit id="dfa191f1537cc68c074af3595a33154568f2f12e" resname="settings.collapse_content_preview.value.description">
+        <source>Should content preview be collapsed by default?</source>
```
SylvainGuittard:
```suggestion
        <source>Collapse preview by default</source>
```

src/bundle/Resources/translations/user_settings.en.xliff
```diff
@@ -76,6 +76,26 @@
...
+        <target state="new">Should content preview be collapsed by default?</target>
```
SylvainGuittard:
```suggestion
        <target state="new">Collapse preview by default</target>
```

src/lib/UserSetting/Setting/CollapseContentPreview.php
```diff
@@ -0,0 +1,124 @@
...
+        );
+    }
+
+    /**
+     * @return string
+     */
+    private function getTranslatedDescription(): string
+    {
+        return $this->translator->trans(
+            /** @Desc("Should content preview be collapsed by default?") */
```
SylvainGuittard:
```suggestion
            /** @Desc("Collapse preview by default") */
```

lserwatka:
@adamwojs we need a rebase here.

src/bundle/Resources/views/content/content_view_fields.html.twig
adamwojs: Fixed https://github.com/ezsystems/ezplatform-admin-ui/pull/860/commits/74642645add985253a5b9050c2405818fc255b20

src/bundle/Resources/public/scss/_field-group.scss
adamwojs: Fixed https://github.com/ezsystems/ezplatform-admin-ui/pull/860/commits/74642645add985253a5b9050c2405818fc255b20

dew326:
Carret fixed:
collapse-icon

SylvainGuittard:
@dew326 Thanks for the caret!

micszo:
@adamwojs seems translation is missing for `prod`.

micszo:
@SylvainGuittard could you confirm that the behaviour from wireframes in invision is a requirement for this iteration? Namely that after collapsing the preview and going somewhere with Browse the preview should remain collapsed.

SylvainGuittard:
> @SylvainGuittard could you confirm that the behaviour from wireframes in invision is a requirement for this iteration? Namely that after collapsing the preview and going somewhere with Browse the preview should remain collapsed.

@micszo Exactly! When the preview is collapsed, it should remain collapsed (navigation with Browse, Content Tree).

dew326:
So this should be saved in localStorage, not on the user setting.

adamwojs:
Somehow I didn’t catch this behavior from the wireframes. We will process this as follow up story on Monday. @dew326 In this case, the value from the user setting will be the default value (if no value is present in the local storage).

dew326:
This will be confusing to have a configuration in two places. If I will set in a user setting to have collapsed and will expand in one content and now all contents are expanded. I want to change this so I'm going to a user setting and I'm surprised that this should be collapsed.

adamwojs:
Actually, I agree. I will fix it on Monday.

adamwojs:
PR updated with a new solution based on local storage rather than user preferences.

PR #861, opened 2019-02-28
EZP-30185: Expose ContentTree configuration to AdminUI

webhdx:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30185
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/UI/Module/ContentTree/NodeFactory.php
+++ b/src/lib/UI/Module/ContentTree/NodeFactory.php
@@ -35,27 +35,33 @@ final class NodeFactory
     private $maxDepth;
 
     /** @var string[] */
-    private $contentTypeIdentifierIgnoreList;
+    private $allowedContentTypes;
+
+    /** @var string[] */
+    private $ignoredContentTypes;
 
     /**
      * @param \eZ\Publish\API\Repository\SearchService $searchService
      * @param int $displayLimit
      * @param int $childrenLoadMaxLimit
      * @param int $maxDepth
-     * @param array $contentTypeIdentifierIgnoreList
+     * @param array $allowedContentTypes
+     * @param array $ignoredContentTypes
      */
     public function __construct(
         SearchService $searchService,
         int $displayLimit = 20,
         int $childrenLoadMaxLimit = 100,
         int $maxDepth = 10,
-        array $contentTypeIdentifierIgnoreList = []
+        array $allowedContentTypes = [],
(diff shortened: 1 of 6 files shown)


webhdx:
Please note I've changed from using single ignore list to allowed and ignored content types instead.

`allowed_content_types` is an array of allowed content types in Content Tree. This can be set to `'*'` or an empty array which means it will display all content types. In this case you can utilize `ignored_content_types` to narrow down the list.

cc @MagdalenaZuba This is useful information for the doc which was previously described in #854. Please note mentioned changes.

src/lib/UI/Module/ContentTree/NodeFactory.php
```diff
@@ -35,27 +35,33 @@ final class NodeFactory
...
+        array $allowedContentTypes = [],
+        array $ignoredContentTypes = []
     ) {
         $this->searchService = $searchService;
         $this->displayLimit = $displayLimit;
         $this->childrenLoadMaxLimit = $childrenLoadMaxLimit;
         $this->maxDepth = $maxDepth;
-        $this->contentTypeIdentifierIgnoreList = $contentTypeIdentifierIgnoreList;
+        $this->allowedContentTypes = $allowedContentTypes;
+        $this->ignoredContentTypes = $ignoredContentTypes;
```
adamwojs: The number of configuration values which are passed via constructor is growing dangerously so might be good idea to introduce dedicated class representing configuration. 

```php
final class NodeFactorySettings {

    /** @var int **/    
    private $displayLimit;
    /** @var int **/
    private $childrenLoadMaxLimit;
    /** @var int **/
    private $maxDepth;
    /** @var string[] */
    private $allowedContentTypes;
    /** @var string[] */
    private $ignoredContentTypes;

    // ... 
}
```

Applying this approach will also simplify implementation of `EzSystems\EzPlatformAdminUi\UI\Config\Provider\Module\ContentTree`

webhdx: Do you want to create a service from `NodeFactorySettings`? I don't agree with this approach - a service from basically a ValueObject does not convince me. I'd rather pass `ConfigResolver` here and use it to get settings.

PR #863, opened 2019-02-28
Set correct branch alias

mnocon:
Currently builds are failing: https://travis-ci.org/ezsystems/ezplatform-admin-ui/builds/500215494

This is caused by incorrectly set branch alias for tmp_ci_branch - it should be equal to master alias.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
     "extra": {
         "_ezplatform_branch_for_behat_tests": "master",
         "branch-alias": {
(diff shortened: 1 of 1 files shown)


ViniTou:
This worked when we tested on my fork, so :+1:

mnocon:
@lserwatka can we merge this? This fixes AdminUI build.

lserwatka:
Done!

PR #869, opened 2019-03-01
EZP-29802: Inline custom tags support

dew326:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29802
| Requires | https://github.com/ezsystems/ezplatform-richtext/pull/29
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review
- [x] backend

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
@@ -5,11 +5,24 @@
         constructor() {
             this.ezNamespace = 'http://ez.no/namespaces/ezpublish5/xhtml5/edit';
             this.xhtmlNamespace = 'http://www.w3.org/1999/xhtml';
-            this.customTags = Object.keys(global.eZ.adminUiConfig.richTextCustomTags);
+            this.customTags = Object.keys(global.eZ.adminUiConfig.richTextCustomTags).filter(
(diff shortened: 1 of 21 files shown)


lserwatka:
@adamwojs we need to run cs-fixer

src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
```diff
@@ -5,7 +5,12 @@
         constructor() {
             this.ezNamespace = 'http://ez.no/namespaces/ezpublish5/xhtml5/edit';
             this.xhtmlNamespace = 'http://www.w3.org/1999/xhtml';
-            this.customTags = Object.keys(global.eZ.adminUiConfig.richTextCustomTags);
+            this.customTags = Object.keys(global.eZ.adminUiConfig.richTextCustomTags).filter(
```
tischsoic: `customTags` might be confusing, maybe `notInlineCustomTags` or `blockCustomTags`?

tischsoic: But now I see that in other files we have terminology inlineCustomTag/customTag, so OK. :)

PR #875, opened 2019-03-05
[composer] Add missing ezplatform-user dependency

ViniTou:


webhdx:
Looks good. Travis is failing but it's not related (#874).

PR #877, opened 2019-03-05
EZP-30214: Add Content Tree to Bookmarks, Trash & Search

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30214
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

## Screenshots
screen shot 2019-03-08 at 10 39 16

screen shot 2019-03-08 at 10 39 31

screen shot 2019-03-08 at 10 39 45

screen shot 2019-03-08 at 10 40 02

screen shot 2019-03-08 at 10 40 26

screen shot 2019-03-08 at 10 46 43

#### Checklist:
- [x] Ready for Code Review


webhdx:
Can't we move it the container that has left sidebar in it, can we? This would simplify things a lot since we would have it only in one place. If not - we have to put it in separate file and include that file to avoid code duplication.

dew326:
Well, this would require some rework of styling, but I think if we want to have this in multiple places there is a lot of value to put this in the sidebar.

src/bundle/Resources/views/admin/bookmark/list.html.twig
```diff
@@ -115,6 +117,10 @@
     </div>
 {% endblock %}
 
+{% block react_modules %}
```
tischsoic: I moved Content Tree container div to parent twig, but I've left JS in lower templates as I don't want to load them globally.

tischsoic: If anyone has idea how to solve this problem I would readily hear it. :)

tischsoic:
Contnet Tree container moved to left_sidebar - ping @dew326 @webhdx

tischsoic:
ping @webhdx @dew326

webhdx:
Please ignore above revert commit - I accidentally clicked `revert` button which created a revert branch.

PR #881, opened 2019-03-06
EZP-30231: When resizing Content Tree text is accidentally selected

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30231
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

### Problem:
tbrsdrvftb

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_general.scss
+++ b/src/bundle/Resources/public/scss/_general.scss
@@ -156,3 +156,7 @@ select {
 .ez-prevent-click * {
     pointer-events: none !important;
 }
+
+.ez-is-tree-resizing {
+    user-select: none;
+}
--- a/src/bundle/Resources/public/scss/_general.scss
+++ b/src/bundle/Resources/public/scss/_general.scss
@@ -158,5 +158,8 @@ select {
 }
 
 .ez-is-tree-resizing {
+    -webkit-user-select: none;
+    -moz-user-select: none;
+    -ms-user-select: none;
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/public/scss/_view-content.scss
```diff
@@ -31,10 +31,8 @@
...
         min-width: 0;
     }
+}
 
-    &.ez-is-tree-resizing {
-        .ez-location-view-container {
-            pointer-events: none;
-        }
-    }
+.ez-is-tree-resizing {
```
dew326: Seems like wrong file. `general`?

tischsoic:
Moved to `_global.scss` - ping @dew326

src/bundle/Resources/public/scss/_general.scss
```diff
@@ -156,3 +156,7 @@ select {
 .ez-prevent-click * {
     pointer-events: none !important;
 }
+
+.ez-is-tree-resizing {
+    user-select: none;
```
dew326: According to https://caniuse.com/#search=user-select this will work only on chrome. We should add prefixes.

tischsoic: Good catch!

PR #886, opened 2019-03-07
EZEE-2716: Richtext block: tooltip for links can render in way that part of it won't be visible

czechmarcin:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZEE-2716
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/css/alloyeditor/alloyeditor-ez.css
+++ b/src/bundle/Resources/public/css/alloyeditor/alloyeditor-ez.css
@@ -1045,6 +1045,6 @@
   line-height: 1;
 }
 
-.ae-toolbar-styles {
+.alloy-editor-visible {
   z-index: 1;
 }
--- a/src/bundle/Resources/public/css/alloyeditor/alloyeditor-ez.css
+++ b/src/bundle/Resources/public/css/alloyeditor/alloyeditor-ez.css
@@ -1044,3 +1044,7 @@
   font-size: 14px;
   line-height: 1;
 }
+
+.ae-toolbar-styles {
+  z-index: 1;
+}


dew326:
Seems like this could fix other issue: https://jira.ez.no/browse/EZEE-2701 or the fix can be easily change to fix both.

dew326:
From the ticket it looks like this is a partial fix. I underdstand the ticket in way that issue is with position of the toolbar. It should be moved to the left to fit the visible window, but probably @mateuszbieniek should say what is expected by the customer.

mateuszbieniek:
I think this approach is more of a workaround than real fix, as it enables the user to use the feature, but does not resolve all UI issues.

On the other hand, the Customer expects that all toolbar buttons are visible and clickable, which this PR provides and preparing more sofisticated solution will be a lot more complicated and time consuming.

To sum it up: +1 from me.

dew326:
Ok. So we need to check if this fix the other issue with buttons (probably customer will complain on it but he didn't see it yet) at the bottom and we are good to go.

czechmarcin:
Used common class `alloy-editor-visible` instead of `ae-toolbar-styles`. Works on both - anchor and custom tag.

lserwatka:
@dew326 can we proceed with QA here?

src/bundle/Resources/public/css/alloyeditor/alloyeditor-ez.css
```diff
@@ -1044,3 +1044,7 @@
   font-size: 14px;
   line-height: 1;
 }
+
+.alloy-editor-visible {
```
dew326: I don't know this class. Where it's added?

czechmarcin: This class is added for modals with `custom tag` and `anchor`.

czechmarcin: Thanks to this change, the `custom tag`:
- visibility of `custom tag` will have a bigger priority than buttons bar
- if the `custom tag` element exceeds the height limit, overflow appears

dew326: This problem probably will be for other toolbars as well. What do you think to add this to the parent `ae-toolbars` and kill bug once for all?

czechmarcin: Adding `z-index: 1;` to `.ae-toolbars` won't solve the problem.
Child needs this attribute.
Screenshot 2019-03-14 at 10 04 19
Screenshot 2019-03-14 at 10 04 09

czechmarcin:
@barbaragr ATM I cannot reproduce your case. Despite that, my proposition is to apply these commits because:
- this change makes the modal visible instead of hidden under some layers (top and bottom)
- even if we don't have full modal, overflow appears and we can scroll to see the box

micszo:
It works with tag v2.4.2, it doesn't with stable 2.4. The issue is ezplatform-admin-ui-assets version. Tag has 3.2.0, stable branch installs 3.1.0. It's caused by composer.json entry "~3.1.0@dev" (vs. "^3.1.0@dev"). Thanks @dew326 !

lserwatka:
@micszo feel free to make a PR to update assets package version.

lserwatka:
Could you merge it up?

czechmarcin:
done

PR #888, opened 2019-03-08
EZP-30269: [Content Tree] Collapse all button jumps a little bit when scrolling

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30269
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

I was inspired by: https://css-tricks.com/styling-based-on-scroll-position/

Previously we were afraid of poor performance because of frequent updates of DOM element and use of `getBoundingClientRect`. Current solution should be performant.

## Before
ndupy70uqx

## After
8rim84byo5

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.content.tree.js
+++ b/src/bundle/Resources/public/js/scripts/admin.content.tree.js
@@ -2,14 +2,13 @@
     const KEY_CONTENT_TREE_EXPANDED = 'ez-content-tree-expanded';
     const CLASS_CONTENT_TREE_EXPANDED = 'ez-content-tree-container--expanded';
     const CLASS_BTN_CONTENT_TREE_EXPANDED = 'ez-btn--content-tree-expanded';
-    const VIEWPORT_CHANGE_TIMEOUT = 50;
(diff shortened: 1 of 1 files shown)


lserwatka:
While working on something you must provide support for all supported browsers. It is waste of time time fixing it for Chrome only and traveling back in time to fix it in Firefox, Edge, etc.

tischsoic:
This does not solved problem in 100% in Firefox but it is significant improvement and only minor flickering is sometimes visible (at least on my computer).

micszo:
On Firefox when tree is collapsed it's not so bad but when it's expanded it's not only flickering but also jumping.
Please see recording.

ezp-30269-ff-480p

tischsoic:
OK, so on my Firefox it looks way better:

497pdmetga

I guess it is connected with performance of the machine and browser capabilities. We may leave this PR for now and think about better solution. Or maybe merge this as code improvement but not as PR fixing bug?

PR #890, opened 2019-03-10
EZP-30244: Number of words is incorrect when inline custom tag is included

konradoboza:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30244](https://jira.ez.no/browse/EZP-30244), [EZP-30234](https://jira.ez.no/browse/EZP-30234)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Inline entities are now cleared and splitting into words was changed, as relying on single spaces wasn't sufficient.

#### Checklist:
- [x] Ready for Code Review


src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
```diff
@@ -282,7 +284,7 @@
         }
 
         getCountableTags(html) {
-            const allowedTags = ['p:not(.ez-embed-content)', 'li', 'h1', 'h2', 'h3', 'h4', 'h5' , 'h6', 'th', 'td', 'pre'];
+            const allowedTags = ['p:not(.ez-embed-content)', 'li', 'h1', 'h2', 'h3', 'h4', 'h5' , 'h6', 'th', 'td', 'pre', '[data-ezelement=eztemplate]'];
```
dew326: Inline custom tag has `data-ezelement="eztemplateinline"` not `eztemplate`

src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
```diff
@@ -291,10 +293,36 @@
...
+            const removables = ['[data-ezelement=ezconfig]'];
+
+            html.querySelectorAll(extractables.join(','))
+                .forEach((node) => {
+                    node.replaceWith(node.innerText);
+                });
+
+            html.querySelectorAll(removables.join(','))
+                .forEach((node) => {
+                    node.remove();
```
dew326: I believe if you remove some node from this html it will not be stored in textarea and not send to the backend.

konradoboza:
Counting values has been rebuilt, now the script iterates recursively over all children (also nested tags) and verify them against blacklist which contains selectors of unwanted nodes.

TODO:// cleaning from additional white characters.

src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
```diff
@@ -263,38 +263,50 @@
             const counterWrapper = container.parentElement.querySelector('.ez-character-counter');
 
             if (counterWrapper) {
-                const countableTags = this.getCountableTags(editorHtml);
                 const wordWrapper = counterWrapper.querySelector('.ez-character-counter__word-count');
                 const charactersWrapper = counterWrapper.querySelector('.ez-character-counter__character-count');
-                let characterCount = 0;
-                let wordCount = 0;
+                let words = this.getTextNodeValues(editorHtml);
```
dew326: `const`

src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
```diff
@@ -263,38 +263,50 @@
...
-                });
+                wordWrapper.innerText = words.length;
+                charactersWrapper.innerText = words.join(' ').length;
+            }
+        }
 
-                wordWrapper.innerText = wordCount;
-                charactersWrapper.innerText = characterCount;
+        getTextNodeValues(node) {
+            const TEXT_NODE = 3;
```
dew326: If you want to define such constant it should be at the top of a file.

src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
```diff
@@ -263,38 +263,50 @@
...
+            const notAllowedAttributes = ['ezconfig', 'ezvalue'];
+            let values = [];
+
+            const pushValue = (node) => {
+                if (node.nodeType === TEXT_NODE &&
+                    !notAllowedAttributes.includes(node.parentNode.getAttribute('data-ezelement'))
+                ) {
+                    let nodeValue = this.sanitize(node.nodeValue);
```
dew326: `const` and empty line after const

src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
```diff
@@ -263,38 +263,50 @@
...
-                wordWrapper.innerText = wordCount;
-                charactersWrapper.innerText = characterCount;
+        getTextNodeValues(node) {
+            const TEXT_NODE = 3;
+            const notAllowedAttributes = ['ezconfig', 'ezvalue'];
+            let values = [];
+
+            const pushValue = (node) => {
+                if (node.nodeType === TEXT_NODE &&
+                    !notAllowedAttributes.includes(node.parentNode.getAttribute('data-ezelement'))
```
dew326: Not sure if we should iterate over `notAllowedAttributes`. If the element is in not allowed we should skip it. So this check should be in `iterateThroughChildNodes` with the early return when has `ezconfig` attribute (less HTML to go through).

src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
```diff
@@ -263,38 +264,50 @@
...
                 const wordWrapper = counterWrapper.querySelector('.ez-character-counter__word-count');
                 const charactersWrapper = counterWrapper.querySelector('.ez-character-counter__character-count');
-                let characterCount = 0;
-                let wordCount = 0;
+                const words = this.getTextNodeValues(editorHtml);
 
-                countableTags.forEach((tag) => {
-                    const sanitizedText = this.cleanWhiteCharacters(tag.innerText);
+                wordWrapper.innerText = words.length;
+                charactersWrapper.innerText = words.join(' ').length;
```
tischsoic: It seems that, if two words are separated by more than 1 space we count only first space as character. Does it work like this? is this intended?

konradoboza: That's true, this case will be handled separately.

PR #893, opened 2019-03-12
EZEE-2481: Time picker bases on system time ignoring time zone in user settings

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZEE-2481, https://jira.ez.no/browse/EZEE-2480
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
@@ -1,4 +1,4 @@
-(function(global, doc, eZ, flatpickr) {
+(function(global, doc, eZ, moment, flatpickr) {
     const SELECTOR_FIELD = '.ez-field-edit--ezdatetime';
     const SELECTOR_INPUT = '.ez-data-source__input[data-seconds]';
     const SELECTOR_FLATPICKR_INPUT = '.flatpickr-input';
@@ -65,19 +65,21 @@
         time_24hr: true,
         formatDate: (date) => new Date(date).toLocaleString(),
     };
-    const updateInputValue = (sourceInput, date) => {
+    const updateInputValue = (sourceInput, dates) => {
         const event = new CustomEvent(EVENT_VALUE_CHANGED);
 
-        if (!date.length) {
+        if (!dates.length) {
             sourceInput.value = '';
             sourceInput.dispatchEvent(event);
 
             return;
         }
 
-        date = new Date(date[0]);
-        sourceInput.value = Math.floor(date.getTime() / 1000);
+        const selectedDate = moment(dates[0]);
+        const selectedDateWithUserTimezone = selectedDate.tz(eZ.adminUiConfig.timezone, true);
+        const timestamp = Math.floor(selectedDateWithUserTimezone.valueOf() / 1000);
 
+        sourceInput.value = timestamp;
         sourceInput.dispatchEvent(event);
     };
     const clearValue = (sourceInput, flatpickrInstance, event) => {
@@ -91,7 +93,8 @@
         const sourceInput = field.querySelector(SELECTOR_INPUT);
         const flatPickrInput = field.querySelector(SELECTOR_FLATPICKR_INPUT);
         const btnClear = field.querySelector('.ez-data-source__btn--clear-input');
-        const defaultDate = sourceInput.value ? new Date(sourceInput.value * 1000) : null;
+        const defaultDateWithUserTimezone = eZ.helpers.timezone.convertDateToTimezone(sourceInput.value * 1000);
+        const defaultDate = sourceInput.value ? new Date(defaultDateWithUserTimezone.format('YYYY-MM-DD HH:mm:ss')) : null;
         const flatpickrInstance = flatpickr(
             flatPickrInput,
             Object.assign({}, datetimeConfig, {
@@ -109,4 +112,4 @@
     };
 
     datetimeFields.forEach(initFlatPickr);
-})(window, window.document, window.eZ, window.flatpickr);
+})(window, window.document, window.eZ, window.moment, window.flatpickr);
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
@@ -93,8 +93,15 @@
         const sourceInput = field.querySelector(SELECTOR_INPUT);
         const flatPickrInput = field.querySelector(SELECTOR_FLATPICKR_INPUT);
         const btnClear = field.querySelector('.ez-data-source__btn--clear-input');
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
```diff
@@ -65,19 +65,21 @@
...
             sourceInput.value = '';
             sourceInput.dispatchEvent(event);
 
             return;
         }
 
-        date = new Date(date[0]);
-        sourceInput.value = Math.floor(date.getTime() / 1000);
+        const selectedDate = moment(dates[0]);
+        const selectedDateWithUserTimezone = selectedDate.tz(eZ.adminUiConfig.timezone, true);
```
dew326: What `true` parameter does?
Should use the timezone helper.

src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
```diff
@@ -91,7 +93,8 @@
         const sourceInput = field.querySelector(SELECTOR_INPUT);
         const flatPickrInput = field.querySelector(SELECTOR_FLATPICKR_INPUT);
         const btnClear = field.querySelector('.ez-data-source__btn--clear-input');
-        const defaultDate = sourceInput.value ? new Date(sourceInput.value * 1000) : null;
+        const defaultDateWithUserTimezone = eZ.helpers.timezone.convertDateToTimezone(sourceInput.value * 1000);
+        const defaultDate = sourceInput.value ? new Date(defaultDateWithUserTimezone.format('YYYY-MM-DD HH:mm:ss')) : null;
```
dew326: Why this format is hardcoded? Seems like should be from the config.

src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
tischsoic: I didn't use timezone helper, because I wanted to pass `true`, with `true` _tz_ changes timezone without changing hours, minutes, seconds. We need different timezone to correctly calculate timestamp with `valueOf`.

src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
tischsoic: I wanted to get date of type _Date_ with given year, day, month, hour etc. but with browser timezone, which is used in _flatpickr_. 
Ref: https://stackoverflow.com/questions/29132096/momentjs-todate-timezone-gets-reset#answer-29170326

Maybe there is better way of doing this.

dew326: In the helper there is method `formatDate` - will format without timezone as far as I know.

src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
dew326: So the helper should be enhanced to accept the `true` parameter (default: false).

src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
tischsoic: But I would also have to pass `YYYY-MM-DD HH:mm:ss` to this `formatDate` function, because e.g. _userPreferedFullDateFormat_ may not include seconds. Given this, are there any benefits of using `formatDate` from timezone helper?

tischsoic: And for me, downside of `formatDate` is that this formatter may change in the future as it has changed recently: https://github.com/ezsystems/ezplatform-admin-ui/commit/ef673a1c1ac8403c54c2381d9323296af50f2c9b#diff-7176012c51b2ab10ab527e44f38c1544R10
so I don't think it is safe to use it when you want to have full date without timezone. I think that `formatDate` should be used only for formatting date when displaying it to user.

dew326: As I understand this is the date that will be displayed in the input. It should be consistent with other part of the application

src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
tischsoic: Enhance it here? Base branch is 1.4.

dew326: Yes, why not.

tischsoic:
As discussed with @SylvainGuittard and @supriyabhargava, _eztime_ field should not have timezone. For all users, whatever timezone they have, it should show the same value. 

@SylvainGuittard mentioned "Default value: current time" option in _ezdatetime_ in connected PR in different repository: https://github.com/ezsystems/date-based-publisher/pull/92 
I checked how it works and value is pre-filled by backend during creation and on the frontend it is displayed with user timezone, so it should be OK.

I didn't change function responsible for formatting date. There is a problem with _ezdatetime_ and _eztime_ with seconds - my default date format in Demo does not include seconds, hence when date is formatted with format set in user settings it won't display seconds. This problem will also be present here: https://jira.ez.no/browse/EZP-30263 .
This problem was also discussed with @SylvainGuittard.

src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
```diff
@@ -91,13 +94,24 @@
         const sourceInput = field.querySelector(SELECTOR_INPUT);
         const flatPickrInput = field.querySelector(SELECTOR_FLATPICKR_INPUT);
         const btnClear = field.querySelector('.ez-data-source__btn--clear-input');
-        const defaultDate = sourceInput.value ? new Date(sourceInput.value * 1000) : null;
+        const secondsEnabled = sourceInput.dataset.seconds === '1';
+        const formatDate = secondsEnabled ? (date) => date.toLocaleString() : (date) => eZ.helpers.timezone.formatDate(date);
```
tischsoic: Note: As agreed with @SylvainGuittard, if seconds are enabled we fall back to `date.toLocaleString()` because there it high probability that date format from user settings does not include seconds.

tischsoic:
Changed to use only `convertDateToTimezone` - no _moment_, or _tz_ - ping @dew326.

tischsoic:
WIP - during fixing Behat tests

tischsoic:
Behat tests fixed by @mnocon 🙂

lserwatka:
Could you merge it up?

tischsoic:
Merged up:
- master: https://github.com/ezsystems/ezplatform-admin-ui/commit/b4c564713faf9d677102c39ab74035fc5024b3f0

PR #894, opened 2019-03-12
EZP-30164: After removing option from ezselection it it not possible to add more options

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30164
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Instead of counting options, which creates problem with overlapping ids when prior to add option is removed, we use `id` extracted from `name` from last input.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.contenttype.selection.js
+++ b/src/bundle/Resources/public/js/scripts/admin.contenttype.selection.js
@@ -20,7 +20,7 @@
                 return 0;
             }
             const lastInput = container.querySelector(SELECTOR_OPTIONS_LIST).lastElementChild.querySelector('input[type="text"]');
-            const lastId = parseInt(lastInput.name.match(/.+\[(\d)]$/)[1]);
+            const lastId = lastInput.dataset.ezselectionOptionId;
 
             return lastId + 1;
         };
--- a/src/bundle/Resources/public/js/scripts/admin.contenttype.selection.js
+++ b/src/bundle/Resources/public/js/scripts/admin.contenttype.selection.js
@@ -15,12 +15,21 @@
 
             container.querySelector(SELECTOR_BTN_REMOVE)[methodName]('disabled', disabledState);
         };
+        const getNextId = () => {
+            if (countExistingOptions() === 0) {
+                return 0;
+            }
+            const lastInput = container.querySelector(SELECTOR_OPTIONS_LIST).lastElementChild.querySelector('input[type="text"]');
+            const lastId = parseInt(lastInput.name.match(/.+\[(\d)]$/)[1]);
+
+            return lastId + 1;
+        };
         const addOption = () => {
             const template = container.querySelector(SELECTOR_TEMPLATE).innerHTML;
 
             container
                 .querySelector(SELECTOR_OPTIONS_LIST)
-                .insertAdjacentHTML('beforeend', template.replace(NUMBER_PLACEHOLDER, countExistingOptions()));
+                .insertAdjacentHTML('beforeend', template.replace(NUMBER_PLACEHOLDER, getNextId()));
         };
         const removeOptions = () => {
             findCheckedOptions().forEach(element => element.closest(SELECTOR_OPTION).remove());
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/public/js/scripts/admin.contenttype.selection.js
```diff
@@ -15,12 +15,21 @@
 
             container.querySelector(SELECTOR_BTN_REMOVE)[methodName]('disabled', disabledState);
         };
+        const getNextId = () => {
+            if (countExistingOptions() === 0) {
+                return 0;
+            }
+            const lastInput = container.querySelector(SELECTOR_OPTIONS_LIST).lastElementChild.querySelector('input[type="text"]');
+            const lastId = parseInt(lastInput.name.match(/.+\[(\d)]$/)[1]);
```
dew326: This should be stored in data-attribute. The same like @Nattfarinn did in ezmatrix.

ViniTou: @dew326 
fixd

src/bundle/Resources/views/admin/content_type/field_types.html.twig
```diff
@@ -212,7 +212,10 @@
                         class="ezselection-settings-option-checkbox ez-selection-table-checkbox"
                         {{ is_translation ? 'disabled' : '' }}>
                     {{ form_errors(option) }}
-                    {{ form_widget(option, { 'attr': { 'class': 'form-control ml-1 mb-0'}}) }}
+                    {{ form_widget(option, { 'attr': {
+                        'class': 'form-control ml-1 mb-0',
+                        'data-ezselection-option-id': option.vars.name }})
```
dew326: I meant more like adding data-attribute only to the parent `ezselection-settings` like `data-last-id={{ max(form.options|keys) }}`.

src/bundle/Resources/public/js/scripts/admin.contenttype.selection.js
```diff
@@ -15,12 +15,21 @@
 
             container.querySelector(SELECTOR_BTN_REMOVE)[methodName]('disabled', disabledState);
         };
+        const getNextId = () => {
+            if (countExistingOptions() === 0) {
+                return 0;
+            }
+            const lastInput = container.querySelector(SELECTOR_OPTIONS_LIST).lastElementChild.querySelector('input[type="text"]');
```
dew326: This query will be much simpler if you use what is proposed in previous comment

src/bundle/Resources/public/js/scripts/admin.contenttype.selection.js
```diff
@@ -15,12 +15,21 @@
 
             container.querySelector(SELECTOR_BTN_REMOVE)[methodName]('disabled', disabledState);
         };
+        const getNextId = () => {
+            if (countExistingOptions() === 0) {
+                return 0;
+            }
```
dew326: after if there is always empty line

src/bundle/Resources/public/js/scripts/admin.contenttype.selection.js
```diff
@@ -15,12 +15,21 @@
 
             container.querySelector(SELECTOR_BTN_REMOVE)[methodName]('disabled', disabledState);
         };
+        const getNextId = () => {
+            if (countExistingOptions() === 0) {
```
dew326: you can use just `!countExistingOptions()`

src/bundle/Resources/public/js/scripts/admin.contenttype.selection.js
```diff
@@ -15,12 +15,21 @@
...
+                return 0;
+            }
+            const lastInput = container.querySelector(SELECTOR_OPTIONS_LIST).lastElementChild.querySelector('input[type="text"]');
+            const lastId = lastInput.dataset.ezselectionOptionId;
+
+            return lastId + 1;
```
dew326: if the `ezselectionOptionId` is '3' this will return '31'. I believe it's not what you want ;)

ViniTou: Don;t know how I missed parseInt, it was here before :D

lserwatka:
Could you merge it up?

PR #897, opened 2019-03-14
EZP-30107: Name of Content Item is not updated after changing the main language

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30107
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no (as this was not released yet)
| Tests pass?   | yes
| Doc needed?   | no

Following changes in https://github.com/ezsystems/ezpublish-kernel/pull/2563 were needed.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/Form/DataTransformer/ContentTransformer.php
+++ b/src/lib/Form/DataTransformer/ContentTransformer.php
@@ -28,13 +28,13 @@ class ContentTransformer implements DataTransformerInterface
     }
 
     /**
-     * Transforms a domain specific ContentInfo object into a Content's ID.
+     * Transforms a domain specific Content object into a Content's ID.
      *
      * @param \eZ\Publish\API\Repository\Values\Content\Content|null $value
      *
      * @return int|null
      */
-    public function transform($value)
+    public function transform($value): ?int
     {
         if (null === $value) {
             return null;
@@ -62,7 +62,7 @@ class ContentTransformer implements DataTransformerInterface
             return null;
         }
 
-        if (!is_numeric($value)) {
+        if (!filter_var($value, FILTER_VALIDATE_INT)) {
             throw new TransformationFailedException('Expected a numeric string.');
         }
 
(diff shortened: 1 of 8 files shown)


src/lib/Form/DataTransformer/ContentTransformer.php
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
+     */
+    public function reverseTransform($value): ?Content
+    {
+        if (empty($value)) {
+            return null;
+        }
+
+        if (!is_numeric($value)) {
```
adamwojs: It's not a big deal, but better choice for `int` validation will be (filter_var)[http://php.net/manual/en/function.filter-var.php] or (ctype_digit)[http://php.net/manual/en/function.ctype-digit.php]  `is_numeric` also returns `true` for '3e4', '12.6' etc.

src/lib/Form/DataTransformer/ContentTransformer.php
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * Transforms a domain specific ContentInfo object into a Content's ID.
+     *
+     * @param \eZ\Publish\API\Repository\Values\Content\Content|null $value
+     *
+     * @return int|null
+     */
+    public function transform($value)
```
adamwojs:
```suggestion
    public function transform($value): ?int
```

src/lib/Form/Type/Content/ContentType.php
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * @param \eZ\Publish\API\Repository\ContentService $contentService
+     */
+    public function __construct(ContentService $contentService)
+    {
+        $this->contentService = $contentService;
+    }
+
+    public function buildForm(FormBuilderInterface $builder, array $options)
```
adamwojs: Nitpick: Missing docblock (also applicable in other places)

PR #900, opened 2019-03-14
EZEE-2718: Adjust tests for redirects

mnocon:
After https://jira.ez.no/browse/EZEE-2718 the behaviour after Trashing a content item has changed:
Before:
Trashing a Content Item that has a Landing Page as a parent redirects to the Parent in Content View
After:
Trashing a Content Item that has a Landing Page as a parent redirects to the Parent:
- ezplatform: in Content View
- ezplatform-ee: in Page Builder

This is a bit tricky, as we don't have a mechanism right now to deal with logic like this.

Possible solutions:
1) the proposed one - adding an IF based on environment

I'm not a fan of this, but it works - we need to think of a better approach for future uses.

2) Defining the same step definition in two Contexts and specifying the correct one in behat_suites.yml

I've decided against at as it does not work for use cases like "I want to run AdminUI suite on ezplatform-ee (we would have to define the suite twice, with the different Context being the only difference)

3) Creating two scenarios, tagging it with `ezplatform` and `ezplatform-ee` and filtering it in suites: 

Again, this does not work for use case "I want to run AdminUI suite on ezplatform-ee"

4) Creating two scenarios, tagging it with `ezplatform` and `ezplatform-ee` and filtering in an BeforeScenario hook

This works, but only partially - the ezplatform-ee scenario might require steps that are not defined in any AdminUI Context, so it will fail when running with --strict flag because of undefined steps

5) Create a Behat extension that will allow to skip tests marked with a certain tag (for example "ezplatform" when running on "ezplatform-ee") and will not mark undefined steps in these Scenarios as missing

I've an idea how to do it, but it's not something I can do right now.

### DIFF
--- a/src/lib/Behat/Helper/Hooks.php
+++ b/src/lib/Behat/Helper/Hooks.php
@@ -17,13 +17,6 @@ class Hooks extends RawMinkContext
 
     /** @BeforeScenario
      */
-    public function restartSessionBeforeScenario()
-    {
-        $this->getSession()->restart();
-    }
-
-    /** @BeforeScenario
-     */
     public function setInstallTypeBeforeScenario()
     {
         $env = new Environment($this->getContainer());
--- a/features/ContentCreation.feature
+++ b/features/ContentCreation.feature
@@ -256,7 +256,7 @@ Scenario: Content can be moved to trash from non-root location
   Scenario Outline: Content can be moved to trash from root location
     Given I navigate to content "<itemName>" of type "<itemType>" in root path
     When I send content to trash
-    Then there's no "<itemType>" "<itemName>" on Sub-items list of root
+    Then I should be redirected to root in default view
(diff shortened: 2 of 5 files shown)


src/lib/Behat/Helper/Hooks.php
```diff
@@ -15,13 +15,6 @@ class Hooks extends RawMinkContext
 {
     use KernelDictionary;
 
-    /** @BeforeScenario
```
mnocon: I've decided to use this opportunity to remove this, I don't think we need it and it causes failures from time to time.

lserwatka:
@mnocon is this ready for merge?

mnocon:
@lserwatka yes, this one can be merged

lserwatka:
Great work! Could you merge it up?

mnocon:
Done

PR #909, opened 2019-03-19
EZP-30263 : User date format don't apply to date fields in content item update and preview

pawbuj:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30263](https://jira.ez.no/browse/EZP-30263)
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Adds into ez_full_datetime, ez_full_date, ez_full_time in ezdate_fields.

Warning, this PR can be merged only if https://github.com/ezsystems/ezplatform-user/pull/26 is merged 

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
@@ -65,6 +65,7 @@
     const datetimeConfig = {
         enableTime: true,
         time_24hr: true,
+        formatDate: (date) => eZ.helpers.timezone.formatFullDateTime(date, null),
     };
     const updateInputValue = (sourceInput, dates) => {
         const event = new CustomEvent(EVENT_VALUE_CHANGED);
@@ -95,7 +96,6 @@
         const flatPickrInput = field.querySelector(SELECTOR_FLATPICKR_INPUT);
         const btnClear = field.querySelector('.ez-data-source__btn--clear-input');
         const secondsEnabled = sourceInput.dataset.seconds === '1';
-        const formatDate = secondsEnabled ? (date) => date.toLocaleString() : (date) => eZ.helpers.timezone.formatDate(date);
         let defaultDate = null;
 
         if (sourceInput.value) {
@@ -111,7 +111,6 @@
                 onChange: updateInputValue.bind(null, sourceInput),
                 defaultDate,
                 enableSeconds: secondsEnabled,
-                formatDate,
             })
         );
 
--- a/src/bundle/Resources/public/js/scripts/helpers/timezone.helper.js
+++ b/src/bundle/Resources/public/js/scripts/helpers/timezone.helper.js
@@ -13,10 +13,10 @@
 
         return moment(date).formatICU(format);
     };
-    const formatFullDateTime = (date, timezone = null, format = userPreferedFullDateTimeFormat) => {
+    const formatFullDateTime = (date, timezone = userPreferedTimezone, format = userPreferedFullDateTimeFormat) => {
         return formatDate(date, timezone, format);
     };
-    const formatShortDateTime = (date, timezone = null, format = userPreferedShortDateTimeFormat) => {
+    const formatShortDateTime = (date, timezone = userPreferedTimezone, format = userPreferedShortDateTimeFormat) => {
         return formatDate(date, timezone, format);
     };
 
--- a/src/bundle/Resources/public/js/scripts/helpers/timezone.helper.js
+++ b/src/bundle/Resources/public/js/scripts/helpers/timezone.helper.js
@@ -1,29 +1,77 @@
 (function(global, doc, eZ, moment) {
     const userPreferedTimezone = eZ.adminUiConfig.timezone;
-    const userPreferedFullDateFormat = eZ.adminUiConfig.dateFormat.full;
-    const userPreferedShortDateFormat = eZ.adminUiConfig.dateFormat.short;
+    const userPreferedFullDateTimeFormat = eZ.adminUiConfig.dateFormat.full_datetime;
+    const userPreferedFullDateFormat = eZ.adminUiConfig.dateFormat.full_date;
+    const userPreferedFullTimeFormat = eZ.adminUiConfig.dateFormat.full_time;
+    const userPreferedShortDateTimeFormat = eZ.adminUiConfig.dateFormat.short_datetime;
+    const userPreferedShortDateFormat = eZ.adminUiConfig.dateFormat.short_date;
+    const userPreferedShortTimeFormat = eZ.adminUiConfig.dateFormat.short_time;
 
     const convertDateToTimezone = (date, timezone = userPreferedTimezone, forceSameTime = false) => {
         return moment(date).tz(timezone, forceSameTime);
     };
-    const formatDate = (date, format = userPreferedFullDateFormat) => {
+    const formatDate = (date, format, timezone = null) => {
+        if (timezone) {
+            date = convertDateToTimezone(date, timezone);
+        }
+
         return moment(date).formatICU(format);
     };
-    const formatDateWithTimezone = (date, timezone = userPreferedTimezone, format = userPreferedFullDateFormat) => {
-        return formatDate(convertDateToTimezone(date, timezone), format);
+    const formatFullDateTime = (date, timezone = null, format = userPreferedFullDateTimeFormat) => {
+        return formatDate(date, format, timezone);
     };
-    const formatShortDate = (date, format = userPreferedShortDateFormat) => {
-        return formatDate(date, format);
+    const formatFullDate = (date, timezone = null, format = userPreferedFullDateFormat) => {
+        return formatDate(date, format, timezone);
     };
-    const formatShortDateWithTimezone = (date, timezone = userPreferedTimezone, format = userPreferedShortDateFormat) => {
-        return formatDateWithTimezone(date, timezone, format);
+    const formatFullTime = (date, timezone = null, format = userPreferedFullTimeFormat) => {
+        return formatDate(date, format, timezone);
+    };
+    const formatShortDateTime = (date, timezone = null, format = userPreferedShortDateTimeFormat) => {
+        return formatDate(date, format, timezone);
+    };
+    const formatShortDate = (date, timezone = null, format = userPreferedShortDateFormat) => {
+        return formatDate(date, format, timezone);
+    };
+    const formatShortTime = (date, timezone = null, format = userPreferedShortimeFormat) => {
+        return formatDate(date, format, timezone);
+    };
+
+    const deprecatedFormatDate = (date, format = userPreferedFullDateTimeFormat) => {
+        console.warn('[DEPRECATED] formatDate function is deprecated');
+        console.warn('[DEPRECATED] it will change behaviour from ezplatform-admin-ui 2.0');
+        console.warn('[DEPRECATED] use formatFullDateTime instead');
+
+        return formatFullDateTime(date, null, format);
+    };
+    const deprecatedFormatShortDate = (date, format = userPreferedShortDateTimeFormat) => {
+        console.warn('[DEPRECATED] formatShortDate function is deprecated');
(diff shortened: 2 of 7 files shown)


src/bundle/Resources/views/fieldtypes/preview/content_fields.html.twig
```diff
@@ -80,7 +76,7 @@
 {% block ezdate_field %}
 {% spaceless %}
     {% if not ez_is_field_empty( content, field ) %}
-        {% set field_value = field.value.date|localizeddate( 'short', 'none', parameters.locale, 'UTC' ) %}
+        {% set field_value = field.value.date|ez_full_date %}
```
webhdx: UTC has to be forced here to avoid timezone issues.

src/bundle/Resources/views/fieldtypes/preview/content_fields.html.twig
```diff
@@ -89,11 +85,7 @@
 {% block eztime_field %}
 {% spaceless %}
     {% if not ez_is_field_empty( content, field ) %}
-        {% if fieldSettings.useSeconds %}
-            {% set field_value = field.value.time|localizeddate( 'none', 'medium', parameters.locale, 'UTC' ) %}
-        {% else %}
-            {% set field_value = field.value.time|localizeddate( 'none', 'short', parameters.locale, 'UTC' ) %}
-        {% endif %}
+        {% set field_value = field.value.time|ez_full_time %}
```
webhdx: Same as above - UTC has to be forced.

src/bundle/Resources/views/fieldtypes/preview/content_fields.html.twig
```diff
@@ -67,11 +67,7 @@
 {% block ezdatetime_field %}
 {% spaceless %}
     {% if not ez_is_field_empty( content, field ) %}
-        {% if fieldSettings.useSeconds %}
-            {% set field_value = field.value.value|localizeddate('short', 'medium', null, ez_user_settings['timezone']) %}
-        {% else %}
-            {% set field_value = field.value.value|localizeddate('short', 'short', null, ez_user_settings['timezone']) %}
-        {% endif %}
+        {% set field_value = field.value.value|ez_full_datetime %}
```
webhdx: We have an issue here as you can set `display seconds` in content type. This change will make the setting not really useful in Admin interface. This is something @SylvainGuittard should take a look at.

SylvainGuittard: This topic has been discussed with @tischsoic 
The idea was to force the rendering of seconds if the field has `use seconds`.

pawbuj: @SylvainGuittard we need to decide what we want to display, seconds or format chosen by the user in settings.

SylvainGuittard: @pawbuj format chosen by the user + seconds if the field has `use seconds` checked.

pawbuj: @SylvainGuittard there some technical issues here I have sent you a direct message.

src/bundle/Resources/views/fieldtypes/preview/content_fields.html.twig
```diff
@@ -69,14 +69,19 @@
...
+        {% endif %}
     {% endif %}
 {% endspaceless %}
 {% endblock %}
 
 {% block ezdate_field %}
 {% spaceless %}
     {% if not ez_is_field_empty( content, field ) %}
-        {% set field_value = field.value.date|ez_full_date %}
+        {% set field_value = field.value.date|ez_full_date('UTC') %}
```
tischsoic: We will be displaying ezdate in UTC, regardless of user time zone from settings?

pawbuj: @tischsoic same as in the prev implementation, I do not change this behaviour

tischsoic: Now it will be inconsistent with date picked in edit mode considering last changes: https://github.com/ezsystems/ezplatform-admin-ui/pull/893
Or maybe we are going to revert those changes?
ping @SylvainGuittard

pawbuj: @webhdx

SylvainGuittard: It cannot be `UTC`. The editor will be confused.

webhdx: Date should not be affected by timezone. When you select `22.03.2019` and you change the timezone you still expect to see `22.03.2019`. We had this issue in the past and this is why we want to display this date in UTC as we store timestamp in the database.

webhdx: Same thing for `eztime` - once `15:00` is selected it should be `15:00` even when user changed timezone in the settings. Basically only `ezdatetime` is timezone dependent.

lserwatka: I agree with @webhdx especially that this is a fieldtype value not a regular UI date or time

SylvainGuittard: @lserwatka @webhdx I understand your point. But the date picker depends on the user preferences and the time zone can be different than UTC. So the date in preview might be different from the date the user selected. Or did I miss something?

lserwatka: Let's discuss this in private on Monday.

tischsoic: I was thinking about it and it actully makes sense for me to have _eztime_ and _ezdate_ always in UTC.

SylvainGuittard: After internal discussion, `ezdate` and `eztime` should use UTC. 
Thanks guys.

src/bundle/Resources/public/js/scripts/helpers/timezone.helper.js
```diff
@@ -1,29 +1,77 @@
 (function(global, doc, eZ, moment) {
     const userPreferedTimezone = eZ.adminUiConfig.timezone;
-    const userPreferedFullDateFormat = eZ.adminUiConfig.dateFormat.full;
-    const userPreferedShortDateFormat = eZ.adminUiConfig.dateFormat.short;
+    const userPreferedFullDateTimeFormat = eZ.adminUiConfig.dateFormat.full_datetime;
+    const userPreferedFullDateFormat = eZ.adminUiConfig.dateFormat.full_date;
+    const userPreferedFullTimeFormat = eZ.adminUiConfig.dateFormat.full_time;
+    const userPreferedShortDateTimeFormat = eZ.adminUiConfig.dateFormat.short_datetime;
+    const userPreferedShortDateFormat = eZ.adminUiConfig.dateFormat.short_date;
+    const userPreferedShortTimeFormat = eZ.adminUiConfig.dateFormat.short_time;
```
dew326: all keys in `adminUiConfig` should be camelCase

src/bundle/Resources/public/js/scripts/helpers/timezone.helper.js
```diff
@@ -1,29 +1,77 @@
...
+        }
+
         return moment(date).formatICU(format);
     };
-    const formatDateWithTimezone = (date, timezone = userPreferedTimezone, format = userPreferedFullDateFormat) => {
-        return formatDate(convertDateToTimezone(date, timezone), format);
+    const formatFullDateTime = (date, timezone = null, format = userPreferedFullDateTimeFormat) => {
+        return formatDate(date, format, timezone);
+    };
+    const formatFullDate = (date, timezone = null, format = userPreferedFullDateFormat) => {
```
dew326: What is the reason for having this method? It's not used anywhere, there are more methods like this here.

src/bundle/Resources/public/js/scripts/helpers/timezone.helper.js
```diff
@@ -1,29 +1,77 @@
...
+    const formatShortDate = (date, timezone = null, format = userPreferedShortDateFormat) => {
+        return formatDate(date, format, timezone);
     };
-    const formatShortDateWithTimezone = (date, timezone = userPreferedTimezone, format = userPreferedShortDateFormat) => {
-        return formatDateWithTimezone(date, timezone, format);
+    const formatShortTime = (date, timezone = null, format = userPreferedShortimeFormat) => {
+        return formatDate(date, format, timezone);
+    };
+
+    const deprecatedFormatDate = (date, format = userPreferedFullDateTimeFormat) => {
```
dew326: Not sure if deprecation should change the way of displaying date.

src/bundle/Resources/public/js/scripts/helpers/timezone.helper.js
```diff
@@ -1,29 +1,77 @@
...
+        if (timezone) {
+            date = convertDateToTimezone(date, timezone);
+        }
+
         return moment(date).formatICU(format);
     };
-    const formatDateWithTimezone = (date, timezone = userPreferedTimezone, format = userPreferedFullDateFormat) => {
-        return formatDate(convertDateToTimezone(date, timezone), format);
+    const formatFullDateTime = (date, timezone = null, format = userPreferedFullDateTimeFormat) => {
+        return formatDate(date, format, timezone);
```
dew326: Can we also unify the parameters order? Here is `date, format, timezone` but in the wrapper function is `date, timezone, format`

src/bundle/Resources/public/js/scripts/helpers/timezone.helper.js
pawbuj: Those function will be added to `helpers.timezone` in 2.0 , currently it is not possible because of a name conflict with deprecated function `formatShortDate`

dew326: So this is the dead code? No point of having this in LTS 2.5.

src/bundle/Resources/public/js/scripts/helpers/timezone.helper.js
```diff
@@ -1,29 +1,61 @@
...
 
     eZ.addConfig('helpers.timezone', {
         convertDateToTimezone,
-        formatDate,
-        formatShortDate,
-        formatDateWithTimezone,
-        formatShortDateWithTimezone,
+        formatFullDateTime,
+        formatShortDateTime,
+        formatDate: deprecatedFormatDate,
```
dew326: What about other bundles which are using those deprecated methods? We should correct them to use not deprecated one

pawbuj: there you go 
https://github.com/ezsystems/ezplatform-page-builder/pull/327
https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/173

dew326: I think there is also one in data-based-publisher.

pawbuj: https://github.com/ezsystems/date-based-publisher/pull/97

src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js
```diff
@@ -63,7 +63,7 @@
     const datetimeConfig = {
         enableTime: true,
         time_24hr: true,
-        formatDate: (date) => new Date(date).toLocaleString(),
+        formatDate: (date) => eZ.helpers.timezone.formatFullDateTime(date),
```
dew326: This one will be converted but other in flatpickr are not. Intentional?

dew326: On the second thought, all should be converted to user timezone, this is displayed date.

PR #916, opened 2019-03-20
EZP-30219: Message displayed for content item visibility should be be…

pawbuj:
…low content title

| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30219](https://jira.ez.no/browse/EZP-30219)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Moves message bar below the content title

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_header.scss
+++ b/src/bundle/Resources/public/scss/_header.scss
@@ -3,6 +3,7 @@
 
     .container {
     	position: relative;
+        overflow: auto;
     }
 
     .breadcrumb {
--- a/src/bundle/Resources/views/parts/page_title.html.twig
+++ b/src/bundle/Resources/views/parts/page_title.html.twig
@@ -1,5 +1,5 @@
 {% if title is defined %}
-    <div class="ez-page-title pt-3 pb-3 px-4">
+    <div class="ez-page-title py-3 px-4">
         <h1 class="ez-page-title__content-item">
         {% if iconName is defined%}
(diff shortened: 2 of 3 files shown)


src/bundle/Resources/public/scss/_header.scss
```diff
@@ -3,6 +3,7 @@
 
     .container {
     	position: relative;
+        overflow: auto;
```
tischsoic:
```suggestion
       overflow: auto;
```

src/bundle/Resources/views/parts/page_title.html.twig
```diff
@@ -1,5 +1,5 @@
 {% if title is defined %}
-    <div class="ez-page-title pt-3 pb-4 px-4">
+    <div class="ez-page-title pt-3 pb-3 px-4">
```
tischsoic:
```suggestion
    <div class="ez-page-title py-3 px-4">
```

> y - for classes that set both *-top and *-bottom

(https://getbootstrap.com/docs/4.0/utilities/spacing/)

src/bundle/Resources/public/scss/_header.scss
tischsoic: Discussed with @pawbuj - Github is showing me additional space, which does not exist in code.
Everything is OK. :)

PR #930, opened 2019-03-26
EZP-30346: Broken notifications with html

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30346
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | yes

https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/170

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.notifications.js
+++ b/src/bundle/Resources/public/js/scripts/admin.notifications.js
@@ -3,14 +3,19 @@
     const notifications = JSON.parse(notificationsContainer.dataset.notifications);
     const template = notificationsContainer.dataset.template;
     const addNotification = ({ detail }) => {
-        const { onShow, label, message } = detail;
+        const { onShow, label, message, rawPlaceholdersMap } = detail;
         // @TODO: Unify 'error' and 'danger' label in 3.0.
         const configKey = label === 'danger' ? 'error' : label;
         const config = eZ.adminUiConfig.notifications[configKey];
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/public/js/scripts/admin.notifications.js
```diff
@@ -3,14 +3,19 @@
     const notifications = JSON.parse(notificationsContainer.dataset.notifications);
     const template = notificationsContainer.dataset.template;
     const addNotification = ({ detail }) => {
-        const { onShow, label, message } = detail;
+        const { onShow, label, message, rawPlaceholdersMap } = detail;
```
dew326: Probably will fail if backend notification occurs. It needs the default value `{}`

dew326:
Open question: Shouldn't this part go to 1.3 as the issue was targeted 1.3 and the problem is there? But on the other hand, we don't have HTML in notifications in 1.3. cc @lserwatka

lserwatka:
No, if we don't have it in 1.3, we should go to 1.4.

lserwatka:
@ViniTou what is needed here?

ViniTou:
@lserwatka 
@dew326 

final round of review (I have fixed default value) and QA.

lserwatka:
Could you merge it up?

PR #931, opened 2019-03-26
EZP-30366: Custom plugins for RichText Editor

SerheyDolgushev:
| Question      | Answer
| ------------- | ---
| Tickets       | [https://jira.ez.no/browse/EZP-30366](EZP-30366)
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | yes

This change allows using custom plugins in Alloy Editor. Just add the following configs:
```
ezrichtext:
    alloy_editor:
        extra_plugins: [myCustomPlugin1, myCustomPlugin2]
```
Also, you would need to implement `myCustomPlugin1` and `myCustomPlugin2` AlloyEditor (CK Editor) plugin.

#### Checklist:
- [x] Ready for Code Review


src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
```diff
@@ -43,6 +43,7 @@
                     };
                 }
             );
+            this.editorPlugins = global.eZ.adminUiConfig.editorPlugins;
```
dew326: The `editorPlugins` key is very generic, I would change this to `richtTextEditorPlugins`.

SerheyDolgushev:
@dew326 it is done. Can you please review one more time?

SerheyDolgushev:
@webhdx Configuration Parser is moved to RichText. Please check https://github.com/ezsystems/ezplatform-richtext/pull/31.

PR #942, opened 2019-04-01
Remove Alloy Editor dist folder, as it is not used because of Webpack Encore

SerheyDolgushev:
| Question      | Answer
| ------------- | ---
| Tickets       | N/A
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

There is no need to have `dist` folder for Alloy Editor.

#### Checklist:
- [x] Ready for Code Review


lserwatka:
@dew326 is this valid?

dew326:
Yes, this is valid for master, in 2.5 we need to keep those files.

SerheyDolgushev:
@dew326 sorry, you mean we **do not** need to keep those files. Please correct me if I'm wrong.

dew326:
We need those files in 2.5 for backward compatibility, someone could include AlloyEditor in own twig as we had in 2.4 in Richtext block in Page Builder. If we remove those files in 2.5 we would break their project. 

We can safely remove it from 3.0.

SerheyDolgushev:
@dew326 you are correct. Wasnt thinking about that. Does it mean we should close this PR for now?

lserwatka:
Yes, and we will take over on removing it later from 3.0 release.

PR #950, opened 2019-04-08
EZP-29694: Draft conflict does not work as expected for a content item with multiple translations

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29694
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | **yes**
| Tests pass?   | yes
| Doc needed?   | yes

This, together with https://github.com/ezsystems/ezpublish-kernel/pull/2598 should align how adminUI behaves comparing to ezPublish. 
This ticket should also be solved by this:
https://jira.ez.no/browse/EZP-29691

BC:
Additional route parameter was needed.
not sure if `Specifications` are under BC though

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/sidebar/btn/location.edit.js
+++ b/src/bundle/Resources/public/js/scripts/sidebar/btn/location.edit.js
@@ -12,12 +12,10 @@
         form.submit();
         $('#version-draft-conflict-modal').modal('hide');
     };
-    const redirectToUserEdit = () => {
+    const redirectToUserEdit = (languageCode) => {
         const versionNo = form.querySelector('#user_edit_version_info_version_no').value;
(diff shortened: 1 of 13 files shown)


src/bundle/Resources/public/js/scripts/sidebar/btn/location.edit.js
```diff
@@ -49,6 +48,10 @@
         attachModalListeners(wrapper);
     };
     const changeHandler = () => {
+        const checkedBtn = btns.find((btn) => btn.checked);
```
dew326: What do you think about small optimization?
You can get the changed button from the event - `event.currentTarget`
Also, `languageCode` can be passed this to the `redirectToUserEdit` method.

PR #953, opened 2019-04-09
EZP 29094: Cannot create user from Content on the Fly

pawbuj:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-29094](https://jira.ez.no/browse/EZP-29094)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

bugfix

#### Checklist:
- [x] Ready for Code Review


src/bundle/Controller/ContentOnTheFlyController.php
```diff
@@ -70,18 +86,24 @@ public function createContentAction(
         ContentType $contentType,
         Location $parentLocation
     ) {
+        if ($contentType->identifier == 'user') {
```
webhdx: It's not going to work. You have go rely on the fact that the content type has `ezuser` fieldtype or use AdminUI configuration which tells you what content types are reserved for users. You can ask @mikadamczyk or @ViniTou about the second approach.

src/bundle/Resources/views/content/locationview.html.twig
```diff
@@ -158,3 +158,7 @@
 {% block javascripts %}
     {{ encore_entry_script_tags('ezplatform-admin-ui-location-view-js', null, 'ezplatform') }}
 {% endblock %}
+
+{% block meta %}
+    <meta name="LocationID" content="{{ location.id }}" />
```
dew326: Seems like unused?

pawbuj: It is used by the JS from admin-ui-modules https://github.com/ezsystems/ezplatform-admin-ui-modules/blob/88de2f7b8f806d0e90430a06523150c5aef63437/src/modules/universal-discovery/components/content-creator/content.creator.component.js#L33

src/bundle/Controller/ContentOnTheFlyController.php
ViniTou: \EzSystems\EzPlatformAdminUi\Specification\ContentType\ContentTypeIsUser

src/bundle/Resources/views/content/locationview.html.twig
dew326: But this should be in special COTF template not in locationview.

src/bundle/Controller/ContentOnTheFlyController.php
```diff
@@ -99,6 +128,40 @@ public function createContentAction(
...
+        string $languageCode,
+        ContentType $contentType,
+        Location $parentLocation
+    ) {
+        $language = $this->languageService->loadLanguage($languageCode);
+        $parentGroup = $this->userService->loadUserGroup($parentLocation->contentId);
+
+        $data = (new UserCreateMapper())->mapToFormData($contentType, [$parentGroup], [
+                'mainLanguageCode' => $language->languageCode,
+            ]);
```
webhdx: Indentation here are invalid. Please run cs fixer.

src/bundle/Controller/ContentOnTheFlyController.php
```diff
@@ -99,6 +128,40 @@ public function createContentAction(
         ]);
     }
 
+    public function createUserAction(
+        Request $request,
+        string $languageCode,
+        ContentType $contentType,
+        Location $parentLocation
+    ) {
```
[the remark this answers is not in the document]
pawbuj: can not add it, because the function can return two types :(

src/bundle/Resources/views/content/content_on_the_fly/content_create_on_the_fly.html.twig
```diff
@@ -29,9 +29,11 @@
 
 {% block form_fields %}
     {{ parent() }}
-    {{ form_widget(form.publish, {'attr': {'hidden': 'hidden'}}) }}
-    {{ form_widget(form.saveDraft, {'attr': {'hidden': 'hidden'}}) }}
-    {{ form_widget(form.cancel, {'attr': {'hidden': 'hidden'}}) }}
+    {% if form.publish is defined %}{{ form_widget(form.publish, {'attr': {'hidden': 'hidden'}}) }}{% endif %}
+    {% if form.saveDraft is defined %}{{ form_widget(form.saveDraft, {'attr': {'hidden': 'hidden'}}) }}{% endif %}
+    {% if form.cancel is defined %}{% do form.cancel.setRendered() %}{% endif %}
```
dew326: Why `cancel` should not be rendered? I think this one is used to discard draft?

src/bundle/Controller/ContentOnTheFlyController.php
```diff
@@ -70,6 +93,10 @@ public function createContentAction(
         ContentType $contentType,
         Location $parentLocation
     ) {
+        if ((new ContentTypeIsUser($this->userContentTypeIdentifier))->isSatisfiedBy($contentType)) {
+            return $this->createUserAction($request, $languageCode, $contentType, $parentLocation);
+        }
+
         $language = $this->languageService->loadLanguage($languageCode);
```
mikadamczyk: We can move this line before it and then pass Language object to createUserAction method. Then it is not needed to load language in createUserAction method. I know that we are using language code, but I think we should start use object instead in new methods if it is possible.

src/bundle/Controller/ContentOnTheFlyController.php
```diff
@@ -178,4 +241,4 @@ private function createContentCreateStruct(
 
         return $contentCreateStruct;
     }
-}
+}
```
mikadamczyk: No newline

src/bundle/Resources/views/content/content_on_the_fly/content_create_on_the_fly.html.twig
webhdx: Yeah, `cancel` should be rendered. `saveDraft` button should not be a part of this form if you do what I proposed in https://github.com/ezsystems/ezplatform-admin-ui/pull/953#discussion_r274301870

pawbuj: `cancel` is also used in `UserCreateType` and there is no option to disable it  by options

PR #966, opened 2019-04-12
EZP-30580: Upgrade `ezsystems/ezplatform-admin-ui` to use Symfony 4 components

webhdx:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30580
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Tests pass?   | yes
| Doc needed?   | no


webhdx:
I tried to provide descriptive commit names so you can check what changed without digging deep into all small changes. Tests are passing (CS and Unit tests), Behat is failing but this is due to unresolved dependencies which is caused by using dev branches for `ezsystems/*` packages. It's okay because these behat tests are passing on meta repository.

PR #968, opened 2019-04-12
EZP-30417: Add notifications to the password reset view

czechmarcin:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30417](https://jira.ez.no/browse/EZP-30417)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Added notifications to the password reset view in order to show validation errors.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.notifications.js
+++ b/src/bundle/Resources/public/js/scripts/admin.notifications.js
@@ -6,10 +6,10 @@
         const { onShow, label, message, rawPlaceholdersMap = {} } = detail;
         // @TODO: Unify 'error' and 'danger' label in 3.0.
         const configKey = label === 'danger' ? 'error' : label;
-        const config = eZ.adminUiConfig.notifications[configKey];
+        const config = eZ.adminUiConfig ? eZ.adminUiConfig.notifications[configKey] : null;
         const timeout = config ? config.timeout : 0;
         const container = doc.createElement('div');
-        let finalMessage = eZ.helpers.text.escapeHTML(message);
+        let finalMessage = eZ.helpers.text ? eZ.helpers.text.escapeHTML(message) : message;
 
         Object.entries(rawPlaceholdersMap).forEach(([placeholder, rawText]) => {
             finalMessage = finalMessage.replace(`{{ ${placeholder} }}`, rawText);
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/public/js/scripts/admin.notifications.js
```diff
@@ -6,10 +6,10 @@
         const { onShow, label, message, rawPlaceholdersMap = {} } = detail;
         // @TODO: Unify 'error' and 'danger' label in 3.0.
         const configKey = label === 'danger' ? 'error' : label;
-        const config = eZ.adminUiConfig.notifications[configKey];
+        const config = eZ.adminUiConfig ? eZ.adminUiConfig.notifications[configKey] : null;
```
dew326: I don't think this is a good approach. This timeout for notifications should be respected.

src/bundle/Resources/public/js/scripts/admin.notifications.js
```diff
@@ -6,10 +6,10 @@
...
         const timeout = config ? config.timeout : 0;
         const container = doc.createElement('div');
-        let finalMessage = eZ.helpers.text.escapeHTML(message);
+        let finalMessage = eZ.helpers.text ? eZ.helpers.text.escapeHTML(message) : message;
```
dew326: This was added for security reasons. Don't think we should ever bypass this.

src/bundle/Resources/public/js/scripts/admin.notifications.js
czechmarcin: AdminUI is not the only place when we need this. The Password Reset view is in `Security` scope. Should I implement the `adminUiConfig` in `src/bundle/Resources/views/Security/base.html.twig` instead?

dew326: Good question. My concern is that those notifications will not respect the timeouts. Maybe we should introduce some limited adminUiConfig? Only with needed things?
@Nattfarinn @webhdx

src/bundle/Resources/public/js/scripts/admin.notifications.js
czechmarcin: In order to have any helpers in `Security/base.html.twig` I should implement the [addConfig](https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/bundle/Resources/views/layout.html.twig#L20) function. Would it be okay if I separate this code into other `twig` file and include it in both files `Security/base.html.twig` and `layout.html.twig`?

czechmarcin:
Problem solved another way, please refer to PRs:
https://github.com/ezsystems/repository-forms/pull/288
https://github.com/ezsystems/ezplatform-user/pull/30

PR #972, opened 2019-04-15
EZP-30471: Sending Relation to trash will crash Relating Content view in Back Office

mateuszbieniek:
More in this ticket: https://jira.ez.no/browse/EZP-30471


mateuszbieniek:
TestS fails due to CS Errors in a file not related to this PR (`src/lib/UI/Config/Provider/FieldType/RichText/AlloyEditor.php`)

PR #977, opened 2019-04-17
EZP-30489: Get console logs in failed tests

mnocon:
Don't merge, contains TMP commit.

Example how we could access browser logs in tests, output:
```
      Failed asserting that two strings are equal.
      --- Expected
      +++ Actual
      @@ @@
      -'Content type 'Test Content TypeFAILURE' updated.'
      +'Content type 'Test Content Type' updated.'
    │
    │  http://web/assets/ezplatform/build/ezplatform-admin-ui-layout-js.js 0:7119 "Test error"http://web/assets/ezplatform/build/ezplatform-admin-ui-layout-js.js 0:7119 "Test error"
    │
    └─ @AfterStep # EzSystems\EzPlatformAdminUi\Behat\Helper\Hooks::getBrowserLogAfterFailedStep()

```

Any ideas/suggestions are welcome.

*Notes:* 
1) I don't think it's possible to get a better output (without `echo`) without writing our extension for this - not sure if it's worth it.
2) I've went for AfterStep because this will show the console log closest to the moment when the error happened, but maybe it could be AfterScenario?

### DIFF
--- a/src/lib/Behat/Helper/Hooks.php
+++ b/src/lib/Behat/Helper/Hooks.php
@@ -6,13 +6,19 @@
  */
 namespace EzSystems\EzPlatformAdminUi\Behat\Helper;
 
+use Behat\Behat\Hook\Scope\AfterStepScope;
+use Behat\Mink\Driver\Selenium2Driver;
 use Behat\MinkExtension\Context\RawMinkContext;
 use Behat\Symfony2Extension\Context\KernelDictionary;
+use Behat\Testwork\Tester\Result\TestResult;
 use EzSystems\EzPlatformAdminUi\Behat\PageElement\ElementFactory;
 use EzSystems\EzPlatformAdminUi\Behat\PageObject\PageObjectFactory;
+use WebDriver\LogType;
 
 class Hooks extends RawMinkContext
 {
+    private const CONSOLE_LOGS_LIMIT = 10;
+
     use KernelDictionary;
 
     /** @BeforeScenario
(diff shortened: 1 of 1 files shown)


m-tyrala:
Can we precise the amount of lines we want to fetch?

mnocon:
Not really, looks like it fetches all of them. But it's a good idea, maybe we should only display only last 10 errors. What do you think?

m-tyrala:
I was thinking of 10 warning/errors, to handle the cases when we would have a waterfall of errors. Also, maybe it would be better to take the first 10, not the last 10? As the last errors can be caused by the first ones, that we would not see.

mnocon:
Limited the output to first 10 errors. Also changed output formatting:
```
      Failed asserting that two strings are equal.
      --- Expected
      +++ Actual
      @@ @@
      -'Content type 'Test Content TypeFAILURE' updated.'
      +'Content type 'Test Content Type' updated.'
    │
    │  JS console errors:
    │  http://web/assets/ezplatform/build/ezplatform-admin-ui-layout-js.js 0:7119 "Test error"
    │  http://web/assets/ezplatform/build/ezplatform-admin-ui-layout-js.js 0:7119 "Test error"
    │  
    │
    └─ @AfterStep # EzSystems\EzPlatformAdminUi\Behat\Helper\Hooks::getBrowserLogAfterFailedStep()
```

m-tyrala:
About what next - I'd wait for the [EZP-30376](https://jira.ez.no/browse/EZP-30376) - we'd probably make console output as an attachment to the scenario in the report :)

mnocon:
@lserwatka this one can be merged as well

PR #980, opened 2019-04-23
EZP-30473: Reorganize templates in folders

pawbuj:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30473](https://jira.ez.no/browse/EZP-30473)
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Tests pass?   | no
| Doc needed?   | yes

new structure for template files
#### Checklist:

- [x] Ready for Code Review


src/bundle/Controller/ContentOnTheFlyController.php
```diff
@@ -91,7 +91,7 @@ public function createContentAction(
             }
         }
 
-        return new ContentCreateOnTheFlyView('@ezdesign/content/content_on_the_fly/content_create_on_the_fly.html.twig', [
+        return new ContentCreateOnTheFlyView('@ezdesign/ui/on_the_fly/content_create_on_the_fly.html.twig', [
```
mikadamczyk: `Content on the fly` is invented name, so we should use `content_on_the_fly` here

src/bundle/Resources/views/ui/layout.html.twig
```diff
@@ -111,7 +111,7 @@
         <div id="react-udw"></div>
 
         {% block navigation %}
-            {% include '@ezdesign/part/navigation.html.twig' %}
+            {% include '@ezdesign/ui/menu/top_navigation.html.twig' %}
```
mikadamczyk: We do not have any other navigation template, also twig block is called `navigation` so I think that this template name could stay as is, or we should change block name too. `menu` in combination with `navigation` looks weird.

mikadamczyk:
What about my previous concerns? Has anybody already got a say in this matter?

PR #993, opened 2019-05-09
[Online Editor] Make sure custom tag elements has all required CSS cl…

SerheyDolgushev:
| Question      | Answer
| ------------- | ---
| Tickets       | N/A
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

1. Convert eZ XML to RichtText using https://github.com/ezsystems/ezplatform-xmltext-fieldtype
2. Setup custom tags
3. Try to edit any content with converted RichtText field which has any custom tag.
4. Custom tag edit UI will behave very strange. This PR fixes it.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/alloyeditor/src/widgets/ez-custom-tag-base.js
+++ b/src/bundle/Resources/public/js/alloyeditor/src/widgets/ez-custom-tag-base.js
@@ -37,6 +37,7 @@ const customTagBaseDefinition = {
      */
     insert: function() {
         const element = CKEDITOR.dom.element.createFromHtml(this.template.output(this.defaults));
+
         const wrapper = this.editor.widgets.wrapElement(element, this.name);
 
         this.editor.widgets.initOn(element, this.name);
@@ -65,6 +66,11 @@ const customTagBaseDefinition = {
     },
 
     init: function() {
+        // Converted from eZ XML custom tags might miss some required classes
+        if (this.element.hasClass('ez-custom-tag') === false) {
+            this.element.addClass('ez-custom-tag');
+        }
+
         this.on('focus', this.fireEditorInteraction);
         this.syncAlignment(true);
         this.renderAttributes();


src/bundle/Resources/public/js/alloyeditor/src/widgets/ez-custom-tag-base.js
```diff
@@ -65,6 +66,11 @@ const customTagBaseDefinition = {
     },
 
     init: function() {
+        // Converted from eZ XML custom tags might miss some required classes
```
vidarl: From my point of view, it seems a bit odd to fix the symptom here, if the actual problem is that something is missing in the xmltext -> richtext conversion ?

dew326: I agree with @vidarl. Seems like this not the proper way of handling this. `admin-ui` should not have any check if something was properly converted from eZ XML.

SerheyDolgushev:
Sorry, let me rephrase the original issue which this PR is trying to solve:

It is expected that you should be able to edit RichText content in Admin UI Online Editor if that content follows RichtText schema. And there are no additional Admin UI schemas.
But In order to be able to manage custom tags in Admin UI Online Editor, they need to have `ez-custom-tag` in `ezxhtml:class` attribute. This is not part of the RichText schema. And I'm not sure it is mentioned somewhere. In most cases, it does not cause any issues, because each time when you add a new custom tag via Online Editor `ez-custom-tag` will be stored in its `ezxhtml:class` attribute. But it leads to unexpected results in more rare cases:

1. Content is migrated from `5.4` to `2.5` using https://github.com/ezsystems/ezplatform-xmltext-fieldtype. You can not edit migrated custom tags in Online Editor.

2. Content is created by REST API/custom functionality (and it is validated against RichText schema). Again, it will be impossible to edit its custom tags in Online Editor.

3. Custom UI is used to manage content. It respects and follows RichText schema and works fine. At some point, Custom UI will be switched to eZ Platform Admin UI. And it will be impossible to edit custom tags after this switch.

Taking into account all use cases above, I think this should be fixed in https://github.com/ezsystems/ezplatform-admin-ui, but not in any other bundle. Alternatively, eZ Platform Admin UI should provide its additional schema (which brings redundant complexity).

SerheyDolgushev:
@dew326 @vidarl sorry, should this PR be closed?

PR #997, opened 2019-05-15
EZP-30494: Default Time field value is not saved

konradoboza:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30494](https://jira.ez.no/browse/EZP-30494)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Default Time field value from Flatpicker is not saved.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/eztime.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/eztime.js
@@ -112,6 +112,7 @@
         window.flatpickr(flatPickrInput, Object.assign({}, timeConfig, {
             enableSeconds,
             onChange: updateInputValue.bind(null, sourceInput),
+            onClose: updateInputValue.bind(null, sourceInput),
             dateFormat: enableSeconds ? 'H:i:S' : 'H:i',
             defaultDate
         }));


micszo:
Why not 1.4? :)

konradoboza:
Customer is on 2.5, but if there is strong need I guess we should rebase it to 2.4. Not sure though if it wouldn't be problematic when it comes to distributing patches as 2.4 didn't have webpack.

micszo:
Right, webpack. Asking on the simple premise that 2.4 is affected and its EOL is June. But it's not a critical defect so if the cost is high... Let's see what FE thinks.

dew326:
Don't see a reason why missing webpack in 2.4 is a problem. This is not alloyEditor and it wasn't pushed to the bundle as minified in 2.4. It should be low-hanging fruit to include this in 2.4 but it's up to you support guys.

konradoboza:
Rebased to 2.4, please take a look @dew326 @tischsoic.

lserwatka:
You can merge it up.

konradoboza:
Done.

PR #998, opened 2019-05-16
EZEE-2782 [Behat] Problems with form preview and with notification close

m-tyrala:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZEE-2782](https://jira.ez.no/browse/EZEE-2782)
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

fixing `waitUntilElementIsVisible` function

It seems, that there are situations when behat has a problem with nodeElement references when `callback` uses previously (outside) defined base element.

There is still one place, where the base element will be pre-defined (`UniversalDiscoveryWidget.php:62`), but changing that will be harder (as forwarded baseElement is found not only by its locator but also by text content). I'll fix that, but I need a little more time, and I want Form-Builder tests to be green at last.

related PR with test builds:
https://github.com/ezsystems/ezplatform-form-builder/pull/164

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/lib/Behat/Helper/UtilityContext.php
+++ b/src/lib/Behat/Helper/UtilityContext.php
@@ -25,10 +25,10 @@ class UtilityContext extends MinkContext
      */
     public function waitUntilElementIsVisible(string $cssSelector, int $timeout = 5, TraversableElement $baseElement = null): void
     {
-        $baseElement = $baseElement ?? $this->getSession()->getPage();
-
         try {
             $this->waitUntil($timeout, function () use ($cssSelector, $baseElement) {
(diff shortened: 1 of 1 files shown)


m-tyrala:
Yup, that's all technically. I can make it an anonymous function again (for me that's the clearer version 😁 and I needed it that way here for a moment, so I left it, but that's not particularly important for me). Also, I'll rebase it to 1.4.

m-tyrala:
@lserwatka ready to merge

lserwatka:
You can merge it up.

PR #1000, opened 2019-05-22
Applied fixes for tests after suggestions from Infection framework

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | N/A
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Tests/ParamConverter/ContentTypeDraftParamConverterTest.php
+++ b/src/bundle/Tests/ParamConverter/ContentTypeDraftParamConverterTest.php
@@ -30,15 +30,20 @@ class ContentTypeDraftParamConverterTest extends AbstractParamConverterTest
         $this->converter = new ContentTypeDraftParamConverter($this->contentTypeServiceMock);
     }
 
-    public function testApply()
+    /**
+     * @dataProvider dataProvider
+     *
+     * @param mixed $contentTypeId The content type identifier fetched from the request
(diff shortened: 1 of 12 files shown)


src/bundle/Tests/ParamConverter/ContentTypeDraftParamConverterTest.php
```diff
@@ -65,4 +69,13 @@ public function testApplyWithWrongAttribute()
         $this->assertFalse($this->converter->apply($request, $config));
         $this->assertNull($request->attributes->get(self::PARAMETER_NAME));
     }
+
+    public function dataProvider(): array
```
ViniTou: Those came from contentTypeId marked as mixed, right? If yes, not sure about those as we are moving to strict type those as `int` now.

also, 1000th PR! :tada:

mikadamczyk: We test `string` here because param converter casts passed value to an integer

ViniTou: Yup, for some reason I thought it is passed somewhere else, not to ParamConverter.

PR #1002, opened 2019-05-24
EZP-30595 [Behat] Element with selector: .c-select-content-button was not found

m-tyrala:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30595](https://jira.ez.no/browse/EZP-30595)
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

We can't refresh (redefine) NodeElement so in case of the problem with StaleElementReferenceException on that element we can't use them as a base element (we still can do it in the other situations, where the element doesn't change dynamically). Because of that, to have the possibility to redefine element, we should use the full CSS locator to it, not the partial locator and the base element.

Though it's the most probable error cause, we still should test this solution for a few days after merging, before closing the ticket.

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/lib/Behat/PageElement/UniversalDiscoveryWidget.php
+++ b/src/lib/Behat/PageElement/UniversalDiscoveryWidget.php
@@ -26,7 +26,8 @@ class UniversalDiscoveryWidget extends Element
             'confirmButton' => '.m-ud__action--confirm',
             'cancelButton' => '.m-ud__action--cancel',
             'selectContentButton' => '.c-select-content-button',
-            'elementSelector' => '.c-finder-tree-branch:nth-of-type(%d) .c-finder-tree-leaf',
+            'genericElementSelector' => '.c-finder-tree-branch:nth-of-type(%d) .c-finder-tree-leaf',
+            'specificElementSelector' => '.c-finder-tree-branch:nth-of-type(%d) .c-finder-tree-leaf:nth-of-type(%d)',
             'branchLoadingSelector' => '.c-finder-tree-leaf--loading',
             'previewName' => '.c-meta-preview__name',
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -28,7 +28,7 @@ matrix:
     - name: "Admin-UI tests on Clean P"
       env:
       - COMPOSE_FILE="doc/docker/base-dev.yml:doc/docker/selenium.yml"
-      - BEHAT_OPTS="--profile=adminui --suite=adminui"
+      - BEHAT_OPTS="--mode=behat --profile=adminui --suite=adminui"
```
mnocon: This shouldn't be here I guess

src/lib/Behat/PageElement/UniversalDiscoveryWidget.php
```diff
@@ -27,6 +27,7 @@ public function __construct(UtilityContext $context)
             'cancelButton' => '.m-ud__action--cancel',
             'selectContentButton' => '.c-select-content-button',
             'elementSelector' => '.c-finder-tree-branch:nth-of-type(%d) .c-finder-tree-leaf',
+            'certainElementSelector' => '.c-finder-tree-branch:nth-of-type(%d) .c-finder-tree-leaf:nth-of-type(%d)',
```
mnocon: I'd go with `genericElementSelector` and `specificElementSelector`

m-tyrala:
@lserwatka ready to merge

lserwatka:
You can merge it up

m-tyrala:
Done

PR #1006, opened 2019-05-31
EZP-30617 [Behat] Problem with select item in multiselect UDW

m-tyrala:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30617](https://jira.ez.no/browse/EZP-30617)
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

There are three fixes here:

- first, for the occasional problems with waiting for notification - moving notification checking steps as close as it can be to action that causes this notification
- second, for the multi-select UDW, where we occasionally but regularly get errors with select button not showing up after performing mouse-over; as there is no clear solution for that problem at the moment, I decided to enclose this action in the three-iteration loop, and throw if the last iteration fails
- third is the increasing of UDW timeouts

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/src/lib/Behat/PageElement/UniversalDiscoveryWidget.php
+++ b/src/lib/Behat/PageElement/UniversalDiscoveryWidget.php
@@ -13,9 +13,9 @@ use PHPUnit\Framework\Assert;
 class UniversalDiscoveryWidget extends Element
 {
     public const ELEMENT_NAME = 'UDW';
-    private const UDW_TIMEOUT = 20;
-    private const UDW_DISAPPEAR_TIMEOUT = 1;
-    private const UDW_BRANCH_LOADING_TIMEOUT = 5;
+    private const UDW_LONG_TIMEOUT = 20;
+    private const UDW_SHORT_TIMEOUT = 2;
+    private const UDW_BRANCH_LOADING_TIMEOUT = 10;
 
     public function __construct(UtilityContext $context)
     {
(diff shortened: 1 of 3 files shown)


src/lib/Behat/PageElement/UniversalDiscoveryWidget.php
```diff
@@ -55,12 +55,21 @@ function () use ($expectedContentName) {
             });
 
         if ($this->isMultiSelect()) {
-            $itemToSelect = $this->context->getElementByText($expectedContentName, sprintf($this->fields['elementSelector'], $depth));
-
-            $this->context->findElement($this->fields['tabSelector'])->mouseOver();
-            $itemToSelect->mouseOver();
-            $this->context->waitUntilElementIsVisible($this->fields['selectContentButton'], $this->defaultTimeout, $itemToSelect);
-            $this->context->findElement($this->fields['selectContentButton'], $this->defaultTimeout, $itemToSelect)->click();
+            for ($i = 0; $i < 3; ++$i) {
```
mnocon: Would it make sense to add a sleep(1) after each loop? Usually retry mechanisms are implemented that way.

src/lib/Behat/PageElement/UniversalDiscoveryWidget.php
```diff
@@ -14,8 +14,8 @@ class UniversalDiscoveryWidget extends Element
 {
     public const ELEMENT_NAME = 'UDW';
     private const UDW_TIMEOUT = 20;
-    private const UDW_DISAPPEAR_TIMEOUT = 1;
-    private const UDW_BRANCH_LOADING_TIMEOUT = 5;
+    private const UDW_DISAPPEAR_TIMEOUT = 2;
+    private const UDW_BRANCH_LOADING_TIMEOUT = 10;
```
mnocon: Is it really necessary? The issue is that selecting an item from a single select UDW will take at least 10 seconds every time (`isMultiSelect` call is blocking for UDW_BRANCH_LOADING_TIMEOUT seconds when UDW is single-select).

Maybe `isMultiSelect` can be changed so that UDW mode is detected in a better way (as far as I remember it wasn't possible when it was implemented).

m-tyrala:
Yes, 1.5 is the target here because in 2.5 we are testing multi-select in admin-ui suite (Role assignment) - it will be rebased before merging.

src/lib/Behat/PageElement/UniversalDiscoveryWidget.php
m-tyrala: I'd prefer not - if you look at actions taken, it looks like:
1. find the item to select
2. mouse over the first UDW tab
3. mouse over the item found
4. wait until select button in the item element is visible
5. click the select button
if at any point we'll get the exception, we repeat from the first step. Plus, we have the built-in 5 seconds timeout in steps 1, 2 and 5. I don't think we'll benefit from this additional second.

m-tyrala:
@lserwatka ready to merge

lserwatka:
You can merge it up.

m-tyrala:
Done

PR #1015, opened 2019-06-13
EZP-30665: Custom styles dropdown becomes unusable when contains many…

czechmarcin:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30665
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

It is not possible to select custom style when having configured many of them.

PR result:
Screenshot 2019-06-13 at 12 55 34

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/css/alloyeditor/alloyeditor-ez.css
+++ b/src/bundle/Resources/public/css/alloyeditor/alloyeditor-ez.css
@@ -722,7 +722,7 @@
     font-size: 16px;
     height: 3em;
     line-height: 28px;
-    margin-right: 0;
+    margin-right: 15px;
     max-height: 44px;
     min-width: 100%;
     padding: 8px 12px;
--- a/src/bundle/Resources/public/css/alloyeditor/alloyeditor-ez.css
+++ b/src/bundle/Resources/public/css/alloyeditor/alloyeditor-ez.css
@@ -711,8 +711,10 @@
     list-style: none;
     margin: 0;
     min-height: 44px;
+    max-height: 400px;
     min-width: 132px;
     padding: 0;
+    overflow-y: auto;
 }
 
 .ae-ui .ae-dropdown .ae-listbox .ae-toolbar-element,


src/bundle/Resources/public/scss/_alloyeditor-custom-styles.scss
```diff
@@ -20,3 +20,7 @@
         font-family: courier, monospace;
     }
 }
+ul.ae-listbox {
```
dew326: Those styles are not related to custom styles. It will affect all `lists` in AE so I would move this to the `ezplatform-admin-ui/src/bundle/Resources/public/css/alloyeditor/alloyeditor-ez.css`
Also, I would change the selector to match AE selector: `.ae-ui .ae-dropdown .ae-listbox`

czechmarcin: @dew326 Please have a look now. I used and tested your proposition.

czechmarcin:
> On Firefox & Safari (not on Chrome & Edge) second scrollbar appears at the bottom.
> Question is can we do something about it at a low cost? It's not a big deal, just looks bad.
> Screenshot 2019-06-13 at 15 35 06

I'm not sure if that's the case here. @dew326, do you think it should be changed?

dew326:
@czechmarcin We discussed this issue before @micszo put the comment. It would be good to fix this.

czechmarcin:
@micszo As I'm not able to reproduce it even on the same versions of web browsers and the customer is waiting for the fix, could we agree to proceed with existing changes without fixing the width? Cc: @dew326

czechmarcin:
@micszo After a deeper investigation, I am able to reproduce this issue. Working on a fix

dew326:
@czechmarcin can you merge it up?

czechmarcin:
@dew326 merged

PR #1035, opened 2019-07-12
EZP-30766: CSS files not loaded in case of error in search results tempate

ITernovtsiy:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30766
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

In case of error appearing after the "encore_entry_link_tags" call (in form handling function) - UI is loading without CSS (as encore think it's already rendered to the page).

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/SearchController.php
+++ b/src/bundle/Controller/SearchController.php
@@ -23,6 +23,7 @@ use Pagerfanta\Pagerfanta;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Symfony\WebpackEncoreBundle\Asset\EntrypointLookupCollectionInterface;
 
 class SearchController extends Controller
 {
@@ -47,6 +48,9 @@ class SearchController extends Controller
     /** @var \eZ\Publish\API\Repository\ContentTypeService */
     private $contentTypeService;
 
+    /** @var \Symfony\WebpackEncoreBundle\Asset\EntrypointLookupCollectionInterface */
+    private $entrypointLookupCollection;
+
     /** @var int */
     private $defaultPaginationLimit;
 
@@ -61,6 +65,7 @@ class SearchController extends Controller
      * @param \EzSystems\EzPlatformAdminUi\Form\SubmitHandler $submitHandler
      * @param \eZ\Publish\API\Repository\SectionService $sectionService
      * @param \eZ\Publish\API\Repository\ContentTypeService $contentTypeService
+     * @param \Symfony\WebpackEncoreBundle\Asset\EntrypointLookupCollectionInterface $entrypointLookupCollection
      * @param int $defaultPaginationLimit
      * @param array $userContentTypeIdentifier
      */
@@ -72,6 +77,7 @@ class SearchController extends Controller
         SubmitHandler $submitHandler,
         SectionService $sectionService,
         ContentTypeService $contentTypeService,
+        EntrypointLookupCollectionInterface $entrypointLookupCollection,
         int $defaultPaginationLimit,
         array $userContentTypeIdentifier
     ) {
@@ -84,6 +90,7 @@ class SearchController extends Controller
         $this->contentTypeService = $contentTypeService;
         $this->defaultPaginationLimit = $defaultPaginationLimit;
         $this->userContentTypeIdentifier = $userContentTypeIdentifier;
+        $this->entrypointLookupCollection = $entrypointLookupCollection;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


lserwatka:
@ITernovtsiy thank you for your contribution. Could you create a JIRA ticket for this issue and rebase with 1.5 branch?

src/bundle/Controller/SearchController.php
```diff
@@ -209,6 +216,7 @@ public function searchAction(Request $request): Response
             if ($result instanceof Response) {
                 return $result;
             }
+            $this->entrypointLookupCollection->getEntrypointLookup('ezplatform')->reset();
```
dew326: I wonder if we should reset also the `default` build?

ITernovtsiy: I tested this with `ezplatform` only and it works fine.

ITernovtsiy:
@lserwatka Jira ticket created and PR switched to 1.5 branch

webhdx:
Hello @ITernovtsiy. I have concerns regarding your solution. Why do we have to reset something in `SearchController`? This will most likely also happen in other places as well.

ITernovtsiy:
Hello @webhdx
> Why do we have to reset something in SearchController?

To make it work and load styles :) Please advise better place for this if you have it in mind.

> This will most likely also happen in other places as well.

I just reviewed all the controllers and actions and this is the only place where this bug appears. In all other cases, "redirectToRoute" response used (and "encore_entry_link_tags" not called inside a form handler).
Do you know some other place where you think this can happen as well? I'm not 100% sure I didn't miss it.

lserwatka:
@mnocon @m-tyrala Behat is failing here for some reason.

m-tyrala:
One problem is the error in Code style check and Unit tests:

> Your requirements could not be resolved to an installable set of packages.
>   Problem 1
>     - friendsofphp/php-cs-fixer v2.7.5 requires php ^5.6 || >=7.0 <7.3 -> your PHP version (7.3.2) does not satisfy that requirement.
>     - friendsofphp/php-cs-fixer v2.7.4 requires php ^5.6 || >=7.0 <7.3 -> your PHP version (7.3.2) does not satisfy that requirement.
>     - friendsofphp/php-cs-fixer v2.7.3 requires php ^5.6 || >=7.0 <7.3 -> your PHP version (7.3.2) does not satisfy that requirement.
>     - friendsofphp/php-cs-fixer v2.7.2 requires php ^5.6 || >=7.0 <7.3 -> your PHP version (7.3.2) does not satisfy that requirement.
>     - friendsofphp/php-cs-fixer v2.7.1 requires php ^5.6 || >=7.0 <7.3 -> your PHP version (7.3.2) does not satisfy that requirement.
>     - Installation request for friendsofphp/php-cs-fixer ~2.7.1 -> satisfiable by friendsofphp/php-cs-fixer[v2.7.1, v2.7.2, v2.7.3, v2.7.4, v2.7.5].

And the second is that in automated tests on `behat` env there's Symfony debug bar on the bottom of the page, despite it should not be there according to our configuration.

mnocon:
It's failing because it has been built against `master` branch - Travis does not trigger a build when switching base branches. What happened:

1. @ITernovtsiy rebased his solution on ezsystems/1.5 and force-pushed
2. Travis triggered a build against the branch that this PR has been set at that time (master) - this is the build we're seeing right now
3. Base branch has been changed from master to 1.5 - this action does not trigger a build

@ITernovtsiy I believe that if you trigger a Travis build now it will start correctly now, you can do it by simply doing another push (usually I do something like: `git commit --amend` with`git push -f`)

m-tyrala:
That explains a lot, I wouldn't have thought of that!

ITernovtsiy:
Force pushed, it's failing on another step now (not related to this PR).

m-tyrala:
That's one of the known non-deterministic errors we have once in a few hundred tests. We can ignore that in this case.

webhdx:
Hello @ITernovtsiy. I'm looking at the issue currently to possibly find better solution. Can you tell me how to reproduce the issue? I haven't been able to do so yet.

ITernovtsiy:
Hi @webhdx 
You can put `{{ path('_ezpublishLocation', {'locationId': ''}) }}` at the begining of `src/bundle/Resources/views/admin/search/search_table_row.html.twig` file and to see this error message 
`An exception has been thrown during the rendering of a template ("Parameter "locationId" for route "_ezpublishLocation" must match "[^/]++" ("" given) to generate a corresponding URL.").`

webhdx:
@ITernovtsiy I've fixed the issue: https://github.com/ezsystems/ezplatform-admin-ui/pull/1044

As I suspected, resetting encore was not the solution here. The issue was deeper - due to misuse of `SubmitHandler`, `layout.html.twig` (the one calling encore tags to render) was rendered twice. This is why Encore was not rendering `<link>` tags as it thought these were already rendered.

Error handling is in line with other parts of AdminUI and it won't render results page but rather display Exception page (in dev) and nice exception placeholder for prod environments.

I'm closing this PR in favor of https://github.com/ezsystems/ezplatform-admin-ui/pull/1044

PR #1043, opened 2019-07-30
EZP-30802: Fixed OE JS error in case of missing element path

SerheyDolgushev:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30802](https://jira.ez.no/browse/EZP-30802)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Steps to reproduce:

1. Start to create any content with Richt Text field.
2. In Online Editor add any custom tag
3. Set "Right" align for added custom tag
4. Remove an empty paragraph above the custom tag (using trash icon)
5. Click on the area in Online Editor left to the custom tag

You will get following JavaScript error in the console:
```
TypeError: Cannot read property 'lastElement' of null
```

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/alloyeditor/src/toolbars/config/ez-table-cell.js
+++ b/src/bundle/Resources/public/js/alloyeditor/src/toolbars/config/ez-table-cell.js
@@ -8,9 +8,8 @@ export default class EzTableCellConfig extends EzConfigTableBase {
     test(payload) {
         const nativeEditor = payload.editor.get('nativeEditor');
         const path = nativeEditor.elementPath();
-        const lastElement = path.lastElement;
 
-        return lastElement.is('td');
+        return path && path.lastElement.is('td');
     }
 }
 
--- a/src/bundle/Resources/public/js/alloyeditor/src/toolbars/config/ez-table-row.js
+++ b/src/bundle/Resources/public/js/alloyeditor/src/toolbars/config/ez-table-row.js
@@ -8,9 +8,8 @@ export default class EzTableRowConfig extends EzConfigTableBase {
     test(payload) {
         const nativeEditor = payload.editor.get('nativeEditor');
         const path = nativeEditor.elementPath();
-        const lastElement = path.lastElement;
 
-        return lastElement.is('tr');
+        return path && path.lastElement.is('tr');
     }
 }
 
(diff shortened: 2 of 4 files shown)


SerheyDolgushev:
Seems like a good scenario to reproduce the issue was picked up. So there is another JavaScript error after this commit:
```
TypeError: Cannot read property 'elements' of null
```

SerheyDolgushev:
Please note, all JavaScript errors seem to be fixed now. But there are still some issues with positioning "ez-add" button. In the described scenario the button sticks at the top of the window. After spending a few hours on trying to fix it, we realized we need help from Online Editor developers here.
Please feel free to complete this PR.

SerheyDolgushev:
@alongosz please have a look here.

SylvainGuittard:
@dew326 and @tischsoic Could you please have a look? Thanks

dew326:
I quickly checked the issue and don't have any good solution for that.
The problem is that after a click on the left none of the elements has focus.

SerheyDolgushev:
@dew326 in this case, maybe it worth to merge this PR, just to get rid of JavaScript error. And create have a separate task about correct positioning for `ez-add` toolbar?

When I was trying to fix it, I was able to set the correct top offset for the plus button. But the buttons which are showing up that plus button is clicked had zero top offset.

dew326:
Is this happen only when there is one custom tag and nothing else? If so we could set the position to the middle of the richtext area (or to the top). I know this is not the best solution but for me sounds okish.

SerheyDolgushev:
Sorry, not sure, but there might be some other strange scenarios when the elements path is missing.
Even if `ez-add` has the wrong position - it is still better than JavaScript error which breaks the Online Editor.

dew326:
I totally agree that misplaced button is better than breaking AE and be forced to refresh the page.

dew326:
I added a commit where add button is moved to the top of a richtext area when the path is missing. 

@tischsoic please do a review

dew326:
@katarzynazawada fixed

lserwatka:
@mikadamczyk could you merge it up?

mikadamczyk:
@lserwatka done

PR #1044, opened 2019-07-30
EZP-30766: CSS might be not loaded when exception occurs in Twig

webhdx:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30766
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

# Description
### SubmitHandler misuse in SearchController causing whole layout to load twice
`SubmitHandler` was misused in `SearchController::searchAction` as it's not designed to be used with responses rendering twig templates - it should only return `RedirectResponse` objects. 
This affected how page was rendered - when exception occurred in the template rendered inside callable it would make `SubmitHandler` return `null` which in case of `SearchController::searchAction` was rendering "default" template (page no results). The issue mentioned in the ticket (no CSS) was happening because of the Encore - it was omitting already embedded scripts and stylesheets which happened inside mentioned callable.

### Error page is loaded without CSS when exception has happened in Twig
Another commit is fixing related issue - on `prod` environment, Encore has already stored information about rendered CSS so Error page was loading without stylesheets. This was fixed by resetting it's entrypoint lookup table.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/SearchController.php
+++ b/src/bundle/Controller/SearchController.php
@@ -130,90 +130,85 @@ class SearchController extends Controller
 
         $form->handleRequest($request);
 
-        if ($form->isSubmitted()) {
-            $result = $this->submitHandler->handle($form, function (SearchData $data) use ($form) {
-                $limit = $data->getLimit();
-                $page = $data->getPage();
-                $queryString = $data->getQuery();
-                $section = $data->getSection();
-                $contentTypes = $data->getContentTypes();
-                $lastModified = $data->getLastModified();
-                $created = $data->getCreated();
-                $creator = $data->getCreator();
-                $subtree = $data->getSubtree();
-                $query = new Query();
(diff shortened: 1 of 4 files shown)


micszo:
There are still some issues in symfony env `prod`.
For `{{ path('_ezpublishLocation') }}` at the beginning of `src/bundle/Resources/views/admin/search/search_table_row.html.twig` without the fix result is:
Screenshot 2019-07-30 at 14 23 18

With the fix result is:
Screenshot 2019-07-30 at 14 24 25

webhdx:
ping @micszo Fixed.

lserwatka:
@mnocon Behat is failing here

lserwatka:
@webhdx you can merge it up

PR #1047, opened 2019-08-02
EZP-30809: changed all default ezsettings to admin_group settings

pawbuj:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30809](https://jira.ez.no/browse/EZP-30809)
| New feature?  | no
| BC breaks?    | yes
| Tests pass?   | ye
| Doc needed?   | yes

@ezsystems/documentation-team 
1) All those settings with default values should be in the documentation
2) There also should be an information that AdminUI provides defaults ezsettings for admin_group siteaccess group and those settings can be overided by global or site access ezsettings (defaults ezsettings do not override them)
3) Also, all ezsettings values listed in the documentation are not guaranteed because other installed bundles can override them in an implicit way.
 text proposed in an internal discussion  `Keep in mind that all bundles, including 3rd party and yours, might override those settings.`

#### Checklist:
- [ ] Ready for Code Review


mnocon:
@pawbuj can you rebase here so that Travis runs correctly here?

Also I think this value should be changed: https://github.com/ezsystems/ezplatform/blob/master/config/packages/behat/ezplatform.yaml#L22, as it won't work anymore after this PR is merged

mnocon:
Travis is failing with:
```
> @php bin/console ezplatform:install clean

In ChainConfigResolver.php line 92:

                                                                            

  Parameter 'content_translate_view' with namespace '' could not be found.  
```
which seems related to this PR

pawbuj:
@mnocon tests pass

PR #1056, opened 2019-08-13
EZP-30848: Fixed usage of hasAccess method with section/view permission in Admin UI

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30848
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

When a user has a Section Limitation we can not rely on hasAccess method to check if we can load Section. We must use a canUser method. 

If we need to load a list of Sections it would be better and more readable to compare the result of hasAccess method with `false`. Because section list is filtered, so we should display it when `hasAccess` method return `true` or an array of Limitation.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/Form/Type/Search/SearchType.php
+++ b/src/lib/Form/Type/Search/SearchType.php
@@ -69,7 +69,7 @@ class SearchType extends AbstractType
             ])
         ;
 
-        if ($this->permissionResolver->hasAccess('section', 'view')) {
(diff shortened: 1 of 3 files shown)


lserwatka:
@mikadamczyk you can merge it up.

PR #1060, opened 2019-08-16
EZP-30797: As an administrator, I want to configure a password expiration for users

adamwojs:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30797](https://jira.ez.no/browse/EZP-30797)
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | -
| Doc needed?   | -

### Screenshots

Information about expired credentials in the backoffice: 

Zrzut ekranu 2019-08-28 o 09 54 44

Password expiration warning after login into backoffice

image

New settings in ezuser field type

image

Password expiration info in the ezuser field type preview

image

Information about expired credentials in the frontend siteaccess

Zrzut ekranu 2019-08-28 o 10 42 35

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/config/services.yml
+++ b/src/bundle/Resources/config/services.yml
@@ -127,6 +127,18 @@ services:
         tags:
             - {name: kernel.event_subscriber, priority: -250}
 
+    EzSystems\EzPlatformAdminUi\EventListener\CredentialsExpiredListener:
+        arguments:
+            $siteAccessGroups: '%ezpublish.siteaccess.groups%'
+        tags:
+            - {name: kernel.event_subscriber}
+
+    EzSystems\EzPlatformAdminUi\EventListener\CredentialsExpirationWarningListener:
+        arguments:
+            $siteAccessGroups: '%ezpublish.siteaccess.groups%'
+        tags:
+            - {name: kernel.event_subscriber}
+
     EzSystems\EzPlatformAdminUiBundle\Templating\Twig\UserPreferencesGlobalExtension:
         lazy: true
 
--- a/src/bundle/Resources/views/fieldtypes/preview/content_fields.html.twig
+++ b/src/bundle/Resources/views/fieldtypes/preview/content_fields.html.twig
@@ -208,6 +208,25 @@
             </tr>
         </tbody>
     </table>
+
+    {% if parameters['password_expires_at'] %}
+        <p class="mt-2 mb-0">
+            {% if parameters['is_password_expired'] %}
+                {{ 'ezuser.password_already_expired'|trans|desc('Password has expired') }}
+            {% else %}
+                {% set password_expire_in = parameters['password_expires_in'] %}
+
+                {% if password_expire_in.d > 0 %}
+                    {{ 'ezuser.password_expires_in'|trans({
+                        '%count%': password_expire_in.d + (password_expire_in.h >= 12 ? 1 : 0)
+                    })|desc('Current password <b>expires in %count% days</b>')|raw }}
+                {% else %}
+                    {{ 'ezuser.password_expires_today'|trans({
+                    })|desc('Current password <b>expires today</b>')|raw }}
+                {% endif %}
+            {% endif %}
+        </p>
+    {% endif %}
 </div>
 {% endapply %}
 {% endblock %}
--- a/src/lib/EventListener/CredentialsExpirationWarningListener.php
+++ b/src/lib/EventListener/CredentialsExpirationWarningListener.php
@@ -1,0 +1,116 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformAdminUi\EventListener;
+
+use DateTime;
+use DateTimeInterface;
+use eZ\Publish\API\Repository\UserService;
+use eZ\Publish\Core\MVC\Symfony\Security\UserInterface;
+use EzSystems\EzPlatformAdminUi\Notification\NotificationHandlerInterface;
+use EzSystems\EzPlatformAdminUi\Specification\SiteAccess\IsAdmin;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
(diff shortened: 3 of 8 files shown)


src/bundle/Resources/views/fieldtypes/preview/content_fields.html.twig
```diff
@@ -208,6 +208,18 @@
...
         </tbody>
     </table>
+
+    {% if parameters['password_expires_at'] %}
+        <p class="mt-2 mb-0">
+            {% if parameters['is_password_expired'] %}
+                {{ 'ezuser.password_already_expired'|trans|desc('Password has expired') }}
+            {% else %}
+                {{ 'ezuser.password_expire_in'|trans({
+                    '%count%': parameters['password_expires_in'].days
```
ViniTou: I think we should stick with `password_expire_in`

src/lib/EventListener/CredentialsExpirationWarningListener.php
```diff
@@ -0,0 +1,102 @@
...
+    /** @var \Symfony\Component\Translation\TranslatorInterface */
+    private $translator;
+
+    /** @var \Symfony\Component\Routing\Generator\UrlGeneratorInterface */
+    private $urlGenerator;
+
+    /** @var \eZ\Publish\API\Repository\UserService */
+    private $userService;
+
+    /** @var array */
```
ViniTou:
```suggestion
    /** @var string[] */
```

adamwojs: Actually, the correct type hint is `string[][]` 😉 

image

src/bundle/Resources/views/fieldtypes/preview/content_fields.html.twig
adamwojs: translation key needs to be changed: `password_expires_in` is the correct form. Fixed in  https://github.com/ezsystems/ezplatform-admin-ui/pull/1060/commits/6ef932bb76c280e35cd453553fcfabbf15dae82b

src/lib/EventListener/CredentialsExpirationWarningListener.php
adamwojs: Fixed in https://github.com/ezsystems/ezplatform-admin-ui/pull/1060/commits/6ef932bb76c280e35cd453553fcfabbf15dae82b

adamwojs:
PR updated according to @ViniTou code review suggestions.

src/bundle/Resources/config/services.yml
```diff
@@ -127,6 +127,18 @@ services:
         tags:
             - {name: kernel.event_subscriber, priority: -250}
 
+    EzSystems\EzPlatformAdminUi\EventListener\CredentialsExpiredListener:
```
alongosz: Does "credentials" naming already exist in AdminUI or is it fresh, based on old `isCredentialsNonExpired` in Kernel?
Asking because in general this is about just password, not all credentials.

adamwojs: It's fresh, based on `\Symfony\Component\Security\Core\Exception\CredentialsExpiredException`.

alongosz: > It's fresh, based on `\Symfony\Component\Security\Core\Exception\CredentialsExpiredException`.

Hmm, ok, I won't argue with that then, but only because it would be confusing to use a different one here.
:+1:

src/bundle/Resources/views/fieldtypes/preview/content_fields.html.twig
```diff
@@ -208,6 +208,18 @@
             </tr>
         </tbody>
     </table>
+
+    {% if parameters['password_expires_at'] %}
```
alongosz: Now when I look at the usage I feel like this should be more strict, using object and methods rather than string keys.

src/lib/EventListener/CredentialsExpiredListener.php
```diff
@@ -0,0 +1,65 @@
...
+        if (!($view instanceof LoginFormView)) {
+            return;
+        }
+
+        if ($view->getLastAuthenticationException() instanceof CredentialsExpiredException) {
+            $view->setTemplateIdentifier('@ezdesign/Security/error/credentials_expired.html.twig');
+        }
+    }
+
+    private function isAdminSiteAccess(Request $request): bool
```
alongosz: Don't we have this defined somewhere already?

adamwojs: Actually in all places where `\EzSystems\EzPlatformAdminUi\Specification\SiteAccess\IsAdmin` is used. Would be great to introduce something like: 

```php
namespace EzSystems\EzPlatformAdminUi\API;

interface SiteAccessService 
{
	# Other SA's related stuff ...  

	public function isAdmin(SiteAccess $siteaccess): bool;
}
```

and remove all code duplications. But right now, I prefer to have this logic in the private method for readability.

barbaragr:
Rebase is needed here, I suppose.

barbaragr:
1. Information about expired credentials in the backoffice - is the phrase "An email has been sent to you with instructions." put here on purpose? When we click "Reset password", we will receive another message about an email. Lots of emails.  :)

2. Password expiration warning after login into backoffice - maybe we can modify it a little to put the whole path: "Consider updating it in User Settings/My Account settings" or just User menu, because we can just click "Change password" or select "User settings/My Account settings" and then change the password. ping @SylvainGuittard  

3. "Your password will expire in 1 days" maybe we can change it for "day(s)" to avoid this kind of issues.
Screen Shot 2019-09-10 at 13 58 29

4. Settings in ezuser field type - in prototype they are required. In application they are not. Are we OK with that?

5. If I edit existing User content type, fill expiration days and save, new rules will work only for newly created users. It will not affects already existing user. Do we need any script for updating database to involve the old ones?

inakijv:
tT
–t
T

alongosz:
> If I edit existing User content type, fill expiration days and save, new rules will work only for newly created users. It will not affects already existing user. Do we need any script for updating database to involve the old ones?

No, we don't. It would be a classic example of a BC break :upside_down_face: We can provide in the Doc optional script with a warning that it will affect all users.

SylvainGuittard:
> 2. Password expiration warning after login into backoffice - maybe we can modify it a little to put the whole path: "Consider updating it in User Settings/My Account settings" or just User menu, because we can just click "Change password" or select "User settings/My Account settings" and then change the password. 

The other option is to have a link in the notification bar. Can we have a link in this element?

PR #1064, opened 2019-08-23
EZP-28997: Fixed issue with access to root of tree content and parent…

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-28997
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    |no
| Tests pass?   | yes
| Doc needed?   |no

The issue with viewing a location was related to displaying a Content Tree. User with this set of permissions was not able to see the root of Content Tree. In this case, we will show the Content Tree with viewing location as a root of the Tree. This behavior will only be in the situation when none of the defined roots is available.

The second change is about the accessibility of parentLocation, when a user enters edit view. In those situations, we used to load resource by `sudo` method, like in `supplyIsLocationInvisible` method.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/ContentViewController.php
+++ b/src/bundle/Controller/ContentViewController.php
@@ -468,23 +468,38 @@ class ContentViewController extends Controller
      */
     private function supplyContentTreeParameters(ContentView $view): void
     {
-        $location = $view->getLocation();
+        $view->addParameters([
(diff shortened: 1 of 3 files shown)


lserwatka:
You can merge it up

PR #1068, opened 2019-08-26
EZP-30884: Implement scenarios for Subtree editor

mnocon:
JIRA: https://jira.ez.no/browse/EZP-30884

Adding scenarios that test whether SubtreeEditor can (and cannot) perform his usual actions in AdminUI.

### DIFF
--- a/features/personas/SubtreeEditor.feature
+++ b/features/personas/SubtreeEditor.feature
@@ -1,0 +1,69 @@
+@javascript @subtreeEditor
+Feature: Verify that an Editor with Subtree limitations can perform all his tasks
+
+  Background:
+    Given I open Login page
+    And I login as "SubtreeEditor" with password "Passw0rd-42"
+    And I go to "Content structure" in "Content" tab
+
+  Scenario Outline: I can create and publish Content in locations I'm allowed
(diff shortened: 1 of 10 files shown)


features/personas/SubtreeEditor.feature
```diff
@@ -0,0 +1,69 @@
...
+
+  Scenario Outline: I can create and publish Content in locations I'm allowed
+    Given I navigate to content "<parentContentItemName>" of type "DedicatedFolder" in "<contentPath>"
+    And I start creating a new content "DedicatedFolder"
+    And I set content fields
+      | label      | value         |
+      | Name       | <contentName> |
+      | Short name | <contentName> |
+    When I click on the edit action bar button "Publish"
+    Then success notification that "Content published." appears
```
micszo: Nitpick: Is the dot here on purpose?

mnocon: Yes, it would fail without the dot (we're checking for string equality in https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/lib/Behat/BusinessContext/NotificationContext.php#L38).

PR #1073, opened 2019-08-29
EZP-30788: Harmonize Confirm button in widgets

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30788
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

### DIFF
--- a/src/bundle/Resources/public/scss/_buttons.scss
+++ b/src/bundle/Resources/public/scss/_buttons.scss
@@ -25,7 +25,7 @@
     }
 }
 
-.ez-btn--wide {
+.ez-btn--wide.ez-btn--wide.ez-btn--wide {
     padding: 0.5rem 3rem;
 }
 
--- a/src/bundle/Resources/public/scss/_buttons.scss
+++ b/src/bundle/Resources/public/scss/_buttons.scss
@@ -25,10 +25,6 @@
     }
 }
 
-.ez-btn--wide.ez-btn--wide.ez-btn--wide {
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/scss/_buttons.scss
```diff
@@ -25,7 +25,7 @@
     }
 }
 
-.ez-btn--wide {
+.ez-btn--wide.ez-btn--wide.ez-btn--wide {
```
dew326: This maybe is clever but not really understandable at first look and someone in the future may remove this because will take this as a mistake.

I would go with regular `.ez-btn.ez-btn--wide`

tischsoic: We need 3 classes, because `.ez-context-menu .btn` is defined below `.ez-btn--wide` in generated CSS. Do we want `.ez-btn.ez-btn.ez-btn--wide`?

dew326: Let's just move it below the `.ez-context-menu .btn` and use the selector for "BC":
```
.ez-btn--wide,
.ez-btn.ez-btn--wide
```

tischsoic:
merged up: 
- master: https://github.com/ezsystems/ezplatform-admin-ui/commit/4a0f6dc370a3329a183f349a685d124447f36114

PR #1083, opened 2019-09-10
EZP-30320: As user I want to see container-adapting tabs collapsed into a dropdown

tischsoic:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30320
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

*Related PR:* https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/210

## Screenshot
jgGWiZd4OO


tischsoic:
In last commit there is alternative UX solution:
BuONIcdN6y

But code needs further discussion - it may be too hacky.

inakijv:
If we could add the text "More" to the tab in the alternative solution, I'd rather go for this option. From the UX perspective, active label is always visible here (compared to first one).

micszo:
@lserwatka this could be merged.

PR #1089, opened 2019-09-23
EZP-30961 - Toolbar missed after using navigation path and embed

ITernovtsiy:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30961
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

STR:
1) Open editor
2) add embed object and select it
3) click on item in the navigation path

AR: JS error in console, toolbar not appearing at all (even after further clicks in editor) 
ER: No error returned

video attached in JIRA ticket

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/alloyeditor/src/plugins/ez-elements-path.js
+++ b/src/bundle/Resources/public/js/alloyeditor/src/plugins/ez-elements-path.js
@@ -60,13 +60,16 @@
             return;
         }
 
-        editor.fire('editorInteraction', {
-            nativeEvent: {
-                editor: editor,
-                target: element.$,
-            },
-            selectionData: editor.getSelectionData(),
-        });
+        const selectionData = editor.getSelectionData();
+        if (selectionData) {
+            editor.fire('editorInteraction', {
+                nativeEvent: {
+                    editor: editor,
+                    target: element.$,
+                },
+                selectionData: selectionData,
+            });
+        }
     };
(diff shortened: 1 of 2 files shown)


dew326:
Hi @ITernovtsiy,
I cannot reproduce your issue. It works with the provided steps. When I watch your video I'm wondering how did you add Embed Image to the table because we do not have such a feature.
Could you provide more detailed steps?

ITernovtsiy:
Hi @dew326 
It's not only with tables, you can use custom tag content to reproduce it. Uploading a new video to JIRA ticket.

p.s. _regarding embed in table, we have embed images from content migrated from legacy, it's not related to this bug_

dew326:
Sorry I still cannot reproduce the issue. But on the new video, I spot one difference in the path. I have `Custom Tag > p > Embed Inline` and this is the proper path, but you have `div > p > Embed Inline` for some reason your custom tag is not detected as a custom tag and probably this is the root of the issue.

ITernovtsiy:
Hm, you're correct, there is a bug when "div" appearing in path. Attached video of how I got a div instead of "Custom Tag". Weird, but I can't reproduce it locally, only on production.

ITernovtsiy:
Sorry, my fault. The issue caused by chrome plugin (Grammarly). Can be closed.

PR #1094, opened 2019-09-25
[Behat] Moved AdminUI features

mnocon:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30884
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

The full regression builds are failing: https://travis-ci.com/ezsystems/ezplatform-page-builder/jobs/238853969

because it's not possible to separate from each other feature files from AdminUI: the "standard" ones and the new "personas"

Using definition like:
```
regression:
      suites:
          regression:
            paths:
              - '%paths.base%/vendor/ezsystems/ezplatform-admin-ui/features'
```
means that all feature files (both from "features" directory and "personas" subdirectory) will be taken into account, which causes the personas tests to be run as well.

I've decided to move the "standard" (running on admin) files to another directory, which allows us to separate them from others.

Companion PR:
https://github.com/ezsystems/ezplatform-page-builder/pull/395

Any names better than "standard" are more than welcome, coudn't find anything better 😉 

#### Checklist:
- [x] Ready for Code Review


mnocon:
@lserwatka this one can be merged, together with:
https://github.com/ezsystems/ezplatform-page-builder/pull/395

PR #1103, opened 2019-10-04
EZP-31000: Fixed selection of proper tab when using pagination with the embedded content

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31000
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

When the user enters the view of the location and there click a link to embedded content, then he is redirected to view rendered by `ez_urlalias` route. This PR adds support to `_fragment` parameter in the URL to select the proper tab.

This bug occurs for all pagination lists in content preview.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.location.tab.js
+++ b/src/bundle/Resources/public/js/scripts/admin.location.tab.js
@@ -1,5 +1,20 @@
 (function (global, doc, $) {
-    $('.ez-tabs a[href="#' + global.location.hash.split('#')[1] + '"]').tab('show');
+    const location = global.location;
+
+    /**
+     * Returns tab identifier from hash or url parameter
+     *
+     * @function createContentTypeDataMap
+     * @param {Location} location
+     * @returns {string}
+     */
+    const getTabId = (location) => {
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/scripts/admin.location.tab.js
```diff
@@ -1,5 +1,20 @@
 (function (global, doc, $) {
-    $('.ez-tabs a[href="#' + global.location.hash.split('#')[1] + '"]').tab('show');
+    const location = global.location;
+
+    /**
+     * Returns tab identifier from hash or url parameter
+     *
+     * @function createContentTypeDataMap
```
dew326:
```suggestion
     * @function getTabId
```

mikadamczyk:
As @adamwojs pointed this fix is inappropriate and the bug will be fixed in the kernel because the source of the problem is in UrlAliasGenerator.

PR #1108, opened 2019-10-15
EZP-31038: add configuration for UDW for bulk add location

ITernovtsiy:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-31038](https://jira.ez.no/browse/EZP-31038)
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Part of https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/219
Co-Authored-By: @dew326 

#### Checklist:
- [x] Ready for Code Review


src/bundle/Resources/views/themes/admin/content/location_view.html.twig
```diff
@@ -80,12 +80,14 @@
...
                                 <h2 class="ez-fieldgroup__name">{{ 'content.view.subitems'|trans|desc('Sub-items') }}</h2>
+                                {% set udwConfigSingle = ez_udw_config('single_container', {}) %}
                                 <div class="ez-sil"
                                     data-location="{{ location.id }}"
                                     data-mfu-create-permissions-config="{{ subitems_module.content_create_permissions_for_mfu|json_encode() }}"
                                     data-items="{{ subitems_module.items }}"
                                     data-content-types="{{ subitems_module.content_type_info_list }}"
-                                    data-udw-config-bulk-move-items="{{ ez_udw_config('single_container', {}) }}"
+                                    data-udw-config-bulk-move-items="{{ udwConfigSingle }}"
+                                    data-udw-config-bulk-add-location="{{ udwConfigSingle }}"
```
dew326: You have to pass it to the subitems like the `data-udw-config-bulk-move-items` is passed in https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/bundle/Resources/public/js/scripts/admin.location.view.js#L167

ITernovtsiy: Done, but now I'm not 100% sure I understand why we need both `data-udw-config-bulk-move-items` and `data-udw-config-bulk-add-location`  attributes if they are always the same

dew326: `Add` and `move` are different actions and deserves a different config. The JS/React must be ready to take separate configs. If the end developer will need to have separate configs he can easily override this in the template (by changing the config key or adding context and write a subscriber). 

If we would use one config for both and wanted to split it when needed the flow probably would be probably like this:
end developer needs separate config -> contact our support -> support create a ticket -> eZ developer change the behavior (with the BC so with some fallback) -> QA -> patch for the client
if we do it now we have flow:
end developer needs separate config -> contact our support-> support gives the answer. And we do not end up with some fallbacks and technical debt in order to have BC.

So this is done for less maintenance and to give more flexibility for developers.

ITernovtsiy: ok, thanks for the clarification

ITernovtsiy:
all requested changes completed
failed tests are not related to this PR

lserwatka:
@dew326 could you make a final review so we can send it to QA?

dew326:
@lserwatka this PR looks ok, but we have to wait for the related PR (https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/219). There is an open question where this should go 2.5 or 3.0.

lserwatka:
Master is fine. So I would leave as it is.

katarzynazawada:
Error when clearing cache:
`Unrecognized option "limitation_value_templates" under "ezpublish.system.default"` 
I suppose, rebase is needed.

dew326:
@katarzynazawada done

ITernovtsiy:
added "create-location" icon requested in https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/219

PR #1113, opened 2019-10-18
Add workflow scenarios

mnocon:
JIRA: https://jira.ez.no/browse/EZEE-2899

Related PRs:
https://github.com/ezsystems/ezplatform-page-builder/pull/412
https://github.com/ezsystems/ezplatform-ee/pull/112
https://github.com/ezsystems/ezplatform-workflow/pull/79

This PR:
- adds handling of User notification sections
- changes simple table so that I could use it with Workflow tables in Admin panel
- allows to switch tabs on Dashboard page

### DIFF
--- a/src/lib/Behat/BusinessContext/UserNotificationContext.php
+++ b/src/lib/Behat/BusinessContext/UserNotificationContext.php
@@ -1,0 +1,38 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformAdminUi\Behat\BusinessContext;
+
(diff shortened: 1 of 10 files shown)


src/lib/Behat/BusinessContext/UserNotificationContext.php
```diff
@@ -0,0 +1,38 @@
...
+     * @Given there is an unread notification for current user
+     */
+    public function thereIsNotificationForCurrentUser()
+    {
+        $upperMenu = ElementFactory::createElement($this->utilityContext, UpperMenu::ELEMENT_NAME);
+        Assert::assertGreaterThan(0, $upperMenu->getNotificationsCount());
+    }
+
+    /**
+     * @Given I go to item with details:
```
m-tyrala: I don't like this step - there's nothing here that would say that it'll be done through notification popup. It'll be confusing in `feature` files.

PR #1114, opened 2019-10-18
EZP-31056: Link is added to the text in RTE despite not choosing any content via UDW

konradoboza:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-31056](https://jira.ez.no/browse/EZP-31056)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Excerpt from JIRA:
>The customer reported an issue with adding links to a text in the Rich Text Editor. The link is added even though no content was selected via UDW.

#### Checklist:
- [x] Ready for Code Review


dew326:
I don't think this will be a good approach. After cancel UDW it should show the link edit toolbar.

I think this fix will remove always the link. Please test this flow:
1. Add a link (with UDW confirmation and save the link)
2. Edit link
3. Open UDW
4. Cancel UDW selection

I think this will remove the existing link, what will be unwanted behavior.

konradoboza:
@dew326 it should work now the way you described.

konradoboza:
@katarzynazawada should work now. I also discovered and fixed the same issue when adding links to images, please also test this scenario.

katarzynazawada:
JS error is fixed but issue still occurs in Text block in Page Builder: https://recordit.co/r6sNCUlana

konradoboza:
@katarzynazawada the fix for the issue happening in Page Builder needed to be distributed as a separate PR: https://github.com/ezsystems/ezplatform-page-builder/pull/419.

konradoboza:
Added firing `customUpdate` event after removing the link by clicking outside RTE field. This issue was reported in the other PR: https://github.com/ezsystems/ezplatform-page-builder/pull/420#issuecomment-550975121. @dew326 could you please take a look?

konradoboza:
@lserwatka please merge.

PR #1123, opened 2019-10-30
fix: wrong allowed content types list for object relation(s) fields

florianalexandre:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31089
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/UniversalDiscovery/Event/Subscriber/ObjectRelationAllowedContentTypes.php
+++ b/src/lib/UniversalDiscovery/Event/Subscriber/ObjectRelationAllowedContentTypes.php
@@ -42,12 +42,16 @@ class ObjectRelationAllowedContentTypes implements EventSubscriberInterface
             return;
         }
 
-        $config['content_on_the_fly']['allowed_content_types'] = array_unique(
-            array_merge(
-                $config['content_on_the_fly']['allowed_content_types'],
-                $context['allowed_content_types']
-            )
-        );
+        if (!empty($config['content_on_the_fly']['allowed_content_types'])) {
+            $config['content_on_the_fly']['allowed_content_types'] = array_values(
+                array_intersect(
+                    $config['content_on_the_fly']['allowed_content_types'],
+                    $context['allowed_content_types']
+                )
+            );
+        } else {
+            $config['content_on_the_fly']['allowed_content_types'] = $context['allowed_content_types'];
+        }
 
         $event->setConfig($config);
     }


src/lib/UniversalDiscovery/Event/Subscriber/ObjectRelationAllowedContentTypes.php
```diff
@@ -43,10 +43,10 @@ public function onUdwConfigResolve(ConfigResolveEvent $event): void
         }
 
         $config['content_on_the_fly']['allowed_content_types'] = array_unique(
```
Nattfarinn: `array_unique` is no longer needed here if we use `array_intersect`.

src/lib/UniversalDiscovery/Event/Subscriber/ObjectRelationAllowedContentTypes.php
```diff
@@ -43,10 +43,10 @@ public function onUdwConfigResolve(ConfigResolveEvent $event): void
...
-            array_merge(
+            array_values(array_intersect(
```
Nattfarinn: Be aware, that empty `$config['content_on_the_fly']['allowed_content_types']` array resolves as "Every Content Type is allowed". `array_intersect` will result in empty array in that case, hence will not filter the Content Types by field definition.

We'll need early return of `$context['allowed_content_types']` in case empty configuration array so we don't do intersect at all.

florianalexandre: Ok, made a change to handle this scenario

src/lib/UniversalDiscovery/Event/Subscriber/ObjectRelationAllowedContentTypes.php
florianalexandre: removed

micszo:
@florianalexandre please update PR summary and commit message with JIRA ticket.

micszo:
Testing discovered: https://jira.ez.no/browse/EZP-31202

Nattfarinn:
@florianalexandre There's an issue we have to handle on our side (case of empty array meaning "everything"), so we'll take it from here. Here's updated PR with additional fix: https://github.com/ezsystems/ezplatform-admin-ui/pull/1155

PR #1127, opened 2019-11-07
EZP-31093: Introduced strict types for TrashService

adamwojs:
| Question           | Answer
| **JIRA issue**     | [EZP-31093](https://jira.ez.no/browse/EZP-31093)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** |  `master` 
| **BC breaks**      | yes
| **Tests pass**     | yes
| **Doc needed**     | yes

Fixed https://github.com/ezsystems/ezpublish-kernel/pull/2851#pullrequestreview-312507417.

#### Checklist:
- [X] Ready for Code Review

### DIFF
--- a/src/lib/Form/DataTransformer/TrashItemTransformer.php
+++ b/src/lib/Form/DataTransformer/TrashItemTransformer.php
@@ -65,8 +65,12 @@ class TrashItemTransformer implements DataTransformerInterface
             return null;
         }
 
+        if (!ctype_digit($value)) {
+            throw new TransformationFailedException('Expected a numeric string.');
+        }
+
         try {
-            return $this->trashService->loadTrashItem($value);
+            return $this->trashService->loadTrashItem((int)$value);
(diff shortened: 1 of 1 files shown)


src/lib/Form/DataTransformer/TrashItemTransformer.php
```diff
@@ -66,7 +66,7 @@ public function reverseTransform($value): ?TrashItem
         }
```
mikadamczyk: It would be nice to add here 
```
        if (!ctype_digit($value)) {
            throw new TransformationFailedException('Expected a numeric string.');
        }
```

adamwojs: Fixed in https://github.com/ezsystems/ezplatform-admin-ui/pull/1127/commits/43fc63f6bbcfe91a8af66f3a4f73ee339db0d19d

PR #1161, opened 2019-12-12
EZEE-2738: Hid previous location while dragging in D&D

GrabowskiM:
| Question      | Answer
| ------------- | ---
| Tickets       |https://jira.ez.no/browse/EZEE-2738
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Hid previous location while dragging element;
Changed Draggable methods to public;

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/core/draggable.js
+++ b/src/bundle/Resources/public/js/scripts/core/draggable.js
@@ -9,22 +9,11 @@
         attachCustomEventHandlersToItem: () => {},
     };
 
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/scripts/core/draggable.js
```diff
@@ -9,22 +9,11 @@
         attachCustomEventHandlersToItem: () => {},
```
dew326: Now when we can override methods in this class those "after" callback are not needed.

PR #1177, opened 2020-01-02
EZP-30761: added ellipsize to strings

GrabowskiM:
| Question      | Answer
| ------------- | ---
| Tickets       |https://jira.ez.no/browse/EZP-30761
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Added ellipsize to strings in admin ui
Also:
https://github.com/ezsystems/ezplatform-workflow/pull/102
https://github.com/ezsystems/date-based-publisher/pull/147

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_breadcrumb.scss
+++ b/src/bundle/Resources/public/scss/_breadcrumb.scss
@@ -4,6 +4,9 @@
 
 .breadcrumb-item {
     cursor: not-allowed;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
     
     + .breadcrumb-item::before,
     &.active {
--- a/src/bundle/Resources/public/scss/_breadcrumb.scss
+++ b/src/bundle/Resources/public/scss/_breadcrumb.scss
@@ -7,6 +7,7 @@
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
+    max-width: calculateRem(300px);
     
     + .breadcrumb-item::before,
     &.active {
--- a/src/bundle/Resources/public/scss/_field-group.scss
+++ b/src/bundle/Resources/public/scss/_field-group.scss
@@ -93,6 +93,18 @@
             background-color: $ez-white;
             padding: calculateRem(16px) calculateRem(24px);
             border-radius: 0 0 calculateRem(5px) calculateRem(5px);
+
+            .ezstring-field {
+                display: inline-block;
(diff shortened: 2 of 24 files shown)


src/bundle/Resources/public/scss/_breadcrumb.scss
```diff
@@ -4,6 +4,9 @@
 
 .breadcrumb-item {
     cursor: not-allowed;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
```
dew326: When we add ellipsis we also should add `title` attribute to make sure that people can read the whole text.
This one applies to all places where you added the ellipsis

src/bundle/Resources/public/scss/_field-group.scss
```diff
@@ -93,6 +93,18 @@
...
             border-radius: 0 0 calculateRem(5px) calculateRem(5px);
+
+            .ezstring-field {
+                display: inline-block;
+                max-width: 100%;
+                overflow: hidden;
+                text-overflow: ellipsis;
+            }
+
+            .ezrichtext-field p {
```
dew326: We should avoid styling by tag name also this will solve the problem only for the paragraph what about headings and other stuff?

src/bundle/Resources/views/admin/content_type/list.html.twig
```diff
@@ -72,13 +72,13 @@
...
                 </td>
-                <td class="ez-table__cell ez-table__cell--after-icon">
+                <td class="ez-table__cell ez-table__cell--name ez-table__cell--after-icon">
                 {% set view_url = path('ezplatform.content_type.view', {
                         'contentTypeGroupId': content_type_group.id,
                         'contentTypeId': content_type.id
                     }) %}
 
-                    <a href="{{ view_url }}">{{ content_type.name }}</a>
+                    <a href="{{ view_url }}" title="{{ content_type.name }}">{{ content_type.name }}</a>
```
lucasOsti: Sometimes the tag "a" is one line and sometimes multi, maybe we'll set a standard for it?

dew326: It goes multi-line when the line is too long ;)

lserwatka:
Could you merge it up?

GrabowskiM:
Merge up done.

PR #1178, opened 2020-01-03
EZEE-2904: Can't delete Custom URL Alias for Form

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZEE-2904](https://jira.ez.no/projects/EZEE/issues/EZEE-2904)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/button.state.toggle.js
+++ b/src/bundle/Resources/public/js/scripts/button.state.toggle.js
@@ -1,21 +1,27 @@
-(function(global, doc) {
-    const toggleForms = [...doc.querySelectorAll('.ez-toggle-btn-state')];
+(function (global, doc) {
+    doc.addEventListener(
+        'DOMContentLoaded',
+        () => {
+            const toggleForms = [...doc.querySelectorAll('.ez-toggle-btn-state')];
 
-    toggleForms.forEach((toggleForm) => {
-        const checkboxes = [...toggleForm.querySelectorAll('.ez-table__cell.ez-table__cell--has-checkbox input[type="checkbox"]')];
-        const buttonRemove = doc.querySelector(toggleForm.dataset.toggleButtonId);
+            toggleForms.forEach((toggleForm) => {
+                const checkboxes = [...toggleForm.querySelectorAll('.ez-table__cell.ez-table__cell--has-checkbox input[type="checkbox"]')];
+                const buttonRemove = doc.querySelector(toggleForm.dataset.toggleButtonId);
 
-        if (!buttonRemove) {
-            return;
-        }
+                if (!buttonRemove) {
+                    return;
+                }
 
-        const toggleButtonState = () => {
-            const isAnythingSelected = checkboxes.some((el) => el.checked);
+                const toggleButtonState = () => {
+                    const isAnythingSelected = checkboxes.some((el) => el.checked);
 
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/scripts/button.state.toggle.js
```diff
@@ -1,21 +1,27 @@
-(function(global, doc) {
```
andrerom: Nice! 👍 

More for @dew326 to give input on regarding coding style, but technically this could also be something ala `document.addEventListener('DOMContentLoaded', function() {` now. Not that it would be important, but it would avoid the indentation change for the additional nesting, but on the other hand it will need to refer to document insted of doc inside fucntion.

mateuszbieniek:
I'm not sold on this solution - putting the whole chunk of code into the listener seems more like a workaround than a proper fix for me.

Please note that the original code works fine for any content that has no Form Field Type in it - my best guess something is failing due to additional iframe in which form renders.

mateuszdebinski:
the script works fine, but probably iframe causes slowdown loading dom through what, javascript will execute faster than the dom ready, I can use setTimeOut but I think this script should first wait for the loading dom, especially that works on dom elements

> I'm not sold on this solution - putting the whole chunk of code into the listener seems more like a workaround than a proper fix for me.
> 
> Please note that the original code works fine for any content that has no Form Field Type in it - my best guess something is failing due to additional iframe in which form renders.

But where is a problem

mateuszbieniek:
I checked this, and for some reason, this code is fired for Form Content Type when not Dom is not ready (URL panel is still not rendered) - this is not the issue for other Contents (URL panel is already rendered when this js is executed). Also, this does not look like a "slowdown" issue, as you can comment out form rendering and still you will have the same problem.

My point is that probably the real issue is that for some reason this script is executed at the wrong moment for the Form Contents and that should be fixed, not necessary by waiting for the full page to render, as this may be troublesome, especially in dev environment, when waiting for profiler to render may take some time...

mateuszbieniek:
Gave it another look:
It seems that js for the "Submissions" tab interferes in some way with js responsible for delete URL alias button?

When you comment out
```
{% block javascripts %}
    {{ encore_entry_script_tags('ezplatform-form-builder-submissions-tab-js', null, 'ezplatform') }}
{% endblock %}
```
in `src/bundle/Resources/views/themes/admin/content/tab/submissions/tab.html.twig` issue is gone.

Maybe you can find a better fix in the Form Builders submission tab .js file?

mateuszdebinski:
I find another solution to change from enableSingleRuntimeChunk to disableSingleRuntimeChunk in webpack config but in this case, some files will be repeated, the button.state.toggle.js file is added twice in admin UI and form builder

dew326:
I agree with @mateuszbieniek. This is a problem with the order of execution not with code itself.
I tested this and on 3.0 works fine. I think this is because of the version of the webpack-encore-bundle. It could be a bug when the webpack invokes the first occurrence of the file on the page instead of the last one and it was fixed in newer version of webpack-ecnore.
I would suggest trying to update the symfony encore.

mateuszdebinski:
after upgrading to 0.28.2 symfony encore everything works fine, so I close PR

PR #1180, opened 2020-01-07
EZP-31266: Changing items order in Relation List doesn't work on Firefox

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-31266](https://jira.ez.no/browse/EZP-31266)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Firefox for a long time has this error. The arrows aren't elements in their own right, so they can't be focused, therefore need to force focus on input when value changes

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezobjectrelationlist.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezobjectrelationlist.js
@@ -215,7 +215,15 @@
             return [...relationsContainer.querySelectorAll('[type="checkbox"]')];
         };
         const attachRowsEventHandlers = () => {
-            findOrderInputs().forEach((item) => item.addEventListener('blur', updateSelectedItemsOrder, false));
+            findOrderInputs().forEach((item) => {
+                item.addEventListener('blur', updateSelectedItemsOrder, false);
+                item.addEventListener('change', focusOnElement, false);
+            });
+        };
+        const focusOnElement = (event) => {
+            if(doc.activeElement !== event.target) {
+                event.target.focus();
+            }
         };
         const emptyRelationsContainer = () => {
             while (relationsContainer.lastChild) {
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezobjectrelationlist.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezobjectrelationlist.js
@@ -216,8 +216,10 @@
         };
         const attachRowsEventHandlers = () => {
             const isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
+            
             findOrderInputs().forEach((item) => {
                 item.addEventListener('blur', updateSelectedItemsOrder, false);
+
                 if (isFirefox) {
                     item.addEventListener('change', focusOnElement, false);
                 }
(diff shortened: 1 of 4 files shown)


src/bundle/Resources/public/js/scripts/fieldType/ezobjectrelationlist.js
```diff
@@ -215,7 +215,15 @@
...
         };
         const attachRowsEventHandlers = () => {
-            findOrderInputs().forEach((item) => item.addEventListener('blur', updateSelectedItemsOrder, false));
+            findOrderInputs().forEach((item) => {
+                item.addEventListener('blur', updateSelectedItemsOrder, false);
+                item.addEventListener('change', focusOnElement, false);
+            });
+        };
+        const focusOnElement = (event) => {
+            if(doc.activeElement !== event.target) {
```
dew326: please install Prettier and use it on the file

mateuszdebinski: Done

src/bundle/Resources/public/js/scripts/fieldType/ezobjectrelationlist.js
```diff
@@ -215,7 +215,15 @@
             return [...relationsContainer.querySelectorAll('[type="checkbox"]')];
         };
         const attachRowsEventHandlers = () => {
-            findOrderInputs().forEach((item) => item.addEventListener('blur', updateSelectedItemsOrder, false));
+            findOrderInputs().forEach((item) => {
+                item.addEventListener('blur', updateSelectedItemsOrder, false);
+                item.addEventListener('change', focusOnElement, false);
```
dew326: If this is only on FF I don't see a point to attach this event in all browsers. Please add a check if it is FF and attach it only in this case.

mateuszdebinski: the added condition only for FF attach

src/bundle/Resources/public/js/scripts/fieldType/ezobjectrelationlist.js
```diff
@@ -215,7 +215,18 @@
             return [...relationsContainer.querySelectorAll('[type="checkbox"]')];
         };
         const attachRowsEventHandlers = () => {
-            findOrderInputs().forEach((item) => item.addEventListener('blur', updateSelectedItemsOrder, false));
+            const isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
```
dew326:
```suggestion
            const isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
            
```

src/bundle/Resources/public/js/scripts/fieldType/ezobjectrelationlist.js
```diff
@@ -215,7 +215,18 @@
...
+            findOrderInputs().forEach((item) => {
+                item.addEventListener('blur', updateSelectedItemsOrder, false);
+                if (isFirefox) {
```
dew326:
```suggestion

                if (isFirefox) {
```

mateuszbieniek:
Why add prettier to packages.json?

mateuszdebinski:
> Why add prettier to packages.json?

Mistake, when I install prettier and I didn't notice that package.json was added as well

konradoboza:
@adamwojs please merge. :)

PR #1115, opened 2019-10-23
EZP-31020: Revised UI labels and messages

DominikaK:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31020
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes/no
| Doc needed?   | no
| License       | [GPL-2.0](https://github.com/ezsystems/ezplatform-admin-ui/blob/master/LICENSE)

Revises UI texts to:
- use consistent terminology
- use consistent phrasing and syntax
- avoid language pitfalls
- clarify some ambiguous labels and messages

Related PRs:
- https://github.com/ezsystems/date-based-publisher/pull/136
- https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/224
- https://github.com/ezsystems/ezplatform-form-builder/pull/189
- https://github.com/ezsystems/ezplatform-page-builder/pull/414
- https://github.com/ezsystems/ezplatform-richtext/pull/75
- https://github.com/ezsystems/ezplatform-user/pull/53
- https://github.com/ezsystems/ezplatform-workflow/pull/83
- https://github.com/ezsystems/flex-workflow/pull/143
- https://github.com/ezsystems/repository-forms/pull/322
- https://github.com/ezsystems/ezplatform-calendar/pull/45

#### Checklist:
- [x] Ready for Code Review

PR #1187, opened 2020-01-13
EZP-31241: Changed behaviour for showing not checked value in checkbox

GrabowskiM:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31241
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Show 'No' in content view when checkbox is empty.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/content/content_view_fields.html.twig
+++ b/src/bundle/Resources/views/content/content_view_fields.html.twig
@@ -19,9 +19,7 @@
                                 <div class="ez-content-field">
                                     <p class="ez-content-field-name">{{ fieldDefinition.name }}:</p>
                                     <div class="ez-content-field-value">
-                                        {% if ez_is_field_empty(content, fieldDefinition.identifier) and fieldDefinition.fieldTypeIdentifier is same as('ezboolean') %}
-                                            <em>{{ 'ezboolean.no'|trans|desc('No') }}</em>
-                                        {% elseif ez_is_field_empty(content, fieldDefinition.identifier) %}
+                                        {% if ez_is_field_empty(content, fieldDefinition.identifier) and fieldDefinition.fieldTypeIdentifier is not same as('ezboolean') %}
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/views/content/content_view_fields.html.twig
```diff
@@ -19,7 +19,9 @@
                                 <div class="ez-content-field">
                                     <p class="ez-content-field-name">{{ fieldDefinition.name }}:</p>
                                     <div class="ez-content-field-value">
-                                        {% if ez_is_field_empty(content, fieldDefinition.identifier) %}
+                                        {% if ez_is_field_empty(content, fieldDefinition.identifier) and fieldDefinition.fieldTypeIdentifier is same as('ezboolean') %}
+                                            <em>{{ 'ezboolean.no'|trans|desc('No') }}</em>
```
dew326: I'm not sure if this is the way. I think in the case of ezboolean the field should be rendered even if it's empty.

katarzynazawada:
Now there's small inconsistency - **No** (with capital letter) and **yes** (with lowercase letter)

GrabowskiM:
> Now there's small inconsistency - **No** (with capital letter) and **yes** (with lowercase letter)

That's inconsistency that we have in whole system, as these are translation from xliff file: https://github.com/ezsystems/ezplatform-admin-ui/blob/1.5/src/bundle/Resources/translations/fieldtypes_preview.en.xliff#L10
As I see, it's only in 2.5 version (1.5 of module), as it was fixed in 3.0: https://github.com/ezsystems/ezplatform-admin-ui/pull/1115/files#diff-63de7c5f52281012572e3a2b3685d538
So @DominikaK , should I change it in translation file? In theory it will affect whole system.

DominikaK:
True, the recent changes in labels were introduced in 3.0 only, so I don't think the change is necessary

lserwatka:
@GrabowskiM could you merge it up?

PR #1192, opened 2020-01-15
EZEE-2914: Custom inline style does not update textarea after click on

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZEE-2914](https://jira.ez.no/browse/EZEE-2914)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

AlloyEditor.ButtonStylesListItem on click does not update textarea element, because function _onClick is protected and we cannot override, so we need an override render function, for add a new function into onClick in order to update textarea

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-styleslistitem.js
+++ b/src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-styleslistitem.js
@@ -1,0 +1,36 @@
+import React from 'react';
+import AlloyEditor from 'alloyeditor';
+
+export default class EzBtnStylesListItem extends AlloyEditor.ButtonStylesListItem {
+    /**
+     * Lifecycle. Renders the UI of the button.
+     *
+     * @instance
+     * @memberof ButtonStylesListItem
+     * @method render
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-styleslistitem.js
```diff
@@ -0,0 +1,36 @@
...
+     * Lifecycle. Renders the UI of the button.
+     *
+     * @instance
+     * @memberof ButtonStylesListItem
+     * @method render
+     * @return {Object} The content which should be rendered.
+     */
+    render() {
+        const className = this.props.name === this.props.activeStyle ? 'ae-toolbar-element active' : 'ae-toolbar-element';
+        return (
```
dew326:
```suggestion

        return (
```

lserwatka:
Could you merge it up?

mateuszdebinski:
> Could you merge it up?

Done

PR #1197, opened 2020-01-16
EZP-30485: Fix to invalid anchor name

GrabowskiM:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30485
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Changed selector so it accepts all valid ids

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-anchoredit.js
+++ b/src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-anchoredit.js
@@ -85,6 +85,10 @@ export default class EzBtnAnchorEdit extends Component {
 
             container.insertAdjacentHTML('afterbegin', data);
 
+            /*
+                Using [id="value"] instead of just #value in querySelectorAll because with the latter this function
+                accepts only CSS valid id instead of valid HTML id selector
+            */
             return (
                 value &&
                 [...container.querySelectorAll(`[id="${value}"`)].every((element) => {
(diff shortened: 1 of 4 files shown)


src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-anchoredit.js
```diff
@@ -87,7 +87,7 @@ export default class EzBtnAnchorEdit extends Component {
 
             return (
                 value &&
-                [...container.querySelectorAll(`#${value}`)].every((element) => {
+                [...container.querySelectorAll(`[id="${value}"`)].every((element) => {
```
dew326: This is a good candidate for a comment on why it's done by `[id]` and no with `#`. In one year someone might be surprised why it is like this.

katarzynazawada:
Console error does not occur, but an attempt to publish content with # in anchor name ends with a validation error `Validation of XML content failed: Error in 55:0: Element section has extra content: para`. Not sure if this should be done as part of this PR or reported separately.

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-anchoredit.js
```diff
@@ -98,6 +107,16 @@ export default class EzBtnAnchorEdit extends Component {
         });
     }
 
+    isValueValid(value) {
+        const pattern = /^[A-Za-z][A-Za-z0-9\-_:\.]*$/;
```
dew326: I think it would be good to move it to the top of the file and name it `ANCHOR_PATTERN` or something like that

lserwatka:
Could you merge it up?

GrabowskiM:
Merge done, changes moved to richtext: https://github.com/ezsystems/ezplatform-richtext/pull/111

PR #1199, opened 2020-01-16
EZP-31261: Rebuild endpoints to use proper root location per configuration

Nattfarinn:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31261
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Frontend part of UDW changes (https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/251)
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/UniversalDiscoveryController.php
+++ b/src/bundle/Controller/UniversalDiscoveryController.php
@@ -21,6 +21,7 @@ use eZ\Publish\API\Repository\Values\Content\Search\SearchHit;
 use eZ\Publish\API\Repository\Values\User\Limitation;
 use EzSystems\EzPlatformAdminUi\Permission\LookupLimitationsTransformer;
 use EzSystems\EzPlatformAdminUi\Permission\PermissionCheckerInterface;
+use EzSystems\EzPlatformAdminUi\QueryType\LocationPathQueryType;
 use EzSystems\EzPlatformRest\Output\Visitor;
 use EzSystems\EzPlatformRest\Server\Values\Version;
 use Symfony\Component\HttpFoundation\JsonResponse;
@@ -61,6 +62,9 @@ class UniversalDiscoveryController extends Controller
     /** @var \EzSystems\EzPlatformAdminUi\Permission\LookupLimitationsTransformer */
     private $lookupLimitationsTransformer;
 
+    /** @var \EzSystems\EzPlatformAdminUi\QueryType\LocationPathQueryType */
+    private $locationPathQueryType;
+
     private $sortClauseClassMap = [
         self::SORT_CLAUSE_DATE_PUBLISHED => Query\SortClause\DatePublished::class,
         self::SORT_CLAUSE_CONTENT_NAME => Query\SortClause\ContentName::class,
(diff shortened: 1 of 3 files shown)


src/bundle/Controller/UniversalDiscoveryController.php
```diff
@@ -161,58 +153,67 @@ public function accordionGridViewAction(
...
-        $breadcrumbLocations = $this->getBreadcrumbLocations($location);
-        $breadcrumbLocationsLast = count($breadcrumbLocations) - 1;
+        $location = $this->locationService->loadLocation($locationId);
+
+        $locationPath = $this->getRelativeLocationPath($rootLocationId, $location->path);
+        $locationPathLast = count($locationPath) - 1;
 
-        $columnLocations = $breadcrumbLocationsLast < 2
-            ? $breadcrumbLocations
+        $columnLocations = $locationPathLast < 4
```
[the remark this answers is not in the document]
Nattfarinn: It is not. All we need to know here is if `locationPath` has more than 4 elements (because if so we need to pick relevant ones: first one and last three).

But I'll refactor this just a little bit to rely on count here and calculate (-1) last position later.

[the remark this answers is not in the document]
Nattfarinn: Done

src/bundle/Controller/UniversalDiscoveryController.php
```diff
@@ -161,58 +153,71 @@ public function accordionGridViewAction(
...
-            : [
-                $breadcrumbLocations[0], // First
-                $breadcrumbLocations[$breadcrumbLocationsLast - 1], // Before last
-                $breadcrumbLocations[$breadcrumbLocationsLast], // Last
+        $location = $this->locationService->loadLocation($locationId);
+
+        $locationPath = $this->getRelativeLocationPath($rootLocationId, $location->path);
+        $locationPathCount = count($locationPath);
+
+        if ($locationPathCount > 4) {
```
ViniTou: why `4`

src/bundle/Controller/UniversalDiscoveryController.php
```diff
@@ -161,58 +154,71 @@ public function accordionGridViewAction(
...
+                $locationPath[$locationPathLast - 1],
+                $locationPath[$locationPathLast], // Last
             ];
+        } else {
+            $columnLocations = $locationPath;
+        }
+
+        $columnLocations = array_map(function ($id) {
+            return (int)$id;
+        }, $columnLocations);
```
adamwojs: This could be moved to foreach loop body in https://github.com/ezsystems/ezplatform-admin-ui/pull/1199/files#diff-4d222e59b38dddd932c9f1ba8a217019R208

```php
foreach ($columnLocations as $columnLocationId) {
   $columnLocationId = (int)$columnLocationId;

   # ...
}
```

src/bundle/Controller/UniversalDiscoveryController.php
```diff
@@ -226,11 +232,15 @@ private function getParentLocationPath(Location $location): array
...
-            : new Query\Criterion\LocationId($this->getParentLocationPath($location));
+    private function getBreadcrumbLocations(
+        int $locationId,
+        int $rootLocationId = self::ROOT_LOCATION_ID
+    ): array {
+        $filter = $locationId === $rootLocationId
+            ? new Query\Criterion\ParentLocationId($rootLocationId)
+            : new Query\Criterion\LocationId($this->getParentLocationPath(
+                $this->locationService->loadLocation($locationId)
+            ));
```
adamwojs: IMHO you could introduce QueryType here.

webhdx: This won't work. QueryTypes are meant for Content not for Location search. Here the only optimization would be creating a new class extending `Query\Criterion\LocationId` to hide work done in `$this->getParentLocationPath()` but it isn't going to be that reusable,

PR #1205, opened 2020-01-21
EZP-31182: As an Editor I want to access Version compare from the versions table

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       |  [EZP-31182](https://jira.ez.no/browse/EZP-31182) 
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

https://github.com/ezsystems/ezplatform-content-comparison/pull/10

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/themes/admin/content/tab/versions/table.html.twig
+++ b/src/bundle/Resources/views/themes/admin/content/tab/versions/table.html.twig
@@ -3,6 +3,7 @@
 {% set is_draft = is_draft is defined and is_draft %}
 {% set is_archived = is_archived is defined and is_archived %}
 {% set is_draft_conflict = is_draft_conflict is defined and is_draft_conflict %}
+{% set custom_actions = custom_actions|default(false) %}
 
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/views/themes/admin/content/tab/versions/table.html.twig
```diff
@@ -19,7 +20,8 @@
             <th>{{ 'tab.versions.table.created'|trans()|desc('Created') }}</th>
             {% endif %}
             <th>{{ 'tab.versions.table.last_saved'|trans()|desc('Last saved') }}</th>
-            {% if is_draft or is_archived %}
+            {% block custom_actions_column_headers %}{% endblock %}
+            {% if (is_draft or is_archived) or custom_actions %}
```
mikadamczyk: This check is used twice and could be assigned to a dedicated variable

PR #1212, opened 2020-01-29
EZP-31339: Dropped default config enforcement for ContentTree parser

barw4:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31339
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Due to the nature of _ConfigResolver_ the site access specific config is resolved before the group's one which is valid behavior, however, in certain cases when _SemanticConfig_ is being built it forces the default values to be taken in consideration for specific site access and there is no fallback to group config, even though there should be one. 

#### Checklist:
- [x] Ready for Code Review


lserwatka:
Could you merge it up?

PR #1213, opened 2020-01-30
EZP-31307: As an Editor I want to configure link attribute for a custom tag

dew326:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31307
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-customtag-update.js
+++ b/src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-customtag-update.js
@@ -133,6 +133,11 @@ export default class EzBtnCustomTagUpdate extends EzWidgetButton {
         const selectLabel = Translator.trans(/*@Desc("Select:")*/ 'custom_tag_update_btn.select_btn.label', {}, 'alloy_editor');
         const separatorLabel = Translator.trans(/*@Desc("or")*/ 'custom_tag_update_btn.separator.label', {}, 'alloy_editor');
         const linkLabel = Translator.trans(/*@Desc("Link to:")*/ 'custom_tag_update_btn.link_to.label', {}, 'alloy_editor');
+        const placeholder = Translator.trans(
+            /*@Desc("Type or paste link here")*/ 'custom_tag_update_btn.link_input.placeholder',
+            {},
+            'alloy_editor'
+        );
 
         return (
             <div className="attribute__wrapper ez-ae-custom-tag__attribute-link">
@@ -159,7 +164,7 @@ export default class EzBtnCustomTagUpdate extends EzWidgetButton {
                             value={this.state.values[attrName].value}
                             onChange={this.updateValues.bind(this)}
                             data-attr-name={attrName}
-                            placeholder="Type or paste link here"
+                            placeholder={placeholder}
                         />
                     </div>
                 </div>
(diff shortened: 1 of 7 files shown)


src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-customtag-update.js
```diff
@@ -129,8 +130,41 @@ export default class EzBtnCustomTagUpdate extends EzWidgetButton {
...
+                        <label className="attribute__label form-control-label">{linkLabel}</label>
+                        <input
+                            type="text"
+                            defaultValue={config.defaultValue}
+                            required={config.required}
+                            className="attribute__input form-control"
+                            value={this.state.values[attrName].value}
+                            onChange={this.updateValues.bind(this)}
+                            data-attr-name={attrName}
+                            placeholder="Type or paste link here"
```
GrabowskiM: No translation here?

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-customtag-update.js
```diff
@@ -237,6 +271,16 @@ export default class EzBtnCustomTagUpdate extends EzWidgetButton {
         );
     }
 
+    setLinkFromUDW(attrName, item) {
+        const values = Object.assign({}, this.state.values);
```
GrabowskiM: In theory you shouldn't rely on `this.state` when modyfing state, ref: https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

src/bundle/Resources/public/js/alloyeditor/src/buttons/parts/ez-discover-content.js
```diff
@@ -0,0 +1,66 @@
...
+
+        editor._mainUI._setUIHidden = this.originalSetUIHidden;
+
+        this.buttonRef.current.focus();
+    }
+
+    cancelHandler(udwContainer) {
+        super.cancelHandler(udwContainer);
+
+        this.props.editor._mainUI._setUIHidden = this.originalSetUIHidden;
```
GrabowskiM: Aren't you modyfing `this.props` here in some way? It looks strange and incorrect

src/bundle/Resources/public/js/alloyeditor/src/buttons/parts/ez-discover-content.js
```diff
@@ -0,0 +1,66 @@
...
+
+        super.chooseContent();
+    }
+
+    render() {
+        const buttonLabel = Translator.trans(/*@Desc("Select Content")*/ 'discover_content.button.label', {}, 'alloy_editor');
+
+        return (
+            <div>
+                <button className="btn btn-secondary ez-btn-ae" ref={this.buttonRef} onClick={this.chooseContent.bind(this)}>
```
GrabowskiM: Why not bind it in constructor? As I saw, it was the way for all other places, so it's little incosistent.

(Also, by binding it here, you call bind with each rerender, which is unnecessary)

GrabowskiM (changes requested):

src/bundle/Resources/public/js/alloyeditor/src/buttons/parts/ez-discover-content.js
dew326: I know, I know. It is strange but this is needed for the AlloyEditor to stay open while we are discovering content in UDW.

src/bundle/Resources/public/js/alloyeditor/src/buttons/parts/ez-discover-content.js
dew326: Probably during development, I bind some value here but removed it so that's why this stayed here.

GrabowskiM (approved):

OstafinL (approved):

dew326:
Obsolete

PR #1214, opened 2020-01-30
EZP-31314: Added transition to animation open/close for content tree

lucasOsti:
Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31314
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_content-tree.scss
+++ b/src/bundle/Resources/public/scss/_content-tree.scss
@@ -11,12 +11,15 @@
         position: relative;
         flex: 0 0 auto;
         max-width: 0;
-        overflow-x: hidden;
         background: $ez-white;
         color: inherit;
+        transform: scaleX(0);
+        transform-origin: top left;
         transition: all 0.2s ease-out;
 
         &--expanded {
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/scss/_content-tree.scss
```diff
@@ -8,17 +8,20 @@
...
     .ez-content-tree-container {
-        display: none;
+        position: relative;
         flex: 0 0 auto;
         max-width: 0;
-        position: relative;
         background: $ez-white;
         color: inherit;
-        transition: max-width 0.2s $ez-admin-transition;
+        transform: scaleX(0);
```
GrabowskiM: Have you checked it on all browsers? As I remember, there were problems with `scale` and FF/Safari (?)

lucasOsti: In this case is ok :)

PR #1217, opened 2020-01-31
EZP-31137: Fixed styles dropdown when custom ones are in use for RTE

barw4:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31137
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    |no
| Tests pass?   |yes
| Doc needed?   |no

ez-custom-style.js for alloy editor was overriding base config styles by filtering out default styles unnecessarily if custom styles were in use.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-styleslistitem.js
+++ b/src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-styleslistitem.js
@@ -18,12 +18,23 @@ export default class EzBtnStylesListItem extends AlloyEditor.ButtonStylesListIte
                 className={className}
                 dangerouslySetInnerHTML={{ __html: this._preview }}
                 onClick={() => {
+                    this.clearEzAttributes();
                     this._onClick();
                     this.fireCustomUpdateEvent();
                 }}
                 tabIndex={this.props.tabIndex}
             />
         );
+    }
+
+    clearEzAttributes() {
+        const nativeEditor = this.props.editor.get('nativeEditor');
+        const block = nativeEditor.elementPath().block;
+        const attrsToRemove = ['ezelement', 'eztype', 'ezname'];
+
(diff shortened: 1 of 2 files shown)


barw4:
> I'm afraid it's not that simple.
> I raised this during the development and added a comment in the CS ticket.

Yes, it seems about right - from what I've understood it was done on purpose this way, to prevent unnecessary data attributes to be left after switching to one of the default styles. Couldn't we remove those attributes just before changing the style to the custom one as shown in https://github.com/ezsystems/ezplatform-admin-ui/pull/1217/commits/5aa0f2576a2b1aef7a30dcaebee16bcf0927dcb7?

src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-styleslistitem.js
```diff
@@ -18,6 +18,7 @@ export default class EzBtnStylesListItem extends AlloyEditor.ButtonStylesListIte
                 className={className}
                 dangerouslySetInnerHTML={{ __html: this._preview }}
                 onClick={() => {
+                    this.clearEzAttributes();
```
dew326: Now when you click on the same item (custom block style) it will not remove the style, as it does in other elements.

barw4: @dew326 Resolved in https://github.com/ezsystems/ezplatform-admin-ui/pull/1217/commits/b7a1abda55c4f5680717e612d9dca7a8d7d0d72d

lserwatka:
Could you merge it up?

dew326:
@barw4 can you also prepare PR for ezplatform-richtext for master?

barw4:
@dew326 should be done be tommorow

PR #1223, opened 2020-02-04
EZP-31299: Added 'search in specific language' feature

michal-myszka:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31299
| Requires | ezsystems/ezpublish-kernel#2940
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | yes

Added 'search in specific language' feature.
This allows to search also in content translations, in one of the selected search language. The "Use always available" option is ALWAYS ENABLED here.

There are two ways of processing search, dependent on chosen search engine.

**Legacy Search SQL Engine**
Searching returns always "distinct" results, so one content-object couldn't be repeated in different languages (unlike in Solr engine). Results display translation is based on chosen search language or if this translation doesn't exist then the result will be present in the "locale language".

**Solr Search Engine**
Searching results aren't always "distinct" if the query string is provided, so one content-object could be repeated in different languages on the list (unlike LSSE) (`useAlwaysAvailable` and new `excludeTranslationsFromAlwaysAvailable` options). Results display translation is based on matched translation. If query string is omitted, then the results will be "distinct" (`useAlwaysAvailable` option).

* Added dropdown in search form, if there is more than 1 language configured
* Added new choice loader ConfiguredLanguagesChoiceLoader and ConfiguredLanguagesChoiceType for providing also non-available languages

! Required changes from https://github.com/ezsystems/ezpublish-kernel/pull/2940

#### Checklist:
- [x] Test results on legacy search engine
- [x] Styling and scripting (front)
- [x] Prepare migration
- [x] Prepare SQL for fill for current data
- [x] Update tests
- [ ] Merge first https://github.com/ezsystems/ezpublish-kernel/pull/2940
- [x] Ready for Code Review

*Screens for LSSE. One more screen is in the comment with language choose popup when click on edit button from results list.

Screenshot 2020-02-14 at 10 20 09
Screenshot 2020-02-14 at 10 22 21
Screenshot 2020-02-14 at 10 22 32
Screenshot 2020-02-14 at 10 23 36
Screenshot 2020-02-14 at 10 24 25
Screenshot 2020-02-14 at 10 24 25
Screenshot 2020-02-14 at 10 24 52
Screenshot 2020-02-14 at 10 25 32
Screenshot 2020-02-14 at 10 25 45


src/bundle/Resources/config/services/forms.yml
```diff
@@ -47,6 +47,10 @@ services:
         arguments:
             $siteAccessLanguages: '$languages$'
 
+    EzSystems\EzPlatformAdminUi\Form\Type\ChoiceList\Loader\ConfiguredLanguagesChoiceLoader:
+        arguments:
+            $siteAccessLanguages: '$languages$'
```
alongosz: This is risky and currently does not work in 3.0, please inject ConfigResolver instead.

src/bundle/Resources/views/admin/search/list.html.twig
```diff
@@ -1,3 +1,7 @@
+{#
+    ! THIS FILE IS UNUSED / DEPRECATED, PLEASE REMOVE IN NEXT ITERATION !
```
alongosz: iteration? :)
Rather simply:
> This view is deprecated and will be removed in the next major version.

src/bundle/Controller/SearchController.php
```diff
@@ -212,4 +231,17 @@ public function searchAction(Request $request): Response
...
+
+    /**
+     * @param string|null $languageCode
+     *
+     * @return array
+     */
+    private function getSearchLanguageFilter(?string $languageCode): array
+    {
+        return [
+            'languages' => $languageCode ? [$languageCode] : [],
```
adamwojs:
```suggestion
            'languages' => (array)$languageCode,
```

src/lib/Form/Type/ChoiceList/Loader/ConfiguredLanguagesChoiceLoader.php
```diff
@@ -0,0 +1,103 @@
...
+            return [];
+        }
+
+        return $this->loadChoiceList($value)->getValuesForChoices($choices);
+    }
+
+    /**
+     * Sort languages based on siteaccess languages order.
+     *
+     * @return array
```
alongosz: This can be more strict, right? AFAICS:
```suggestion
     * @return \eZ\Publish\API\Repository\Values\Content\Language[]
```

src/lib/Form/Type/ChoiceList/Loader/LanguageChoiceLoader.php
```diff
@@ -8,99 +8,22 @@
 
 namespace EzSystems\EzPlatformAdminUi\Form\Type\ChoiceList\Loader;
 
-use eZ\Publish\API\Repository\LanguageService;
-use Symfony\Component\Form\ChoiceList\ArrayChoiceList;
-use Symfony\Component\Form\ChoiceList\Loader\ChoiceLoaderInterface;
-
-class LanguageChoiceLoader implements ChoiceLoaderInterface
+class LanguageChoiceLoader extends ConfiguredLanguagesChoiceLoader
```
alongosz: I'm not so sure about this one for `1.5`. It's a bit of a BC break - a change of a behavior. This loader is used for e.g. Content Create widget. So far it was possible to create Content in any system language, regardless of if it's present in `admin_group` prioritized languages list. That lead to other issues, true, but still feels a bit too much for `1.5`.

The key difference in the behavior is `array_merge($saLanguages, array_values($languagesByCode));` in `getChoiceList` method - it returns a merged list of SA languages and any other system languages. With this change, if I got it right, it would return SA languages only.

POV ping @webhdx

src/lib/Form/Type/Language/ConfiguredLanguagesChoiceType.php
```diff
@@ -0,0 +1,25 @@
...
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformAdminUi\Form\Type\Language;
+
+use EzSystems\EzPlatformAdminUi\Form\Type\ChoiceList\Loader\ConfiguredLanguagesChoiceLoader;
+
+/**
+ * Form Type allowing to select from all configured (also not enabled) Languages.
+ */
+class ConfiguredLanguagesChoiceType extends LanguageChoiceTypeAbstract
```
alongosz: The implementation of `ConfiguredLanguagesChoiceType` and `LanguageChoiceType` is almost identical, so it feels redundant. How does it happen that for `ConfiguredLanguagesChoiceType` the `ConfiguredLanguagesChoiceLoader` is injected instead? Based on the usage I don't see it (expected at least `choice_loader` option in `\EzSystems\EzPlatformAdminUi\Form\Type\Search\SearchType::buildForm`).

src/lib/Tab/Dashboard/PagerContentToDataMapper.php
```diff
@@ -74,19 +82,31 @@ public function map(Pagerfanta $pager): array
             $contributor = (new UserExists($this->userService))->isSatisfiedBy($contentInfo->ownerId)
                 ? $this->userService->loadUser($contentInfo->ownerId) : null;
 
+            $availableLanguagesCodes = $content->versionInfo->languageCodes;
+            $translationLanguageCode = ($forcedLanguage && in_array($forcedLanguage, $availableLanguagesCodes))
```
alongosz:
```suggestion
            $translationLanguageCode = (null !== $forcedLanguage && in_array($forcedLanguage, $availableLanguagesCodes))
```

src/bundle/Controller/SearchController.php
michal-myszka: If $languageCode will be empty string, the array with empty value will be created:

`$languageCode = '';

var_dump(
    (array)$languageCode
);`

Result:

`array(1) {
  [0]=>
  string(0) ""
}
`

src/lib/Form/Type/ChoiceList/Loader/LanguageChoiceLoader.php
michal-myszka: This is a mistake, it should be backward compatibility. I've updated the logic in ConfiguredLanguagesChoiceLoader.

src/lib/Form/Type/Language/ConfiguredLanguagesChoiceType.php
michal-myszka: `ConfiguredLanguagesChoiceType` and `LanguageChoiceType` have injected different ChoiceLoaders, with a little different strategy. Above classes are autowiring by typehint.

dew326:
Frontend part updated:
Zrzut ekranu 2020-02-20 o 12 25 05

@GrabowskiM @lucasOsti please review the last commit.

/cc @lserwatka

src/bundle/Resources/public/js/scripts/button.translation.edit.js
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        init() {
+            this.toggler.addEventListener('click', this.showTranslationsList, false);
+        }
+    }
+
+    const translationSelectors = doc.querySelectorAll('.ez-translation-selector');
+
+    if (!translationSelectors.length) {
```
GrabowskiM: Is it needed? After all, if length is 0, forEach won't be executed anyway

dew326:
For QA:
Here was merged also select language to edit from the dashboard.
Please test this PR here: https://github.com/ezsystems/ezplatform-admin-ui/pull/1251

src/lib/Tab/Dashboard/PagerContentToDataMapper.php
```diff
@@ -103,10 +104,14 @@ public function map(Pagerfanta $pager, ?string $forcedLanguage = null): array
                 'modified' => $content->versionInfo->modificationDate,
                 'initialLanguageCode' => $content->versionInfo->initialLanguageCode,
                 'content_is_user' => (new ContentIsUser($this->userService))->isSatisfiedBy($content),
-                'available_translations' => $this->languageService->loadLanguageListByCode(
-                    $availableLanguagesCodes
+                'available_translations' => $availableTranslations,
+                'available_enabled_translations' => array_filter(
+                    $availableTranslations,
+                    (function (\eZ\Publish\API\Repository\Values\Content\Language $translation) {
```
adamwojs: `\eZ\Publish\API\Repository\Values\Content\Language` import is missing +

```suggestion
                    (static function (Language $translation): bool {
```

src/lib/Search/PagerContentToDataMapperAbstract.php
```diff
@@ -0,0 +1,136 @@
...
+        if (false === $userExists) {
+            return null;
+        }
+
+        return $this->userService->loadUser($contentInfo->ownerId);
+    }
+
+    /**
+     * @param array $data
+     * @param array $contentTypeIds
```
adamwojs:
```suggestion
     * @param int[] $contentTypeIds
```

src/lib/Search/PagerSearchContentToDataMapper.php
```diff
@@ -0,0 +1,60 @@
...
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformAdminUi\Search;
+
+use Pagerfanta\Pagerfanta;
+
+class PagerSearchContentToDataMapper extends PagerContentToDataMapperAbstract
+{
+    /**
+     * @param Pagerfanta $pager
```
adamwojs:
```suggestion
     * @param \Pagerfanta\Pagerfanta $pager
```

src/lib/Tab/Dashboard/PagerContentToDataMapper.php
```diff
@@ -1,110 +0,0 @@
...
-use eZ\Publish\API\Repository\ContentService;
-use eZ\Publish\API\Repository\ContentTypeService;
-use eZ\Publish\API\Repository\UserService;
-use eZ\Publish\Core\Helper\TranslationHelper;
-use eZ\Publish\Core\MVC\Symfony\Locale\UserLanguagePreferenceProviderInterface;
-use EzSystems\EzPlatformAdminUi\Specification\ContentIsUser;
-use EzSystems\EzPlatformAdminUi\Specification\UserExists;
-use Pagerfanta\Pagerfanta;
-
-class PagerContentToDataMapper
```
adamwojs: Technically it's a break BC break. To avoid it the class should be deprecated in v2.5 and then dropped in 3.0.

src/lib/Form/Type/ChoiceList/Loader/ConfiguredLanguagesChoiceLoader.php
```diff
@@ -0,0 +1,105 @@
...
+    private $siteAccessLanguages;
+
+    /**
+     * @param \eZ\Publish\API\Repository\LanguageService $languageService
+     * @param \eZ\Publish\Core\MVC\ConfigResolverInterface $configResolver
+     */
+    public function __construct(LanguageService $languageService, ConfigResolverInterface $configResolver)
+    {
+        $this->languageService = $languageService;
+        $this->siteAccessLanguages = $configResolver->getParameter('languages');
```
adamwojs: `$configResolver->getParameter('languages')` should be called in https://github.com/ezsystems/ezplatform-admin-ui/pull/1223/files#diff-5f4e37227ab6d0e555d4b6703d406632R96

src/lib/Search/PagerContentToDataMapper.php
```diff
@@ -0,0 +1,50 @@
...
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformAdminUi\Search;
+
+use Pagerfanta\Pagerfanta;
+
+class PagerContentToDataMapper extends PagerContentToDataMapperAbstract
+{
+    /**
+     * @param Pagerfanta $pager
```
adamwojs:
```suggestion
     * @param \Pagerfanta\Pagerfanta $pager
```

src/lib/Search/PagerContentToDataMapper.php
```diff
@@ -0,0 +1,50 @@
...
+     * @return array
+     */
+    public function map(Pagerfanta $pager): array
+    {
+        $data = [];
+        $contentTypeIds = [];
+
+        foreach ($pager as $content) {
+            /** @var \eZ\Publish\API\Repository\Values\Content\Content $content */
+            $contentInfo = $content->getVersionInfo()->getContentInfo();
```
adamwojs: ` $content->getVersionInfo()->getContentInfo()` could be simplified to: `$content->contentInfo`

src/lib/Search/PagerContentToDataMapperAbstract.php
```diff
@@ -0,0 +1,136 @@
...
+            array_unique($contentTypeIds),
+            $this->userLanguagePreferenceProvider->getPreferredLanguages()
+        );
+
+        foreach ($data as $idx => $item) {
+            /** @var \eZ\Publish\API\Repository\Values\ContentType\ContentType $contentType */
+            // get content type from bulk-loaded list or fallback to lazy loaded one if not present
+            $contentType = isset($contentTypes[$item['contentTypeId']])
+                ? $contentTypes[$item['contentTypeId']]
+                : $item['content']->getContentType();
```
adamwojs: It could be simplified to `$contentTypes[$item['contentTypeId']] ?? $item['content']->getContentType()`. To improve readability even more you can all extract `$item['contentTypeId']` to variable.

src/lib/Form/Type/Language/ConfiguredLanguagesChoiceType.php
alongosz: > Above classes are autowiring by typehint.

Where exactly does the auto-wiring for that occur? The proper solution in this case is not to auto-wire, but inject directly the specific implementation (so single constructor in abstract class only and direct injection of `LanguageChoiceLoader` and `ConfiguredLanguagesChoiceLoader`). Still, can't find a place where it should happen. Form type is used as `<ClassName>::class` so seems like constructor is unused.

src/lib/Search/PagerContentToDataMapperAbstract.php
```diff
@@ -0,0 +1,136 @@
...
+use eZ\Publish\API\Repository\Values\Content\Content;
+use eZ\Publish\API\Repository\Values\Content\ContentInfo;
+use eZ\Publish\API\Repository\Values\Content\Language;
+use eZ\Publish\API\Repository\Values\User\User;
+use eZ\Publish\Core\Helper\TranslationHelper;
+use eZ\Publish\Core\MVC\Symfony\Locale\UserLanguagePreferenceProviderInterface;
+use EzSystems\EzPlatformAdminUi\Specification\ContentIsUser;
+use EzSystems\EzPlatformAdminUi\Specification\UserExists;
+
+class PagerContentToDataMapperAbstract
```
ViniTou: Is this class abstract or not?

src/lib/Search/PagerContentToDataMapperAbstract.php
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        return $this->userService->loadUser($contentInfo->ownerId);
+    }
+
+    /**
+     * @param array $data
+     * @param array $contentTypeIds
+     */
+    protected function setTranslatedContentTypesNames(array &$data, array $contentTypeIds): void
```
ViniTou: not rly fan of passing array by reference, but if no one protests ;)

src/lib/Search/PagerContentToDataMapperAbstract.php
```diff
@@ -0,0 +1,136 @@
...
+    /** @var \eZ\Publish\API\Repository\LanguageService */
+    private $languageService;
+
+    /**
+     * @param \eZ\Publish\API\Repository\ContentTypeService $contentTypeService
+     * @param \eZ\Publish\API\Repository\UserService $userService
+     * @param \eZ\Publish\Core\MVC\Symfony\Locale\UserLanguagePreferenceProviderInterface $userLanguagePreferenceProvider
+     * @param \eZ\Publish\Core\Helper\TranslationHelper $translationHelper
+     * @param \eZ\Publish\API\Repository\LanguageService $languageService
+     */
```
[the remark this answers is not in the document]
ViniTou: On the other hand it goes to `1.5` so ¯\_(ツ)_/¯

src/lib/Form/Type/ChoiceList/Loader/ConfiguredLanguagesChoiceLoader.php
```diff
@@ -0,0 +1,105 @@
...
+        $orderedLanguages = [];
+
+        foreach ($this->siteAccessLanguages as $saLanguageCode) {
+            if (isset($languagesAssoc[$saLanguageCode])) {
+                $orderedLanguages[] = $languagesAssoc[$saLanguageCode];
+                unset($languagesAssoc[$saLanguageCode]);
+            }
+        }
+
+        return array_merge($orderedLanguages, array_values($languagesAssoc));
```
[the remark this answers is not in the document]
michal-myszka: There shouldn't be duplicates here, please see at `unset($languagesAssoc[$saLanguageCode]);`

src/lib/Form/Type/Language/ConfiguredLanguagesChoiceType.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    /**
+     * @param \EzSystems\EzPlatformAdminUi\Form\Type\ChoiceList\Loader\ConfiguredLanguagesChoiceLoader $languageChoiceLoader
+     */
```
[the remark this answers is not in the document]
michal-myszka: There is more accurate information (prototype type-hint is only `ChoiceLoaderInterface`, because inheriting classes have different implementations)

src/lib/Search/PagerContentToDataMapperAbstract.php
michal-myszka: Yes, it is. Corrected

src/lib/Search/PagerContentToDataMapperAbstract.php
michal-myszka: I understand :) also not a rly fan of this method, but it's only moved from another place with little refactor (not more time in this task)

michal-myszka:
> > Above classes are autowiring by typehint.
> 
> Where exactly does the auto-wiring for that occur? The proper solution in this case is not to auto-wire, but inject directly the specific implementation (so single constructor in abstract class only and direct injection of LanguageChoiceLoader and ConfiguredLanguagesChoiceLoader). Still, can't find a place where it should happen. Form type is used as <ClassName>::class so seems like constructor is unused.

@alongosz the logic for getting service starts here: `\Symfony\Component\Form\Extension\DependencyInjection\DependencyInjectionExtension::getType()`

The constructors are used and I've verified, that they receive the correct dependency implementations.

src/lib/Form/Type/Language/LanguageChoiceType.php
```diff
@@ -9,41 +9,19 @@
...
-    private $languageChoiceLoader;
-
     /**
      * @param \EzSystems\EzPlatformAdminUi\Form\Type\ChoiceList\Loader\LanguageChoiceLoader $languageChoiceLoader
      */
     public function __construct(LanguageChoiceLoader $languageChoiceLoader)
     {
-        $this->languageChoiceLoader = $languageChoiceLoader;
-    }
-
```
webhdx: What you are doing here is wrong. You don't need `AbstractLanguageChoiceType`. Since you only need to change choice loader in your types, then use `choice_loader` option. Do not create X additional classes to achieve what is already handled well by Symfony.

```php
$formBuilder->add('configured_language_choice_type', LanguageChoiceType::class, [
    'choice_loader' => new ConfiguredLanguagesChoiceLoader(), // or inject from DIC
];
```

src/lib/Form/Type/Search/SearchType.php
```diff
@@ -64,6 +65,16 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'placeholder' => /** @Desc("Any time") */ 'search.any_time',
                 'mapped' => false,
             ])
+            ->add(
+                'search_language',
+                ConfiguredLanguagesChoiceType::class,
+                [
+                    'required' => false,
```
webhdx: See my above comment. Here you could just use `choice_loader` option on `LanguageChoiceType` to avoid adding very specific `ConfiguredLanguagesChoiceType` class.

alongosz:
> the logic for getting service starts here: `\Symfony\Component\Form\Extension\DependencyInjection\DependencyInjectionExtension::getType()`
> 
> The constructors are used and I've verified, that they receive the correct dependency implementations.

Given the amount of lines of code which was required for that auto-wiring to work and the static analysis issues which this led to, I would say that the best approach is to keep one `LanguageType` and configure proper Choice Loaders when building forms.

POV ping @webhdx

src/lib/Tab/Dashboard/PagerContentToDataMapper.php
```diff
@@ -10,101 +10,51 @@
...
+     * @param \eZ\Publish\API\Repository\LanguageService $languageService
      */
     public function __construct(
         ContentService $contentService,
+        TranslationHelper $translationHelper,
         ContentTypeService $contentTypeService,
         UserService $userService,
         UserLanguagePreferenceProviderInterface $userLanguagePreferenceProvider,
-        TranslationHelper $translationHelper
+        LanguageService $languageService
```
alongosz: The safest approach for BC here is not touching this class at all. Since it's no longer used, there's no issue with that. It _is_ no longer used, right? :)

the only change to the constructor should be 
```php
@trigger_error(
    sprintf(
        '%s is deprecated, use %s instead', 
        __CLASS__, 
        \EzSystems\EzPlatformAdminUi\Search\PagerContentToDataMapper::class
    ),
    E_USER_DEPRECATED
)
```
(please import new class as an alias for readability)

src/bundle/Controller/SearchController.php
```diff
@@ -55,7 +56,7 @@ class SearchController extends Controller
 
     /**
      * @param \eZ\Publish\API\Repository\SearchService $searchService
-     * @param \EzSystems\EzPlatformAdminUi\Tab\Dashboard\PagerContentToDataMapper $pagerContentToDataMapper
+     * @param \EzSystems\EzPlatformAdminUi\Search\PagerContentToDataMapper $pagerSearchContentToDataMapper
```
webhdx: Wrong type, should be `\EzSystems\EzPlatformAdminUi\Search\PagerSearchContentToDataMapper`

src/lib/Tab/Dashboard/PagerContentToDataMapper.php
```diff
@@ -10,101 +10,51 @@
...
 use eZ\Publish\Core\MVC\Symfony\Locale\UserLanguagePreferenceProviderInterface;
-use EzSystems\EzPlatformAdminUi\Specification\ContentIsUser;
-use EzSystems\EzPlatformAdminUi\Specification\UserExists;
-use Pagerfanta\Pagerfanta;
 
-class PagerContentToDataMapper
+/**
+ * @deprecated This class has been moved to \EzSystems\EzPlatformAdminUi\Search\PagerContentToDataMapper
+ */
+class PagerContentToDataMapper extends \EzSystems\EzPlatformAdminUi\Search\PagerContentToDataMapper
```
webhdx: If I'm not mistaken `\EzSystems\EzPlatformAdminUi\Search\PagerContentToDataMapper` is not used anywhere but on a dashboard tab. If so, please remove `\EzSystems\EzPlatformAdminUi\Search\PagerContentToDataMapper` and extend `AbstractPagerContentToDataMapper` here.

src/lib/Tab/Dashboard/PagerContentToDataMapper.php
```diff
@@ -10,101 +10,51 @@
...
      * @param \eZ\Publish\Core\Helper\TranslationHelper $translationHelper
+     * @param \eZ\Publish\API\Repository\LanguageService $languageService
      */
     public function __construct(
         ContentService $contentService,
+        TranslationHelper $translationHelper,
         ContentTypeService $contentTypeService,
         UserService $userService,
         UserLanguagePreferenceProviderInterface $userLanguagePreferenceProvider,
-        TranslationHelper $translationHelper
```
webhdx: Changing arguments order is still a BC break. If you need to add something you can only add optional argument to the end of the list.

src/lib/Tab/Dashboard/PagerContentToDataMapper.php
michal-myszka: This `\EzSystems\EzPlatformAdminUi\Search\PagerContentToDataMapper` is using in dashboard tabs, but the results becomes from Search engines (at least for now). I'm not sure about move AbstractPagerContentToDataMapper to dashboard "domain" because this class are extending also by `\EzSystems\EzPlatformAdminUi\Search\PagerSearchContentToDataMapper`, which is not a part of dashboard "domain", but all three classes are connected by search feature now. Maybe we should move them all to namespace like "Mapper" or "PagerContent" for more independent .. what do you think @webhdx ?

src/lib/Tab/Dashboard/PagerContentToDataMapper.php
michal-myszka: That's good idea, but new templates required some changes in this mapper to work. Maybe we should add them also here, in case someone uses the deprecated mapper?

src/lib/Form/Type/ChoiceList/Loader/ConfiguredLanguagesChoiceLoader.php
```diff
@@ -0,0 +1,106 @@
...
+use eZ\Publish\Core\MVC\ConfigResolverInterface;
+use Symfony\Component\Form\ChoiceList\ArrayChoiceList;
+use Symfony\Component\Form\ChoiceList\Loader\ChoiceLoaderInterface;
+
+class ConfiguredLanguagesChoiceLoader implements ChoiceLoaderInterface
+{
+    /** @var \eZ\Publish\API\Repository\LanguageService */
+    private $languageService;
+
+    /** @var ConfigResolverInterface */
```
adamwojs:
```suggestion
    /** @var \eZ\Publish\Core\MVC\ConfigResolverInterface */
```

lserwatka:
Could you merge it up?

PR #1225, opened 2020-02-05
EZP-30639: Used Content view instead of deprecated Location view

Nattfarinn:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-30639
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.location.view.js
+++ b/src/bundle/Resources/public/js/scripts/admin.location.view.js
@@ -87,7 +87,7 @@
             })
             .catch(eZ.helpers.notification.showErrorNotification);
     };
-    const generateLink = (locationId) => Routing.generate('_ezpublishLocation', { locationId });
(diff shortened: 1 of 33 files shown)


src/bundle/Resources/public/js/scripts/admin.location.view.js
```diff
@@ -78,7 +78,7 @@
             })
             .catch(eZ.helpers.notification.showErrorNotification);
     };
-    const generateLink = (locationId) => Routing.generate('_ezpublishLocation', { locationId });
+    const generateLink = (contentId, locationId) => Routing.generate('_ez_content_view', { contentId, locationId });
```
dew326: This is kind of BC break, the `contentId` should be added as a second parameter not first.

PR #1228, opened 2020-02-07
EZP-31311: Add queryfieldtype tests

mnocon:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31311
| Bug fix?      | no
| New feature?  | new tests
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | yes/no

Adding tests for ezcontentquery. This fieldtype is not editable in Content Edit mode:
obraz
which means I had to rewrite a bit of current test logic to take that into account.
Because of that I've also decided that we shoudn't add it to existing examples (which try to set a value in fieldtype and then verify its correctness) but as new separate feature. Scenarios boil down to:
1) A Content Item with contenqeury fieldtype can be created (and Content Query field displays correct data)
1) A Content Item with contenqeury fieldtype can be edited (and Content Query field displays correct data)

PR requires: https://github.com/ezsystems/BehatBundle/pull/126

#### Checklist:
- [x] Ready for Code Review
- [x] Remove TMP commit

### DIFF
--- a/features/standard/ContentTypeFields.feature
+++ b/features/standard/ContentTypeFields.feature
@@ -8,7 +8,7 @@ Feature: Content fields setting and editing
     Given I create a "<fieldName> CT" Content Type in "Content" with "<fieldInternalName>" identifier
       | Field Type  | Name        | Identifier          | Required | Searchable | Translatable | Settings       |
       | <fieldName> | Field       | <fieldInternalName> | no      | no	      | yes          | <fieldSettings>  |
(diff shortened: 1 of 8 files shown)


mnocon:
Yes, the the name is hardcoded here: https://github.com/ezsystems/ezplatform-admin-ui/pull/1228/files#diff-822490910b59e56b965d7166b0d053e4R104

Also will require changes after: https://jira.ez.no/browse/EZP-31378 (Query Types from https://jira.ez.no/browse/EZP-31107 will be used)

PR #1233, opened 2020-02-12
EZP-31382: As a Developer, I want to have access to userId in JS scripts

GrabowskiM:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31382
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Moved userId to to meta data

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.content.tree.js
+++ b/src/bundle/Resources/public/js/scripts/admin.content.tree.js
@@ -5,10 +5,11 @@
     const CLASS_BTN_CONTENT_TREE_EXPANDED = 'ez-btn--content-tree-expanded';
     const token = doc.querySelector('meta[name="CSRF-Token"]').content;
     const siteaccess = doc.querySelector('meta[name="SiteAccess"]').content;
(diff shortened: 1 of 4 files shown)


src/bundle/Resources/public/js/scripts/admin.content.tree.js
```diff
@@ -5,10 +5,11 @@
     const CLASS_BTN_CONTENT_TREE_EXPANDED = 'ez-btn--content-tree-expanded';
     const token = doc.querySelector('meta[name="CSRF-Token"]').content;
     const siteaccess = doc.querySelector('meta[name="SiteAccess"]').content;
+    const userId = doc.querySelector('meta[name="UserId"]').content;
```
adamwojs: Idea: Could we introduce `eZ.getCurrentUserId()`/`eZ.helper.getCurrentUserId()` to avoid code duplication? I feel this line will be in all *.js which somehow needs current user id

PR #1235, opened 2020-02-12
EZP-31381: Fixed z-index overlay covers when deleting items

lucasOsti:
Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31381
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/themes/admin/content/create/create.html.twig
+++ b/src/bundle/Resources/views/themes/admin/content/create/create.html.twig
@@ -9,6 +9,11 @@
 {% block details %}
     <div class="ez-edit-header">
         <div class="ez-edit-header__content-type-name">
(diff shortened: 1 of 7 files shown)


src/bundle/Resources/views/themes/admin/content/create/create.html.twig
```diff
@@ -70,14 +75,6 @@
     {{ knp_menu_render(content_create_sidebar_right, {'template': '@ezdesign/ui/menu/sidebar_right.html.twig'}) }}
 {% endblock %}
 
-{% block close_button %}
```
dew326: You cannot remove this block. A lot of templates extend this one and override this block `close_button` for example page builder, content on the fly and so

PR #1237, opened 2020-02-13
EZP-31310: Fixed VersionDraftConflict for content having multiple locations

mikadamczyk:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31310
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    |no
| Tests pass?   | yes/no
| Doc needed?   | yes/no
| Needs | <ul><li>https://github.com/ezsystems/ezplatform-admin-ui-modules/pull/269 </li><li>https://github.com/ezsystems/repository-forms/pull/333</li></ul>

When the user wants to edit the content that has multiple locations and occurs checking if unpublished draft exists then we need to know in which location the user tries to do this. This is needed to prepare proper links to edit versions. We should do this when it is possible (when a user is on the location view). When the user wants to edit content from a dashboard then the location is unknown. Therefore ContentEditViewBuilder must not only check the main Location but also other possible locations to which a user may have access. 

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/Content/VersionDraftConflictController.php
+++ b/src/bundle/Controller/Content/VersionDraftConflictController.php
@@ -71,7 +71,7 @@ class VersionDraftConflictController extends Controller
     public function draftHasNoConflictAction(
         int $contentId,
         string $languageCode,
-        ?int $locationId
+        ?int $locationId = null
     ): Response {
         $content = $this->contentService->loadContent($contentId);
         $contentInfo = $content->contentInfo;
--- a/src/bundle/Controller/Content/VersionDraftConflictController.php
+++ b/src/bundle/Controller/Content/VersionDraftConflictController.php
@@ -60,16 +60,20 @@ class VersionDraftConflictController extends Controller
     /**
      * @param int $contentId
      * @param string $languageCode
+     * @param int|null $locationId
      *
      * @return Response
      *
+     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
      * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
      * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
      */
-    public function draftHasNoConflictAction(int $contentId, string $languageCode): Response
-    {
+    public function draftHasNoConflictAction(
+        int $contentId,
+        string $languageCode,
+        ?int $locationId
+    ): Response {
         $content = $this->contentService->loadContent($contentId);
-        $location = $this->locationService->loadLocation($content->contentInfo->mainLocationId);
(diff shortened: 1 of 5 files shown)


src/bundle/Controller/Content/VersionDraftConflictController.php
```diff
@@ -60,16 +60,20 @@ public function __construct(
...
+     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
      * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
      * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
      */
-    public function draftHasNoConflictAction(int $contentId, string $languageCode): Response
-    {
+    public function draftHasNoConflictAction(
+        int $contentId,
+        string $languageCode,
+        ?int $locationId
```
ViniTou:
```suggestion
        ?int $locationId = null
```

src/bundle/Resources/views/content/tab/versions/table.html.twig
```diff
@@ -36,7 +36,7 @@
                 'contentId': version.contentInfo.id,
                 'versionNo': version.versionNo,
                 'language': version.initialLanguageCode,
-                'locationId': location.id
+                'locationId': location.id ?? null
```
dew326: This will not throw an exception when location is null?

src/bundle/Controller/Content/VersionDraftConflictController.php
```diff
@@ -89,6 +93,14 @@ public function draftHasNoConflictAction(int $contentId, string $languageCode):
             $versionsDataset = $this->datasetFactory->versions();
             $versionsDataset->load($contentInfo);
             $conflictedDrafts = $versionsDataset->getConflictedDraftVersions($contentInfo->currentVersionNo, $languageCode);
+            $locationId = $locationId ?? $contentInfo->mainLocationId;
+            try {
+                $location = $this->locationService->loadLocation($locationId);
+            } catch (UnauthorizedException $e) {
+                $availableLocations = $this->locationService->loadLocations($contentInfo);
```
andrerom:
```suggestion
                // Will return list of locations user has access to, or empty array if none
                $availableLocations = $this->locationService->loadLocations($contentInfo);
```

src/bundle/Resources/views/content/tab/versions/table.html.twig
mikadamczyk: I tested it 
```
{% set location = null %}
{{ dump(location.id ?? null) }}
```
end result is `null`

lserwatka:
Could you merge it up?

PR #1245, opened 2020-02-19
EZP-31371: Copying a web page link w/o protocol in a Word paragraph leads to a 404 error

konradoboza:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-31371](https://jira.ez.no/browse/EZP-31371)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

`HTTP` protocol prefix should be added on pasting links from external sources, if not specified already. This behavior is aligned with how prefixing works in RTE at the moment.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
@@ -325,6 +325,11 @@
             nativeEditor.on('change', saveRichText);
             nativeEditor.on('customUpdate', saveRichText);
             nativeEditor.on('editorInteraction', saveRichText);
+            nativeEditor.on('paste', () => {
+                setTimeout(() => {
+                    this.setLinksProtocol(container);
+                });
+            });
 
             return alloyEditor;
         }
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
```diff
@@ -325,6 +325,11 @@
             nativeEditor.on('change', saveRichText);
             nativeEditor.on('customUpdate', saveRichText);
             nativeEditor.on('editorInteraction', saveRichText);
+            nativeEditor.on('paste', () => {
+                setTimeout(() => {
```
dew326: Timeout is only for that it should be invoked `after` not `on`?
Maybe you should use the `afterPaste` event ref: https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#event-afterPaste

konradoboza: Thanks, I'll check that and adjust both PR's eventually.

konradoboza: Applied in https://github.com/ezsystems/ezplatform-admin-ui/pull/1245/commits/f8b36ca428a8d358d9712dec4abf149a21934613.

lserwatka:
Could you merge it up?

PR #1252, opened 2020-02-21
EZP-30482: No limitation to image content type when embedding image in RTE

konradoboza:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30482](https://jira.ez.no/browse/EZP-30482)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

When embedding images in RTE, there should be restriction only to the "image" Content-Type. It works fine when embedding content newly created via COTF, but UDW is still affected. This issue is not reproducible in new version of UDW in 3.0. 

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/config/universal_discovery_widget.yml
+++ b/src/bundle/Resources/config/universal_discovery_widget.yml
@@ -37,6 +37,7 @@ system:
                     multiple: false
                 richtext_embed_image:
                     multiple: false
+                    allowed_content_types: ['image']
                     content_on_the_fly:
                         allowed_content_types: ['image']
                 image_asset:


src/bundle/Resources/config/universal_discovery_widget.yml
```diff
@@ -37,6 +37,7 @@ system:
                     multiple: false
                 richtext_embed_image:
                     multiple: false
+                    allowed_content_types: ['image']
```
dew326: As far as I know, this should be copied from the `content_on_the_fly.allowed_content_types` if it's not set by this subscriber https://github.com/ezsystems/ezplatform-admin-ui/blob/1.5/src/lib/UniversalDiscovery/Event/Subscriber/DecoupleContentOnTheFlyAllowedConfiguration.php#L38. 

It was added exactly for that case - cotf allowed content type is set by the global allowed is not.
So I would suggest investigating why it's not set by the subscriber.

Summon @Nattfarinn for confirmation.

dew326: Talked with @Nattfarinn and seems like it was intended in 2.5 to get rid of this subscriber.

lserwatka:
Could you merge it up?

konradoboza:
Done.

PR #1255, opened 2020-02-21
EZP-31405 - fix embed image link removal

ITernovtsiy:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-31405](https://jira.ez.no/browse/EZP-31405)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Make sure <ezlink> not removed by CKEditor HTML parser when `<ezlink>` located above `<ezconfig>` element inside embed image element (which is the case if content migrated from legacy).

not reproducible with eZPlatform v3

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
@@ -84,6 +84,9 @@
                         for (i = 0; i !== element.attributes.length; i++) {
                             importChildNodes(newElement, element.attributes[i], false);
                         }
+                        if (element.localName === 'a' && parent.dataset.ezelement === 'ezembed') {
+                            element.setAttribute('data-cke-survive', '1');
+                        }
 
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
```diff
@@ -84,6 +84,9 @@
                         for (i = 0; i !== element.attributes.length; i++) {
                             importChildNodes(newElement, element.attributes[i], false);
                         }
+                        if (element.localName === 'a' && parent.dataset.ezelement === 'ezembed') {
```
dew326:
```suggestion

                        if (element.localName === 'a' && parent.dataset.ezelement === 'ezembed') {
```

lserwatka:
@dew326 could you merge it up?

dew326:
Done

PR #1256, opened 2020-02-24
EZP-31356: Fixed styles dropdown selection on Firefox and Safari

barw4:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31356
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

The focus event which is hardcoded into the AlloyEditor and attached to the styles dropdown list is not working correctly on Firefox and Safari. The reasoning for the solution was to implement `ButtonDropdown` component without this `focus` event, which does more harm than good and inject it into newly created `EzButtonStylesList` component which overwrites the default AlloyEditor's `ButtonStylesList` in order to prevent other components of being mutated.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-dropdown.js
+++ b/src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-dropdown.js
@@ -1,0 +1,27 @@
+import React from 'react';
+import AlloyEditor from 'alloyeditor';
+
+export default class EzBtnDropdown extends AlloyEditor.ButtonDropdown {
+    /**
+     * Lifecycle. Renders the UI of the button.
+     *
+     * @instance
+     * @memberof ButtonDropdown
+     * @method render
+     * @return {Object} The content which should be rendered.
+     */
+    render() {
+        return (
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/public/js/alloyeditor/src/buttons/ez-btn-dropdown.js
```diff
@@ -0,0 +1,27 @@
...
+                </ul>
+            </div>
+        );
+    }
+}
+
+const eZ = (window.eZ = window.eZ || {});
+
+eZ.ezAlloyEditor = eZ.ezAlloyEditor || {};
+eZ.ezAlloyEditor.EzBtnDropdown = EzBtnDropdown;
```
dew326: In 2.5 you should use `addConfig` method

barw4: @dew326  resolved in https://github.com/ezsystems/ezplatform-admin-ui/pull/1256/commits/faab7985c59d30effdca4df48c77007fbeeef391

PR #1258, opened 2020-02-24
EZP-31410: Unable to edit an article with 100+ embed content/images

SerheyDolgushev:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-31410](https://jira.ez.no/browse/EZP-31410)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

**Steps to reproduce**
1. Set up a new eZ Platform 2.5.8 installation on Platform.sh.
2. Publish a test article and embed 100+ content items/images.
3. Try to edit the test article from the previous step.

**Expected results**
The editor should be able to edit the article without any issues. All the embed content/images should be shown in Online Editor

**Actual results**
REST API request is sent for each embed content item/image. Which results in 100+ simultaneous requests. Because of Platform.sh limits, some of those requests will get 502 responses. And UI for some embed content items/images will be broken.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/alloyeditor/src/widgets/ez-embed-base.js
+++ b/src/bundle/Resources/public/js/alloyeditor/src/widgets/ez-embed-base.js
@@ -3,6 +3,8 @@ const IS_LINKED_CLASS = 'is-linked';
 const SHOW_EDIT_LINK_TOOLBAR_ATTR = 'data-show-edit-link-toolbar';
 const DATA_ALIGNMENT_ATTR = 'ezalign';
 const ENTER_KEY_CODE = 13;
+const MAX_ACTIVE_CONTENT_LOAD_REQUESTS = 20;
+let activeContentLoadRequestsCount = 0;
 
 const embedBaseDefinition = {
     defaults: {
@@ -107,6 +109,12 @@ const embedBaseDefinition = {
             return;
         }
 
+        if (activeContentLoadRequestsCount >= MAX_ACTIVE_CONTENT_LOAD_REQUESTS) {
+            setTimeout(this.initEditMode.bind(this), 1000);
+            return;
+        }
+
+        activeContentLoadRequestsCount++;
         this.loadContent(contentId);
     },
 
@@ -147,6 +155,10 @@ const embedBaseDefinition = {
         });
 
         fetch(request)
+            .then((response) => {
+                activeContentLoadRequestsCount--;
+                return response;
+            })
             .then((response) => response.json())
             .then(this.handleContentLoaded.bind(this))
             .catch((error) => window.eZ.helpers.notification.showErrorNotification(error));


src/bundle/Resources/public/js/alloyeditor/src/widgets/ez-embed-base.js
```diff
@@ -3,6 +3,8 @@ const IS_LINKED_CLASS = 'is-linked';
 const SHOW_EDIT_LINK_TOOLBAR_ATTR = 'data-show-edit-link-toolbar';
 const DATA_ALIGNMENT_ATTR = 'ezalign';
 const ENTER_KEY_CODE = 13;
+const MAX_ACTIVE_CONTENT_LOAD_REQUESTS = 20;
```
dew326: What if there will be 10 richtext fields and each will have 25 embeds inside? The problem will still occur.
We talked about this briefly with @alongosz and @Nattfarinn and think we should change behavior here to avoid the requests at all.

The information about the embedded contents should be added by the backend into some data-attribute as a list of content objects. The alloyEditor could use that information on init time to show the needed content name and avoid requests

SerheyDolgushev: @dew326 sorry, but I would assume that `MAX_ACTIVE_CONTENT_LOAD_REQUESTS` will be for all the rich text fields, but not per each individually. And it might be a good temporary solution. Because getting that info from the back-end is more complex change. Taking into account that a similar approach is used in v3, but GraphQL is used instead of REST.

Nattfarinn: @SerheyDolgushev We're talking about changing the approach in general, so we don't have to do X number of request at all. It still may happen, no matter what number you'll put in `MAX_ACTIVE_CONTENT_LOAD_REQUESTS`, that you'll exceed server limits (because of other request done to server by other fieldtypes or features) so it is better to avoid it in a first place.

dew326: >Taking into account that a similar approach is used in v3, but GraphQL is used instead of REST.

Can you show me where it is done?

SerheyDolgushev: I mean v3 uses GraphQL instead of the REST, but it also fetches info using separate request per each embed content/image.

SerheyDolgushev: @bdunogier please correct me if I`m wrong.

bdunogier: I'm gonna have to correct you unfortunately @SerheyDolgushev : GraphQL is only used by the sub-items list.

SerheyDolgushev:
Closing this PR in favor of https://gitlab.com/contextualcode/ezplatform-alloyeditor-requests-limit.

PR #1286, opened 2020-03-17
EZP-31495: As an Editor I want be able to format text inside table cells

SerheyDolgushev:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-31495](https://jira.ez.no/browse/EZP-31495)
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Right now there is no way to format the text inside table cells. It is because there is no way to add a new paragraph into the table cell.

This PR provides a way to insert paragraphs into the table cell:
insert_para_into_tablecell

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/encore/ez.js.config.js
+++ b/src/bundle/Resources/encore/ez.js.config.js
@@ -26,6 +26,7 @@ const alloyEditor = [
     path.resolve(__dirname, '../public/js/alloyeditor/src/buttons/ez-btn-anchor.js'),
     path.resolve(__dirname, '../public/js/alloyeditor/src/buttons/ez-btn-anchoredit.js'),
     path.resolve(__dirname, '../public/js/alloyeditor/src/buttons/ez-btn-paragraph.js'),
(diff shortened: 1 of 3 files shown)


SerheyDolgushev:
@inakijv @SylvainGuittard

dew326:
I'm ok with exposing this into the toolbar. Especially that this is already available, the user has to be inside the table cell and hit enter - this will add a paragraph into the table cell.

But the last word is up to PMs if we want to introduce this in that way.

dew326:
Ok, I looked closer to this and what is the reason for multiple buttons?
You can achieve the same by clicking on the plus button and paragraph. Why we need a second button for the same action?

SylvainGuittard:
@SerheyDolgushev and I just had a chat. We can confirm that when hitting `enter`, it creates a new paragraph.
It's the same behavior outside a table, hitting `enter` creates a new paragraph. We don't see the need of another button.
Thanks @SerheyDolgushev for your time.
Closing this PR

PR #1261, opened 2020-02-26
EZP-29948: Improved dates select on the search content site

lucasOsti:
Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-29948
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no
| License       | [GPL-2.0](https://github.com/ezsystems/ezplatform-admin-ui/blob/master/LICENSE)

#### Checklist:
- [x] Coding standards (`$ composer fix-cs`)
- [x] Ready for Code Review

PR #1304, opened 2020-03-23
fix: Missing translations

Nattfarinn:
| Question      | Answer
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

~This is output from `translation:extract`. Please, take a look if it's Ok (like, `search.date.range` moved from `search` domain to `messages`?)~

`search.date.range` moved back to `search` domain by @ViniTou

### DIFF
--- a/src/bundle/Resources/translations/messages.en.xliff
+++ b/src/bundle/Resources/translations/messages.en.xliff
@@ -289,11 +289,6 @@
         <target state="new">Any time</target>
         <note>key: search.any_time</note>
       </trans-unit>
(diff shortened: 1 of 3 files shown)


ViniTou:
Those translations changed domain when they were moved in this pr:
https://github.com/ezsystems/ezplatform-admin-ui/pull/1261/files

we can explicitly state domain to keep them in old one, but not sure if we want to have multiple domains in one view. On the other hand changing them just cuz they move to another file also dosent seems right.

@Nattfarinn 
@lucasOsti

PR #1307, opened 2020-03-23
EZP-31459: [UDW] Allowed Content Types are not respected in object relation field

konradoboza:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-31459](https://jira.ez.no/browse/EZP-31459)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

If `allowed_content_types` key is not specified in the configuration (`null` value) the whole subscriber stops at this point even though the next condition provides proper handling for that case.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/UniversalDiscovery/Event/Subscriber/ObjectRelationAllowedContentTypes.php
+++ b/src/lib/UniversalDiscovery/Event/Subscriber/ObjectRelationAllowedContentTypes.php
@@ -36,7 +36,7 @@ class ObjectRelationAllowedContentTypes implements EventSubscriberInterface
         }
 
         if (
-            !isset($context['type'], $context['allowed_content_types'], $config['allowed_content_types'])
+            !isset($context['type'], $context['allowed_content_types'])
             || 'object_relation' !== $context['type']
         ) {
             return;
--- a/src/lib/UniversalDiscovery/Event/Subscriber/ObjectRelationAllowedContentTypes.php
+++ b/src/lib/UniversalDiscovery/Event/Subscriber/ObjectRelationAllowedContentTypes.php
@@ -54,7 +54,9 @@ class ObjectRelationAllowedContentTypes implements EventSubscriberInterface
                 ? null
                 : $intersection;
         } else {
-            $config['allowed_content_types'] = $context['allowed_content_types'];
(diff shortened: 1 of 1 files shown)


konradoboza:
In case if both `config` and `context` set during creating/editing CT are empty the resulting value is `[]` which means none CTs available in the relation list. It needs to be set as `null` instead.

src/lib/UniversalDiscovery/Event/Subscriber/ObjectRelationAllowedContentTypes.php
```diff
@@ -54,7 +54,9 @@ public function onUdwConfigResolve(ConfigResolveEvent $event): void
                 ? null
                 : $intersection;
         } else {
-            $config['allowed_content_types'] = $context['allowed_content_types'];
+            $config['allowed_content_types'] = empty($context['allowed_content_types'])
```
Nattfarinn: Are you sure about `empty` here? `[]` is valid value (meaning context disallow any content type) and it will resolve to `null` now (meaning every content type is allowed).

konradoboza: In the Back Office there are two valid values for `allowed_content_types`:
- `[]` resulting in no available CT to select in UDW,
- `null` resulting in every CT available to select in UDW.

In the FieldType edition view by default, there is no CT selected which stays in contradiction with the label `Allowed Content Types`. So in short: no allowed content type selected means all are available. That was the case in 2.5 at least, so I wanted to keep it consistent. That's why we need `null` value in this case.

ViniTou: If I read this correctly, frontend part sends `[]` when nothing is selected, but then it want's `null` for that same behaviour?

Shouldnt it be more distinct on UDW part?
@dew326 
@Nattfarinn

Nattfarinn: Confirmed, this is OK because of how FieldType works.

PR #1318, opened 2020-03-27
EZP-30881: Cannot add table after list in RTE

konradoboza:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-30881](https://jira.ez.no/browse/EZP-30881)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Currently, placing tables inside list items produces validation errors. Moreover, the table button isn't present in the inline toolbar so disabling it in the embed toolbar seems to be the way to go. PR for 3.0 will be created once this solution is approved. 

#### Checklist:
- [x] Ready for Code Review


konradoboza:
This should be most likely approached differently, probably on the XSLT transformations level. Closing.

PR #1058, opened 2019-08-14
Use TranslatorInterface for typehint

MarioBlazek:
| Question      | Answer
| ------------- | ---
| Tickets       | N/A
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

In some cases, autowire fails with following error:
`[Symfony\Component\DependencyInjection\Exception\RuntimeException]                                                                                                                                                                         
  Cannot autowire service "EzSystems\EzPlatformAdminUi\UI\Config\Mapper\FieldType\RichText\CustomTag": argument "$translator" of method "__construct()" references class "Symfony\Component\Translation\Translator" but no such service exi  
  sts. Try changing the type-hint to "Symfony\Component\Translation\TranslatorInterface" instead.                                           `

PR #1335, opened 2020-04-07
EZP-31547: Autowiring fails for CustomTag service when using 3rd party translator bundle

mateuszbieniek:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31547
| Bug fix?      | yes

This is continuation for https://github.com/ezsystems/ezplatform-admin-ui/pull/1058

This PR adds `$translator` argument to `EzSystems\EzPlatformAdminUi\UI\Config\Mapper\FieldType\RichText\CustomTag` to prevent autowiring issues.

### DIFF
--- a/src/lib/Tests/UI/Config/Mapper/FieldType/RichText/CustomTagTest.php
+++ b/src/lib/Tests/UI/Config/Mapper/FieldType/RichText/CustomTagTest.php
@@ -13,7 +13,8 @@ use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Asset\Packages;
 use Symfony\Component\Translation\MessageCatalogueInterface;
-use Symfony\Component\Translation\Translator;
+use Symfony\Component\Translation\TranslatorBagInterface;
+use Symfony\Component\Translation\TranslatorInterface;
 
 /**
  * UI Config Mapper test for RichText Custom Tags configuration.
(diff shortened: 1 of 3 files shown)


mateuszbieniek:
As mentioned by @webhdx https://github.com/ezsystems/ezplatform-admin-ui/pull/1007#discussion_r290667528 here, the provided solution seems like overkill, but without we will encounter issues like `EZP-31547`. 

Another alternative would be solution used by [Symfony](https://github.com/symfony/translation/blob/master/DataCollectorTranslator.php#L29), but I'm not sold.
WDYT?

src/lib/Tests/UI/Config/Mapper/FieldType/RichText/CustomTagTest.php
```diff
@@ -32,6 +32,7 @@ public function testMapConfig(array $customTagsConfiguration, array $enabledCust
         $mapper = new CustomTag(
             $customTagsConfiguration,
             $this->getTranslatorMock(),
+            $this->getTranslatorMock(),
```
ViniTou, line 35: Is there any way that those could be split into `TranslatorInterface` mock and `TranslatorBagInterface`?

mateuszbieniek: Sure: https://github.com/ezsystems/ezplatform-admin-ui/pull/1335/commits/8c72a020245bf22ef482cb97aaa3bdabc65b211f

mateuszbieniek:
Merge up for `ezplatform-richtext`:
https://github.com/ezsystems/ezplatform-richtext/pull/142

PR #1338, opened 2020-04-08
EZP-31539: Link Validation error in Richtext Editor

konradoboza:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-31539](https://jira.ez.no/browse/EZP-31539)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

There are still unhandled cases when pasting links in RTE (`mailto`, `skype`, `callto` etc). Instead of adding more of restricted keywords (which could also be error-prone if the keyword is a present outside schema), I changed regexp to verify whether any schema was present before changing the actual URL. It should also work properly when someone uses some custom link schema, like `myschema://`.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
@@ -398,11 +398,10 @@
             const links = container.querySelectorAll('a');
             const anchorPrefix = '#';
             const protocolPrefix = 'http://';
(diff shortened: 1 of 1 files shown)


konradoboza:
Added handling for links like `mailto:foo@bar.com` as there is no `//` in https://github.com/ezsystems/ezplatform-admin-ui/pull/1338/commits/da24931d5019fadbbe76353815bc05c75635c534.

PR #1342, opened 2020-04-09
EZP-31555: Changed max-height and added overflow to better fit content

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-31555](https://jira.ez.no/browse/EZP-31555)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

After changing the menu theme in eZ Platform v3.0 if has too many CT's then the menu is not correctly displayed

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_extra-actions.scss
+++ b/src/bundle/Resources/public/scss/_extra-actions.scss
@@ -121,7 +121,7 @@
 
     &--create {
         #{$block}__content {
-            max-height: 90%;
+            max-height: calc(100% - #{calculateRem(60px)});
             overflow: auto;
         }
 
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/scss/_extra-actions.scss
```diff
@@ -121,7 +121,8 @@
 
     &--create {
         #{$block}__content {
-            max-height: initial;
+            max-height: 90%;
```
dew326, line 124: Why "90%"? There might be some resolution where 90% will be more than the available space. I would say it should be `calc(100% - Xpx)`.

PR #1376, opened 2020-05-15
EZP-31576: When in COTF make UDW & consecutive COTF views take the full view

GrabowskiM:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31576
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review


dew326:
Here I'm not sure if we should hide the cancel and save button when the user enters the create form in the inside UDW or we just do it when a user opens the second UDW inside, I think the latter solution is more intuitive and in line with Form builder for example. 
This is something for @inakijv and @SylvainGuittard

SylvainGuittard:
@dew326 I agree with you. I think it makes sense to hide the buttons when user opens a second UDW. It should also be the case if the user opens a third one and so on.

inakijv:
I agree @dew326, second option's transition is consistent with the existing case in Form Builder. I also agree with @SylvainGuittard's proposal.

lserwatka:
You can merge it up.

PR #1381, opened 2020-05-19
EZP-31439: Harmonize tooltips

lucasOsti:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31439
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x ] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/themes/admin/link_manager/list.html.twig
+++ b/src/bundle/Resources/views/themes/admin/link_manager/list.html.twig
@@ -28,7 +28,7 @@
                 }}) }}
 
                 <span class="input-group-btn">
-                    <button class="btn btn-primary" title="{{ 'url.search'|trans|desc("Search") }}">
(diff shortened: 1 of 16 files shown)


src/bundle/Resources/views/themes/admin/link_manager/list.html.twig
```diff
@@ -28,7 +28,7 @@
                 }}) }}
 
                 <span class="input-group-btn">
-                    <button class="btn btn-primary" title="{{ 'url.search'|trans|desc("Search") }}">
```
dew326, line 31: Shouldn't this be removed from extracted translations?

lucasOsti: This translation is used below

PR #1384, opened 2020-05-20
EZEE-3020: Close tooltips when item is deleted

lucasOsti:
Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZEE-3020
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

Added auto hide of tooltips if tooltip node has been deleted

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/helpers/tooltips.helper.js
+++ b/src/bundle/Resources/public/js/scripts/helpers/tooltips.helper.js
@@ -1,5 +1,18 @@
-(function(global, doc, eZ, $) {
+(function (global, doc, eZ, $) {
     const TOOLTIPS_SELECTOR = '[title]';
+    const observerConfig = {
+        childList: true,
+        subtree: true,
+    };
+    const observer = new MutationObserver((mutationsList) => {
+        mutationsList.forEach((mutation) => {
+            const showedTooltipNode = doc.querySelector('.ez-tooltip.show');
+
+            if (mutation.removedNodes.length && showedTooltipNode) {
+                showedTooltipNode.remove();
+            }
+        });
+    });
     const parse = (baseElement = doc) => {
         if (!baseElement) {
             return;
@@ -40,6 +53,8 @@
             $(tooltipNode).tooltip('hide');
         }
     };
+
+    observer.observe(doc.querySelector('body'), observerConfig);
 
     eZ.addConfig('helpers.tooltips', {
         parse,


src/bundle/Resources/public/js/scripts/helpers/tooltips.helper.js
```diff
@@ -1,5 +1,17 @@
-(function(global, doc, eZ, $) {
+(function (global, doc, eZ, $) {
     const TOOLTIPS_SELECTOR = '[title]';
+    const config = {
+        attributes: true,
```
dew326, line 4: Correct me if I'm wrong but I think we don't have to observe the `attributes`?

lucasOsti: yup, you have right

src/bundle/Resources/public/js/scripts/helpers/tooltips.helper.js
```diff
@@ -1,5 +1,17 @@
-(function(global, doc, eZ, $) {
+(function (global, doc, eZ, $) {
     const TOOLTIPS_SELECTOR = '[title]';
+    const config = {
+        attributes: false,
```
dew326, line 4: I think if it is false you can omit the option

src/bundle/Resources/public/js/scripts/helpers/tooltips.helper.js
```diff
@@ -1,5 +1,17 @@
...
+        childList: true,
+        subtree: true,
+    };
+    const observer = new MutationObserver((mutationsList, observer) => {
+        mutationsList.forEach((mutation) => {
+            if (mutation.removedNodes.length && doc.querySelector('.ez-tooltip.show')) {
```
GrabowskiM, line 10: `doc.querySelector('.ez-tooltip.show')` could be put into variable, as it's used two times

src/bundle/Resources/public/js/scripts/helpers/tooltips.helper.js
```diff
@@ -1,5 +1,17 @@
-(function(global, doc, eZ, $) {
+(function (global, doc, eZ, $) {
     const TOOLTIPS_SELECTOR = '[title]';
+    const config = {
+        attributes: true,
+        childList: true,
+        subtree: true,
+    };
+    const observer = new MutationObserver((mutationsList, observer) => {
```
GrabowskiM, line 8: Unnecessary second arg `observer` (could be even misleading, as variable `observer` is already defined)

src/bundle/Resources/public/js/scripts/helpers/tooltips.helper.js
```diff
@@ -1,5 +1,18 @@
-(function(global, doc, eZ, $) {
+(function (global, doc, eZ, $) {
     const TOOLTIPS_SELECTOR = '[title]';
+    const config = {
```
GrabowskiM, line 3: Change to `observerConfig` , so it would be more specific

PR #1396, opened 2020-06-04
Get getNotificationsStatus interval from config

adriendupuis:
| Question      | Answer
| ------------- | ---
| Tickets       | N/A
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Get `getNotificationsStatus` (`notifications/count`) interval from YAML config parameters.

One could want (or need) to increase this interval to reduce server activity.

Doc could be needed to explain that this is set in milliseconds but I couldn't find any doc for other parameters like `.timeout`.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
+++ b/src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
@@ -179,9 +179,11 @@
 
     const notificationsTable = modal.querySelector(SELECTOR_TABLE);
     currentPageLink = notificationsTable.dataset.notifications;
+    let interval = Number.parseInt(notificationsTable.dataset.notificationsCountInterval);
+    if (!interval) { interval = INTERVAL; }
 
     modal.querySelectorAll(SELECTOR_MODAL_RESULTS).forEach((link) => link.addEventListener('click', handleModalResultsClick, false));
 
     getNotificationsStatus();
-    global.setInterval(getNotificationsStatus, INTERVAL);
+    global.setInterval(getNotificationsStatus, interval);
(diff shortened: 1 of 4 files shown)


src/bundle/Resources/public/js/scripts/admin.notifications.modal.js
```diff
@@ -179,9 +179,11 @@
 
     const notificationsTable = modal.querySelector(SELECTOR_TABLE);
     currentPageLink = notificationsTable.dataset.notifications;
+    let interval = Number.parseInt(notificationsTable.dataset.notificationsCountInterval);
+    if (!interval) { interval = INTERVAL; }
```
adriendupuis, line 183: Fallback on hardcoded `INTERVAL` in case of `NaN` or `0`.

Steveb-p, line 183:
```suggestion
    const interval = Number.parseInt(notificationsTable.dataset.notificationsCountInterval) || INTERVAL;
```

dew326: I like this, just one small thing
`const interval = parseInt(notificationsTable.dataset.notificationsCountInterval, 10) || INTERVAL;`

adriendupuis: Thank you @dew326 for suggesting to add the radix/base.

dew326 (approved):

PR #1412, opened 2020-06-30
EZEE-3176: Make disabled fields are non-interactable

Nattfarinn:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZEE-3176
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/fieldType/edit/_base-field.scss
+++ b/src/bundle/Resources/public/scss/fieldType/edit/_base-field.scss
@@ -44,6 +44,11 @@
     }
 }
 
+.ez-field-edit--disabled {
+    opacity: 0.4;
+    pointer-events: none;
+}
+
 .ez-content-edit {
     .ez-field-edit {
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/public/scss/fieldType/edit/_base-field.scss
```diff
@@ -44,6 +44,11 @@
     }
 }
 
+.ez-field-edit--disabled {
```
dew326, line 47: This should share styling with `--nontranslatable` so above where there are styling for nontranslatable you should add just:
```scss
    &--nontranslatable,
    &--disabled {
```

Nattfarinn: Done.

PR #1415, opened 2020-06-30
EZP-31692: Hide tooltip after show/hide details of content

lucasOsti:
jira ref: https://jira.ez.no/browse/EZP-31692


dew326:
Is this valid for 3.0?

lucasOsti:
> Is this valid for 3.0?

in 3.0 we don't have a title for button "question mark"

dew326:
So the question is should we have a tooltip there?

lserwatka:
You can merge it up.

PR #1419, opened 2020-07-02
EZEE-3195: Fixed error message position

lucasOsti:
Jira ref: https://jira.ez.no/browse/EZEE-3195

### DIFF
--- a/src/bundle/Resources/public/scss/_warning-icon.scss
+++ b/src/bundle/Resources/public/scss/_warning-icon.scss
@@ -13,7 +13,7 @@
             justify-content: center;
             display: none;
 
-            .ez-icon {
+            .ez-icon.ez-icon--warning {
                 height: calculateRem(12px);
                 width: calculateRem(12px);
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/public/scss/_warning-icon.scss
```diff
@@ -13,7 +13,7 @@
             justify-content: center;
             display: none;
 
-            .ez-icon {
+            .ez-icon.ez-icon--warning {
```
GrabowskiM, line 16: Are you sure you haven't messed up some other place with this change?

lucasOsti: > Are you sure you haven't messed up some other place with this change?

there was a more general selector before

dew326: It will lose styling for elements that don't have class `ez-icon--warning`.
Why do you need more specific selector?

PR #1420, opened 2020-07-02
EZP-31697: Added remove tooltip when removed node is diffrenty than .…

lucasOsti:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31697
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

Improved tooltip removal. Previously the tooltip was removed if any node was deleted and now I delete it when I delete a node other than .ez-tooltip

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/helpers/tooltips.helper.js
+++ b/src/bundle/Resources/public/js/scripts/helpers/tooltips.helper.js
@@ -1,17 +1,27 @@
 (function (global, doc, eZ, $) {
+    let lastInsertTooltipTarget = null;
     const TOOLTIPS_SELECTOR = '[title]';
     const observerConfig = {
         childList: true,
         subtree: true,
     };
     const observer = new MutationObserver((mutationsList) => {
-        mutationsList.forEach((mutation) => {
-            const showedTooltipNode = doc.querySelector('.ez-tooltip.show');
+        if (lastInsertTooltipTarget) {
+            mutationsList.forEach((mutation) => {
+                const { removedNodes } = mutation;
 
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/scripts/helpers/tooltips.helper.js
```diff
@@ -39,6 +49,10 @@
                                     <div class="tooltip-inner ez-tooltip__inner"></div>
                                </div>`,
                 });
+
+                $(tooltipNode).on('inserted.bs.tooltip', (event) => {
```
dew326, line 53: If this is after the tooltip is set it will be undefined on first time the tooltip is added?

src/bundle/Resources/public/js/scripts/helpers/tooltips.helper.js
```diff
@@ -1,17 +1,27 @@
...
+            mutationsList.forEach((mutation) => {
+                const { removedNodes } = mutation;
 
-            if (mutation.removedNodes.length && showedTooltipNode) {
-                showedTooltipNode.remove();
-            }
-        });
+                if (removedNodes.length) {
+                    removedNodes.forEach((removedNode) => {
+                        if (removedNode.classList && !removedNode.classList.contains('ez-tooltip')) {
```
GrabowskiM, line 15: is this `removedNode.classList` needed? classList is DOM property so it should always exist

PR #1421, opened 2020-07-03
EZP-31719: Verbatim HTML in end of life information on Dashboard

glye:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31719
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | TBD
| Doc needed?   | no

The alert box for end-of-life shows verbatim (escaped) HTML. This adds `raw` operators and indents in the same way as similar texts in the same template.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/themes/admin/ui/dashboard/block/ez.html.twig
+++ b/src/bundle/Resources/views/themes/admin/ui/dashboard/block/ez.html.twig
@@ -14,8 +14,11 @@
     {% if not ez.release %}
         {% set severity = 1 %}
         <div class="alert alert-warning mb-0 mt-3" role="alert">
-            {{ 'dashboard.ez_version.release_not_determined'|trans|desc("The system could not find your <code>composer.lock</code> file. It's needed to determine information about
-            your eZ install, and recommended to be kept on project development to make sure same package versions are used across all environments.")}}
+            {{ 'dashboard.ez_version.release_not_determined'|trans
+            |desc("The system could not find your <code>composer.lock</code> file. It's needed to determine information about
+                your eZ install, and recommended to be kept on project development to make sure same package versions are
+                used across all environments.")
+            |raw }}
         </div>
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/views/themes/admin/ui/dashboard/block/ez.html.twig
```diff
@@ -14,8 +14,11 @@
     {% if not ez.release %}
         {% set severity = 1 %}
         <div class="alert alert-warning mb-0 mt-3" role="alert">
-            {{ 'dashboard.ez_version.release_not_determined'|trans|desc("The system could not find your <code>composer.lock</code> file. It's needed to determine information about
-            your eZ install, and recommended to be kept on project development to make sure same package versions are used across all environments.")}}
+            {{ 'dashboard.ez_version.release_not_determined'|trans
+            |desc("The system could not find your <code>composer.lock</code> file. It's needed to determine information about
+                your eZ install, and recommended to be kept on project development to make sure same package versions are
```
DominikaK, line 19:
```suggestion
                your eZ installation, and recommended to be kept on project development to make sure the same package versions are
```

DominikaK, line 19: Or even better:

```suggestion
                your eZ installation. It is recommended to keep it during project development to make sure the same package versions are
```

glye: Gracias.

glye: I went with the other one.

micszo:
@lserwatka @adamwojs @webhdx (since Glye is on vacation) we need this change for `1.5` also

PR #1437, opened 2020-07-22
EZP-31767: Use version contributor name on dashboard

ITernovtsiy:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-31767](https://jira.ez.no/browse/EZP-31767)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Currently, on content view page, we have displayed name of user who last edited the content.
Screenshot from 2020-07-22 13-19-47
But creator name displayed as a contributor on the dashboard (In common content block)
Screenshot from 2020-07-22 13-20-36

I would expect to see name of user who last edited the content.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/Pagination/Mapper/AbstractPagerContentToDataMapper.php
+++ b/src/lib/Pagination/Mapper/AbstractPagerContentToDataMapper.php
@@ -9,16 +9,16 @@ declare(strict_types=1);
 namespace EzSystems\EzPlatformAdminUi\Pagination\Mapper;
 
 use eZ\Publish\API\Repository\ContentTypeService;
+use eZ\Publish\API\Repository\Exceptions\NotFoundException;
 use eZ\Publish\API\Repository\LanguageService;
 use eZ\Publish\API\Repository\UserService;
 use eZ\Publish\API\Repository\Values\Content\Content;
-use eZ\Publish\API\Repository\Values\Content\ContentInfo;
 use eZ\Publish\API\Repository\Values\Content\Language;
+use eZ\Publish\API\Repository\Values\Content\VersionInfo;
 use eZ\Publish\API\Repository\Values\User\User;
 use eZ\Publish\Core\Helper\TranslationHelper;
 use eZ\Publish\Core\MVC\Symfony\Locale\UserLanguagePreferenceProviderInterface;
 use EzSystems\EzPlatformAdminUi\Specification\ContentIsUser;
-use EzSystems\EzPlatformAdminUi\Specification\UserExists;
 
 abstract class AbstractPagerContentToDataMapper
 {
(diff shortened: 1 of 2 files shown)


src/lib/Pagination/Mapper/AbstractPagerContentToDataMapper.php
```diff
@@ -88,19 +88,17 @@ protected function isContentIsUser(Content $content): bool
     }
 
     /**
-     * @param \eZ\Publish\API\Repository\Values\Content\ContentInfo $contentInfo
+     * @param \eZ\Publish\API\Repository\Values\Content\VersionInfo $versionInfo
      *
      * @return \eZ\Publish\API\Repository\Values\User\User|null
      */
-    protected function getContributor(ContentInfo $contentInfo): ?User
+    protected function getVersionContributor(VersionInfo $versionInfo): ?User
```
ViniTou, line 95: Not sure if renaming protected function in abstract class is the safest way to do, it should be internal but maybe someone else have opinion on that too.

src/lib/Pagination/Mapper/AbstractPagerContentToDataMapper.php
```diff
@@ -88,19 +88,17 @@ protected function isContentIsUser(Content $content): bool
...
     {
-        $userExists = (new UserExists($this->userService))->isSatisfiedBy($contentInfo->ownerId);
-
-        if (false === $userExists) {
+        try {
```
ViniTou, line 97: Any reason why you did not reuse `Specification` just with different ID provided?

ITernovtsiy: @ViniTou This method called for each row, so this will save us a few `userService->loadUser()` calls, which will make the dashboard page a bit faster.
IMHO, `UserExists->isSatisfiedBy` is overcomplicating the simple thing, user service is doing this job quite well already. Probably, it is the reason why it's not used commonly through the code.

Let me know if these few extra calls are not a big deal and I'll change it to use `UserExists`.
Thanks.

src/lib/Pagination/Mapper/AbstractPagerContentToDataMapper.php
ITernovtsiy: May I create a separate `getVersionContributor` method, and keep `getContributor` as is? And, mark `getContributor` with `@deprecated`?

src/lib/Pagination/Mapper/AbstractPagerContentToDataMapper.php
ViniTou: True, I didnt notice the last removed call, fine with me then.

adamwojs (approved):

lserwatka (approved):

kaff (approved):

mikadamczyk (approved):

bogusez (approved):

adamwojs:
Thank you for contribution @ITernovtsiy!

PR #1448, opened 2020-08-12
EZP-31792: Redesign UDW

lucasOsti:
Jira ref: https://jira.ez.no/browse/EZP-31792

### DIFF
--- a/src/bundle/Resources/public/scss/_buttons.scss
+++ b/src/bundle/Resources/public/scss/_buttons.scss
@@ -178,8 +178,8 @@
             border-color: $ibexa-color-base-dark;
 
             &:hover {
-                background-color: $ez-color-base-medium-hover;
-                border-color: $ez-color-base-medium-hover;
+                background-color: $ibexa-color-base-dark;
+                border-color: $ibexa-color-base-dark;
             }
         }
 
@@ -196,8 +196,8 @@
             border-color: $ibexa-color-base-dark;
 
             &:hover {
-                background-color: $ez-color-base-medium-hover;
-                border-color: $ez-color-base-medium-hover;
+                background-color: $ibexa-color-base-dark;
+                border-color: $ibexa-color-base-dark;
             }
         }
     }
@@ -247,3 +247,7 @@
         fill: $ibexa-color-primary;
     }
 }
+
+button:focus {
+    outline: 0;
+  }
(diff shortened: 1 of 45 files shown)


src/bundle/Resources/public/scss/_buttons.scss
```diff
@@ -247,3 +247,7 @@
         fill: $ibexa-color-primary;
     }
 }
+
+button:focus {
+    outline: 0;
+  }
```
dew326, line 253: indent

src/bundle/Resources/public/scss/ui/modules/universal-discovery/_content.create.button.scss
```diff
@@ -1,10 +1,14 @@
...
         padding: calculateRem(4px);
-        width: calculateRem(34px);
-        height: calculateRem(34px);
-        display: flex;
+        color: $ibexa-color-primary;
+        text-decoration: none;
+    }
+
+    .ez-icon {
+        fill: $ibexa-color-primary;
```
dew326, line 12: I think the primary color deserves to have an icon modifier.

GrabowskiM:
Fix PR name

src/bundle/Resources/views/themes/admin/content/edit/edit.html.twig
```diff
@@ -6,79 +6,28 @@
...
+{% block breadcrumbs %}
+    {% include '@ezdesign/ui/breadcrumbs.html.twig' with { items: [
+        { value: content_type.name }
+    ]} %}
+{% endblock %}
+
+{% block page_title %}
+    {% set content_name = content ? content.name : 'new_content_item'|trans({'%contentType%': content_type.name})|desc('New %contentType%') %}
+    {% include '@ezdesign/content/page_title_edit.html.twig' with { 
+        action_name: 'Editing'|trans|desc('Editing'),
```
GrabowskiM, line 18: translation id shouldn't be capitalized, I think

src/bundle/Resources/public/scss/ui/modules/universal-discovery/_content.create.button.scss
lucasOsti: We have modifier `ez-icon--primary`

PR #1449, opened 2020-08-14
As an Editor, I want to be able to copy the content type in admin interface

MarioBlazek:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-31801](https://jira.ez.no/browse/EZP-31801)
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

This PR implements ContentType copy feature in the admin.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/ContentTypeController.php
+++ b/src/bundle/Controller/ContentTypeController.php
@@ -402,15 +402,14 @@ class ContentTypeController extends Controller
             $this->contentTypeService->copyContentType($contentType);
 
             $this->notificationHandler->success(
-            /** @Desc("Content Type '%name%' copied.") */
+                /** @Desc("Content Type '%name%' copied.") */
                 'content_type.copy.success',
                 ['%name%' => $contentType->getName()],
                 'content_type'
             );
-
         } catch (UnauthorizedException $exception) {
             $this->notificationHandler->error(
-            /** @Desc("Content Type '%name%' cannot be copied.") */
+                /** @Desc("Content Type '%name%' cannot be copied.") */
                 'content_type.copy.error',
                 ['%name%' => $contentType->getName()],
                 'content_type'
--- a/src/bundle/Controller/ContentTypeController.php
+++ b/src/bundle/Controller/ContentTypeController.php
@@ -142,9 +142,18 @@ class ContentTypeController extends Controller
             new ContentTypesDeleteData($this->getContentTypesNumbers($types))
         );
 
+        $contentTypeCopyForms = [];
         foreach ($types as $type) {
             $deletableTypes[$type->id] = !$this->contentTypeService->isContentTypeUsed($type);
+
+            $url = $this->generateUrl('ezplatform.content_type.copy', [
+                'contentTypeId' => $type->id,
+                'contentTypeGroupId' => $group->id,
+            ]);
+            $copyData = new ContentTypeCopyData($type, $group);
+            $contentTypeCopyForms[$type->id] = $this->contentTypeFormFactory->contentTypeCopy($copyData, null, ['action' => $url])->createView();
         }
+
 
         return $this->render('@ezdesign/content_type/list.html.twig', [
             'content_type_group' => $group,
@@ -156,6 +165,7 @@ class ContentTypeController extends Controller
             'can_create' => $this->isGranted(new Attribute('class', 'create')),
             'can_update' => $this->isGranted(new Attribute('class', 'update')),
             'can_delete' => $this->isGranted(new Attribute('class', 'delete')),
+            'content_type_copy_forms' => $contentTypeCopyForms,
         ]);
     }
 
--- a/src/bundle/Controller/ContentTypeController.php
+++ b/src/bundle/Controller/ContentTypeController.php
@@ -386,6 +386,42 @@ class ContentTypeController extends Controller
     }
 
     /**
+     * @param \eZ\Publish\API\Repository\Values\ContentType\ContentTypeGroup $group
+     * @param \eZ\Publish\API\Repository\Values\ContentType\ContentType $contentType
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
+     */
+    public function copyAction(ContentTypeGroup $group, ContentType $contentType): Response
+    {
+        $this->denyAccessUnlessGranted(new Attribute('class', 'create'));
+
+        try {
+            $this->contentTypeService->copyContentType($contentType);
+
+            $this->notificationHandler->success(
+            /** @Desc("Content Type '%name%' copied.") */
+                'content_type.copy.success',
+                ['%name%' => $contentType->getName()],
+                'content_type'
+            );
+
+        } catch (\Exception $exception) {
+            $this->notificationHandler->error(
+            /** @Desc("Content Type '%name%' cannot be copied.") */
+                'content_type.copy.error',
+                ['%name%' => $contentType->getName()],
+                'content_type'
+            );
+        }
+
+        return $this->redirectToRoute('ezplatform.content_type_group.view', [
+            'contentTypeGroupId' => $group->id,
+        ]);
+    }
+
+    /**
      * @param \Symfony\Component\HttpFoundation\Request $request
      * @param \eZ\Publish\API\Repository\Values\ContentType\ContentTypeGroup $group
      * @param \eZ\Publish\API\Repository\Values\ContentType\ContentTypeDraft $contentTypeDraft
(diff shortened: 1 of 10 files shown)


MarioBlazek:
PR depends on this https://github.com/ezsystems/ezplatform-kernel/pull/97 as persistence cache doesn't get clear after `copy()` method, so the listing of content types remains the same.

SylvainGuittard:
Thanks @MarioBlazek for this contribution.
Could you please upload screenshot(s), so we can have a look at the overall design and user experience.

MarioBlazek:
Screenshot 2020-08-17 at 09 55 46

Screenshot 2020-08-17 at 09 56 02

@SylvainGuittard

SylvainGuittard:
Thanks @MarioBlazek

inakijv:
Hi @MarioBlazek thank you for this PR, it is a requested feature that will improve the experience of our users 👍 
One thing I am not sure is that if there's a need to create a new column for the "Copy Content Type" button within the table. Other tables in the interface that contain two or more buttons have them within the same column.

src/bundle/Resources/views/themes/admin/content_type/list.html.twig
```diff
@@ -55,6 +55,7 @@
                 <th class="ez-table__header-cell">{{ 'content_type.view.list.column.id'|trans|desc('ID') }}</th>
                 <th class="ez-table__header-cell">{{ 'content_type.view.list.column.modification_date'|trans|desc('Modification date') }}</th>
                 <th class="ez-table__header-cell"></th>
+                <th class="ez-table__header-cell"></th>
```
inakijv, line 58: Not sure if we need to add a new column

MarioBlazek:
@inakijv no problem, I can adjust that.

src/bundle/Resources/views/themes/admin/content_type/list.html.twig
```diff
@@ -83,6 +84,11 @@
                 <td class="ez-table__cell">{{ content_type.identifier }}</td>
                 <td class="ez-table__cell">{{ content_type.id }}</td>
                 <td class="ez-table__cell">{{ content_type.modificationDate|ez_full_datetime }}</td>
+                <td class="ez-table__cell ez-table__cell--has-action-btns text-right">
```
inakijv, line 87: Wouldn't be possible to have the new button within the same column? That way it would look more consitent with other tables in the app

src/bundle/Resources/views/themes/admin/content_type/macros.html.twig
```diff
@@ -13,4 +13,19 @@
...
+{% macro content_type_copy(content_type, content_type_group, class_name) %}
+    {% set copy_url = path('ezplatform.content_type.copy', {
+        'contentTypeGroupId': content_type_group.id,
+        'contentTypeId': content_type.id
+    }) %}
+
+    <a title="{{ 'content_type.action.copy'|trans|desc('Copy') }}"
+       href="{{ copy_url }}"
+       {% if class_name %}class="{{ class_name }}"{% endif %}>
+        <svg class="ez-icon ez-icon-copy">
```
inakijv, line 27: I would suggest to add also `ez-icon--secondary`, given that it is not a primary action. That way it would look more consistent with other tables in the app.

MarioBlazek:
Screenshot 2020-08-17 at 11 10 43

@inakijv @SylvainGuittard adjusted.

src/bundle/Resources/views/themes/admin/content_type/macros.html.twig
```diff
@@ -13,4 +13,19 @@
...
\ No newline at end of file
+{% endmacro %}
+
+{% macro content_type_copy(content_type, content_type_group, class_name) %}
+    {% set copy_url = path('ezplatform.content_type.copy', {
+        'contentTypeGroupId': content_type_group.id,
+        'contentTypeId': content_type.id
+    }) %}
+
+    <a title="{{ 'content_type.action.copy'|trans|desc('Copy') }}"
```
adamwojs, line 24: Translation for this \`content\_type.action.copy\` key seems to be missing.

src/bundle/Controller/ContentTypeController.php
```diff
@@ -385,6 +386,41 @@ public function editAction(Request $request, ContentTypeGroup $group, ContentTyp
...
 
+    /**
+     * @param \eZ\Publish\API\Repository\Values\ContentType\ContentTypeGroup $group
+     * @param \eZ\Publish\API\Repository\Values\ContentType\ContentType $contentType
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
+     */
+    public function copyAction(ContentTypeGroup $group, ContentType $contentType): Response
```
adamwojs, line 397: Hmmm I'm missing here CSRF protection....

MarioBlazek: @adamwojs do you have an example of handling the CSRF in the ezplatform-admin-ui?

MarioBlazek: I can create a simple form in controller to create and validate a CSRF.

SylvainGuittard:
@MarioBlazek Thanks for adjusting the design. 👍

src/bundle/Controller/ContentTypeController.php
ViniTou: Yup, that's what we do too. You could also drop `GET` action in routing then.

lserwatka:
@MarioBlazek we would like to include this feature in 3.2. Do you have time to finish it and address changes after code review? Rebase is needed as well.

MarioBlazek:
Sorry, the last few days I was in a rush. Will update this PR tomorrow, the latest.

lserwatka:
@MarioBlazek thank you, this is great contribution, we should use 3.2 window and release it.

lserwatka:
You don't need to fix design, we will take over, as we are in the middle of changing look & feel.

MarioBlazek:
@lserwatka @ViniTou PR updated.

dew326:
I removed the duplicated table row.

//cc @lserwatka The new design was already there.
Zrzut ekranu 2020-09-11 o 13 11 43

MarioBlazek:
I have just realized that form handling is not working properly. There is already one form for bulk delete out there, so multiple copy forms are added inside it. What do you guys suggest?

dew326:
But were those form are added? I can see there is only added the link to the copy route, and in fact it doesn't work because it redirect to the content types list.

MarioBlazek:
@dew326 I have update the PR with forms. Everything should be ok now.

src/bundle/Resources/views/themes/admin/content_type/list.html.twig
```diff
@@ -86,6 +86,10 @@
                 <td class="ez-table__cell">{{ content_type.id }}</td>
                 <td class="ez-table__cell">{{ content_type.modificationDate|ez_full_datetime }}</td>
                 <td class="ez-table__cell ez-table__cell--has-action-btns">
+                <td class="ez-table__cell ez-table__cell--has-action-btns text-right">
```
dew326, line 89:
```suggestion
```

kaff (approved):

lserwatka:
What is the status here? We are 1 week away from 3.2-rc. If we won't merge it this week, we will need it push it back till December.

GrabowskiM (approved):

src/bundle/Controller/ContentTypeController.php
```diff
@@ -385,6 +391,63 @@ public function editAction(Request $request, ContentTypeGroup $group, ContentTyp
...
+                try {
+                    $contentTypeService->copyContentType($contentType);
+
+                    $notificationHandler->success(
+                        /** @Desc("Content Type '%name%' copied.") */
+                        'content_type.copy.success',
+                        ['%name%' => $contentType->getName()],
+                        'content_type'
+                    );
+                } catch (UnauthorizedException $exception) {
```
ViniTou, line 427: When this exception can occur? Is `$this->denyAccessUnlessGranted(new Attribute('class', 'create'));` not enough?

src/bundle/Resources/config/routing.yaml
```diff
@@ -346,6 +346,16 @@ ezplatform.content_type.edit:
     requirements:
         contentTypeGroupId: \d+
 
+ezplatform.content_type.copy:
+    path: /contenttypegroup/{contentTypeGroupId}/contenttype/{contentTypeId}/copy
+    methods: ['GET', 'POST']
```
ViniTou, line 351: I would stick with POST only.

adamwojs (approved):
+1 besides @ViniTou comments :wink:

Nattfarinn (changes requested):
Hmm, user will end up with two Content Types with the very same name. Even if we prefix the name with "Copy" or something, the name will most likely be up to immediate change to something more meaningful. Wouldn't it be better to make it consistent with what we already have in ie. role Copy?

This way would avoid making JS changes to form action and other workarounds for "Form inside Form" issue as well.

Ping: @NataliaBecla @SylvainGuittard 

Role copy as reference:
Screenshot from 2020-10-07 16-13-20

lucasOsti (approved):

lserwatka:
@Nattfarinn we will address it after RC.

PR #1454, opened 2020-08-20
EZP-31797: New icons

lucasOsti:
jira ref: https://jira.ez.no/browse/EZP-31797


src/bundle/Resources/public/scss/fieldType/edit/_ezimageasset.scss
```diff
@@ -93,7 +94,8 @@
         .ez-icon {
             width: calculateRem(32px);
             height: calculateRem(32px);
-            fill: $ibexa-black;
+            stroke: $ibexa-black;
```
GrabowskiM, line 97: doubled property

src/bundle/Resources/public/scss/ui/modules/common/_tooltip.popup.scss
```diff
@@ -24,7 +24,8 @@
         cursor: pointer;
 
         .ez-icon {
-            fill: $ibexa-black;
+            stroke: $ibexa-black;
```
GrabowskiM, line 27: doubled property

GrabowskiM (changes requested):

PR #1456, opened 2020-08-24
EZEE-3239: Added additional event to show asset preview

ViniTou:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZEE-3239
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Unless there is different way to do this, I need those here:
https://github.com/ezsystems/ezplatform-connector-dam/pull/4

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/base/base-preview-field.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/base/base-preview-field.js
@@ -20,6 +20,7 @@
             const dataMaxSize = +this.inputField.dataset.maxFileSize;
 
             this.maxFileSize = parseInt(dataMaxSize, 10);
+            this.showPreviewEventName = 'ez-base:show-preview';
         }
 
         /**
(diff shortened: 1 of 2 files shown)


lucasOsti (approved):

src/bundle/Resources/public/js/scripts/fieldType/base/base-preview-field.js
```diff
@@ -270,6 +270,8 @@
             window.addEventListener('drop', this.preventDefaultAction, false);
             window.addEventListener('dragover', this.preventDefaultAction, false);
 
+            this.fieldContainer.addEventListener('ez-show-preview', () => this.showPreview());
```
dew326, line 273: I think `ez-show-preview` is too generic, I would make it more like: `ez-field-type-show-preview` or even give a possibility to fire custom event per "final" class not generic one for all use cases.

dew326 (changes requested):

dew326 (approved):

PR #1460, opened 2020-08-25
EZP-31812: Redesign admin section

lucasOsti:
jira ref: https://jira.ez.no/browse/EZP-31812?

### DIFF
--- a/src/bundle/Resources/views/themes/admin/section/list.html.twig
+++ b/src/bundle/Resources/views/themes/admin/section/list.html.twig
@@ -24,7 +24,7 @@
 {% block content %}
     <section class="container ez-container">
         <div class="ez-table-header">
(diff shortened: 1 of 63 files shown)


src/bundle/Resources/views/themes/admin/section/list.html.twig
```diff
@@ -24,7 +24,7 @@
 {% block content %}
     <section class="container ez-container">
         <div class="ez-table-header">
-            <div class="ez-table-header__headline">{{ 'section.list.title'|trans|desc('Sections') }}</div>
```
dew326, line 27: No header?

GrabowskiM (approved):

dew326 (approved):

PR #1463, opened 2020-08-28
EZP-31813: Redesign calendar

lucasOsti:
jira ref: https://jira.ez.no/browse/EZP-31813


dew326 (changes requested):
Why calendar styles are in admin-ui?

dew326 (approved):

PR #1476, opened 2020-09-10
EZP-31784: Draft autosave feature for Content Edit

Nattfarinn:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31784
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes. no. maybe?
| Tests pass?   | yes
| Doc needed?   | yes

Requires:
 - https://github.com/ezsystems/date-based-publisher/pull/192
 - https://github.com/ezsystems/ezplatform-content-forms/pull/33
 - https://github.com/ezsystems/ezplatform-page-builder/pull/653

Also needed:
 - https://github.com/ezsystems/ezplatform-form-builder/pull/248

#### Checklist:
- [x] Ready for Code Review


src/bundle/Resources/public/js/scripts/admin.content.edit.js
```diff
@@ -99,6 +99,50 @@
         }
     };
 
+    const isAutosaveEnabled = () => {
+        const autosaveSubmitBtn = form.querySelector('[name="ezplatform_content_forms_content_edit[autosave]"]');
+
+        return eZ.adminUiConfig.autosave.enabled && autosaveSubmitBtn;
```
GrabowskiM, line 105: Could change order an little, as when `eZ.adminUiConfig.autosave.enabled` is false it's unnecessary to get data from DOM

src/bundle/Resources/views/themes/admin/content/page_title_edit.html.twig
```diff
@@ -14,6 +14,23 @@
     <div class="ez-content-edit-page-title {{ class is defined ? class }}">
         <div class="ez-content-edit-page-title__action">{{ action_name }}</div>
         <div class="ez-content-edit-page-title__title">{{ title }}</div>
+        <div class="ez-content-edit-page-title__content-info">
+            <span
+                class="ez-content-edit-page-title__autosave-wrapper ez-content-edit-page-title__autosave-wrapper--not-saved"
+                title="{{ 'content_edit.autosave.turn_of.message'|trans()|desc('You can turn autosave off in your user settings') }}">
+                <span class="ez-content-edit-page-title__autosave-result">
+                    <svg class="ez-icon ez-icon ez-icon--small-medium ez-icon--autosave">
+                        <use xlink:href="/bundles/ezplatformadminui/img/ez-icons.svg#autosave"></use>
```
GrabowskiM, line 23: use `asset()` func

GrabowskiM (changes requested):

kaff (approved):

src/lib/Form/Processor/Content/AutosaveProcessor.php
```diff
@@ -0,0 +1,50 @@
...
+    public function __construct(
+        ContentFormProcessor $innerContentFormProcessor
+    ) {
+        $this->innerContentFormProcessor = $innerContentFormProcessor;
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            AutosaveEvents::CONTENT_AUTOSAVE => ['processAutosave', 10],
```
adamwojs, line 32: It seems that `AutosaveEvents::CONTENT_AUTOSAVE`is never dispatched so whole  `AutosaveProcessor` is a dead code.

src/lib/UI/Config/Provider/Autosave.php
```diff
@@ -0,0 +1,39 @@
...
+
+    public function __construct(
+        UserSettingService $userSettingService
+    ) {
+        $this->userSettingService = $userSettingService;
+    }
+
+    /**
+     * @throws \Exception
+     */
```
adamwojs, line 28: IMHO this docblock is redundant.

adamwojs (changes requested):

src/lib/Form/Processor/Content/AutosaveProcessor.php
webhdx: I can confirm that. I think it should be triggered because it provides a lot of value if you need to intercept autosaving and run your custom code.

src/lib/UserSetting/AutosaveEnabled.php
```diff
@@ -0,0 +1,126 @@
...
+
+    public function getDescription(): string
+    {
+        return $this->getTranslatedDescription();
+    }
+
+    public function getDisplayValue(string $storageValue): string
+    {
+        switch ($storageValue) {
+            case self::ENABLED_OPTION:
```
webhdx, line 46: The class/user setting name is confusing and does not go well with choice options. I don't think it's optimal to actually store "enabled/disabled" states. Simple `bool` value (more likely 0/1 since we store strings) would be sufficient and more appropriate. It would represent simple on/off checkbox more precisely.

`autosave_enabled: enabled/disabled` is counter intuitive, should be `true/false`

webhdx (changes requested):

GrabowskiM (approved):

src/lib/UserSetting/AutosaveEnabled.php
Nattfarinn: I have renamed option to `autosave`.

src/lib/Form/Processor/Content/AutosaveProcessor.php
Nattfarinn: Dispatched by Content Forms ActionDispatcher: https://github.com/ezsystems/ezplatform-content-forms/blob/master/src/lib/Form/ActionDispatcher/AbstractActionDispatcher.php#L50

webhdx (approved):

ciastektk (approved):

dew326 (approved):

ViniTou (approved):

tomaszszopinski (commented):
Fixed in bfc9691 : 
1. Not needed information for User regarding autosave (this CT doesnt use autosave)
2. Wrong message when autosave is off 
3. Creating draft even if autosave is turned off

tomaszszopinski (approved):
QA approved on IbexaDXP 3.2.0-dev with diff.

PR #1482, opened 2020-09-14
[Behat] Added reloading of the page when Content Type is not present in picker

mnocon:
| Question      | Answer
| ------------- | ---
| Tickets       | https://jira.ez.no/browse/EZP-31856
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

See JIRA ticket for failing builds.

This is our most common issue in tests - the Content Type created by the API is not visible in Content Type picker. This is a highly random issue, because all of our Content Types in tests are created the same way. Most of the time everything is correct, but sometimes some Content Types are not loaded by the Content Type picker.

This PR does not solve the issue, but may help us get more data about it so we can make further improvements.

Since this looks like a cache issue and there is very little data about this PR does to things:
1) Adds more debugging info: displays the available Content Types in the picker so we can have a bit more data about the state of the Page when the issue happens
2) Tries if reloading the Page fixes the issue

Even if reloading helps we still fail the tests to get a notification about it on Slack and be able to look at the build.

Example debug output will look like this: (simulated, based on https://travis-ci.org/github/ezsystems/ezplatform-admin-ui/jobs/726981397 - if I could reproduce the issue locally then it would be a different story 😉 )
```
      | FolderParent          | root/FolderGrandParent              | NewContent1 | root/FolderGrandParent/FolderParent              |
        │ Content Types before reload:
        │ Article
        │ Folder
        │ User
        │ User group
        │ File
        │ Image
        │ 
        │ 
        │ Content Types after reload:
        │ Article
        │ DedicatedFolder
        │ Folder
        │ User
        │ User group
        │ File
        │ Image
        │ 
        │ 
        │ 
        Content Type: DedicatedFolder was not detected the first time
```

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/Behat/PageElement/ContentTypePicker.php
+++ b/src/lib/Behat/PageElement/ContentTypePicker.php
@@ -6,6 +6,7 @@
  */
 namespace EzSystems\EzPlatformAdminUi\Behat\PageElement;
 
+use Behat\Mink\Element\NodeElement;
 use EzSystems\EzPlatformAdminUi\Behat\Helper\UtilityContext;
 
 class ContentTypePicker extends Element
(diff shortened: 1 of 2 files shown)


micszo (approved):
This should help narrow down the issue.

bogusez (approved):

tomaszszopinski (approved):

PR #1568, opened 2020-10-07
EZP-32003: Fix cancel editing redirect for root location

SerheyDolgushev:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-32003](https://jira.ez.no/browse/EZP-32003)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

**Steps to reproduce:**
1. Start editing the root location in admin UI
2.  Click on the "cancel" button in the top left

**Expected result:**

User is redirected to the full view page

**Actual result:**

The following error is shown:
```
Return value of eZ\Publish\Core\Repository\Values\Content\Content::getContentType() must be an instance of eZ\Publish\API\Repository\Values\ContentType\ContentType, null returned
```

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/themes/admin/content/page_title_edit.html.twig
+++ b/src/bundle/Resources/views/themes/admin/content/page_title_edit.html.twig
@@ -4,7 +4,7 @@
     {% if without_close_button is not defined or without_close_button != true %}
         <a
             class="ez-content-edit-container__close"
-            href="{{ path('_ez_content_view', { 'contentId': parent_location.contentId, 'locationId': parent_location.id}) }}"
+            href="{{ path('_ez_content_view', { 'contentId': max(parent_location.contentId, 1), 'locationId': max(parent_location.id, 2)}) }}"
             title="{{ 'tooltip.exit_label'|trans({}, 'content')|desc('Exit') }}"
         >
             <svg class="ez-icon ez-icon--small-medium ez-icon--primary">
--- a/src/bundle/Resources/views/themes/admin/content/page_title_edit.html.twig
+++ b/src/bundle/Resources/views/themes/admin/content/page_title_edit.html.twig
@@ -2,9 +2,10 @@
 
 {% block close_button %}
     {% if without_close_button is not defined or without_close_button != true %}
+        {% set referrer_location = parent_location.id > 1 ? parent_location : location %}
         <a
             class="ez-content-edit-container__close"
-            href="{{ path('_ez_content_view', { 'contentId': max(parent_location.contentId, 1), 'locationId': max(parent_location.id, 2)}) }}"
+            href="{{ path('_ez_content_view', { 'contentId': referrer_location.contentId, 'locationId': referrer_location.id }) }}"
             title="{{ 'tooltip.exit_label'|trans({}, 'content')|desc('Exit') }}"
         >
             <svg class="ez-icon ez-icon--small-medium ez-icon--primary">
--- a/src/bundle/Resources/views/themes/admin/content/page_title_edit.html.twig
+++ b/src/bundle/Resources/views/themes/admin/content/page_title_edit.html.twig
@@ -2,7 +2,7 @@
 
 {% block close_button %}
     {% if without_close_button is not defined or without_close_button != true %}
-        {% set referrer_location = parent_location.id > 1 ? parent_location : location %}
+        {% set referrer_location = is_published ? location : parent_location %}
         <a
             class="ez-content-edit-container__close"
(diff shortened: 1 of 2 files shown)


lserwatka:
Does it affect only master?

micszo:
Could not reproduce on v3.1.3.

SerheyDolgushev:
> Does it affect only master?

Seems like yes, as I was unable to reproduce it on 3.1.3

src/bundle/Resources/views/themes/admin/content/page_title_edit.html.twig
```diff
@@ -4,7 +4,7 @@
     {% if without_close_button is not defined or without_close_button != true %}
         <a
             class="ez-content-edit-container__close"
-            href="{{ path('_ez_content_view', { 'contentId': parent_location.contentId, 'locationId': parent_location.id}) }}"
+            href="{{ path('_ez_content_view', { 'contentId': max(parent_location.contentId, 1), 'locationId': max(parent_location.id, 2)}) }}"
```
dew326, line 7: I think it should be done in a different way. In edit mode, it should get the location not the parent location.
Please see how it was in 3.1
https://github.com/ezsystems/ezplatform-admin-ui/blob/2.1/src/bundle/Resources/views/themes/admin/content/edit/edit.html.twig#L16

dew326 (changes requested):

src/bundle/Resources/views/themes/admin/content/page_title_edit.html.twig
SerheyDolgushev: Done in https://github.com/ezsystems/ezplatform-admin-ui/pull/1568/commits/7d64bddb70506ef426d8fc60286d85dd21d02736

adamwojs (approved):

src/bundle/Resources/views/themes/admin/content/page_title_edit.html.twig
```diff
@@ -2,9 +2,10 @@
 
 {% block close_button %}
     {% if without_close_button is not defined or without_close_button != true %}
+        {% set referrer_location = parent_location.id > 1 ? parent_location : location %}
```
dew326, line 5: This will redirect to the location only when the user edits something under root location, what we need is restore the behavior from 3.1:
When the user creates content under some location he is redirected to the parent location when he closes the create mode.
When the user edits content he is redirected to this location, not to the parent.

dew326 (changes requested):

Nattfarinn (changes requested):
This is regression from `3.1` version. Please restore original bahaviour. There is no need to compare location id, just redirect to contents location if is published or parent if not.

SerheyDolgushev:
@dew326 @Nattfarinn 
Sorry, but I`m a bit confused here. I thought in 3.2 close button URL was changed on purpose: instead of redirecting to the current location view page, the parent location was used. Not sure what the reason was for that, maybe some users feedback who expected to work it the same way as in legacy? And if it is the case we should exception for the root location as we can not show its parent location view page.

From your previous comments seems like you would expect the "close button" work the same way in 3.2 as it was working in 3.1. @SylvainGuittard @inakijv can you please confirm it?

SylvainGuittard:
@SerheyDolgushev sorry for the confusion. We did not want to change the behavior. What @dew326 explained in https://github.com/ezsystems/ezplatform-admin-ui/pull/1568#discussion_r501465628 is the correct behavior

>When the user creates content under some location he is redirected to the parent location when he closes the create mode.
>When the user edits content he is redirected to this location, not to the parent.

SerheyDolgushev:
@Nattfarinn @dew326 done. Please have a look.

dew326 (approved):

dew326:
Note for QA: 
This header is used in:
Content - create/edit/translate
Page Builder - create/edit/translate
User - create/edit
UDW - create on the fly/edit on the fly
site-factory - create/edit

mnocon:
Hi @SerheyDolgushev 

The Travis build was failing after this change: https://travis-ci.org/github/ezsystems/ezplatform-admin-ui/jobs/734277378
I've adjusted the Scenario in https://github.com/ezsystems/ezplatform-admin-ui/pull/1568/commits/b01cddd19641213bccd162a0f74aecd7e84b5c8d so it should be green now.

Nattfarinn (approved):

katarzynazawada (approved):

PR #1620, opened 2020-10-19
EZEE-3325: Cancel publication button has wrong color

GrabowskiM:
JIRA ref: https://jira.ez.no/browse/EZEE-3325

### DIFF
--- a/src/bundle/Resources/views/themes/admin/content_type/field_types.html.twig
+++ b/src/bundle/Resources/views/themes/admin/content_type/field_types.html.twig
@@ -233,7 +233,7 @@
             </button>
             <button
                 type="button"
-                class="ezselection-settings-option-remove ez-font-icon ez-button btn btn-danger ez-button-delete pure-button"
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/views/themes/admin/content_type/field_types.html.twig
```diff
@@ -233,7 +233,7 @@
             </button>
             <button
                 type="button"
-                class="ezselection-settings-option-remove ez-font-icon ez-button btn btn-danger ez-button-delete pure-button"
+                class="ezselection-settings-option-remove ez-font-icon ez-button btn btn-primary ez-button-delete pure-button"
```
dew326, line 236: I'm not sure if this should be our primary color. Does the remove option is the primary action?

GrabowskiM: All our modal "Delete" buttons, trash icons etc have primary color so answering your question, it seems so

lucasOsti (approved):

dew326 (approved):

PR #1709, opened 2021-02-18
EZP-32311: Dropdown bottom options become inaccessible inside modals

konradoboza:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-32311](https://issues.ibexa.co/browse/EZP-32311)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

JIRA excerpt:
>When a Create a new translation window is shown and more than 6 Languages are defined, some of them won't be visible in the "Select a language for the new translation" field.

The idea is to check whether any of the dropdown parents is a modal (not to break dropdowns outside popups) and reduce the dropdown's height to not exceed the modal's height. Thanks to that, a scroll will appear allowing the user to select bottom options that are currently being covered by the modal itself.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
+++ b/src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
@@ -131,21 +131,21 @@
             this.sourceInput.dispatchEvent(new CustomEvent(EVENT_VALUE_CHANGED));
         }
 
-        getItemsContainerHeight(isItemsContainerTop) {
+        getItemsContainerHeight(isItemsContainerAbove) {
             const DROPDOWN_MARGIN = 16;
-            const MODAL_SELECTOR = '.modal[aria-modal=true]';
+            const SELECTOR_MODAL = '.modal[aria-modal=true]';
             const documentElementHeight = document.documentElement.getBoundingClientRect().height;
-            const itemsContainerHeight = this.itemsContainer.getBoundingClientRect().top;
+            const itemsContainerTop = this.itemsContainer.getBoundingClientRect().top;
 
-            if (isItemsContainerTop) {
+            if (isItemsContainerAbove) {
                 return this.container.querySelector(SELECTOR_SELECTION_INFO).getBoundingClientRect().top - DROPDOWN_MARGIN;
             }
 
-            if (this.itemsContainer.closest(MODAL_SELECTOR)) {
-                return itemsContainerHeight - DROPDOWN_MARGIN;
+            if (this.itemsContainer.closest(SELECTOR_MODAL)) {
+                return itemsContainerTop - DROPDOWN_MARGIN;
             }
 
-            return documentElementHeight - itemsContainerHeight - DROPDOWN_MARGIN;
+            return documentElementHeight - itemsContainerTop - DROPDOWN_MARGIN;
         }
 
         onInputClick(event) {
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
```diff
@@ -133,14 +133,19 @@
 
         getItemsContainerHeight(isItemsContainerTop) {
             const DROPDOWN_MARGIN = 16;
+            const MODAL_SELECTOR = '.modal[aria-modal=true]';
+            const documentElementHeight = document.documentElement.getBoundingClientRect().height;
+            const itemsContainerHeight = this.itemsContainer.getBoundingClientRect().top;
```
dew326, line 138:
```suggestion
            const itemsContainerTop = this.itemsContainer.getBoundingClientRect().top;
```

dew326 (changes requested):

src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
```diff
@@ -133,14 +133,19 @@
 
         getItemsContainerHeight(isItemsContainerTop) {
             const DROPDOWN_MARGIN = 16;
+            const MODAL_SELECTOR = '.modal[aria-modal=true]';
```
dew326, line 136:
```suggestion
            const SELECTOR_MODAL = '.modal[aria-modal=true]';
```

src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
konradoboza: In general, I agree with this naming change. But in this very case: wouldn't this be a little misleading given, we pass `isItemsContainerTop` parameter already? As for me, it would be harder to figure out what the code does at first glance.

dew326: Naming `getBoundingClientRect().top` `height` is even worse in understanding what this variable stores. Top position is completely different than height.
Maybe `itemsContainerTopPosition`?
Or renaming `isItemsContainerTop` as seems like this is not very descriptive either?

konradoboza: WDYT @dew326?
https://github.com/ezsystems/ezplatform-admin-ui/pull/1709/commits/ef4eaa409ef77ed53a68cc69e9c4f1d26f2b5887

dew326 (approved):

src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
```diff
@@ -131,16 +131,21 @@
             this.sourceInput.dispatchEvent(new CustomEvent(EVENT_VALUE_CHANGED));
         }
 
-        getItemsContainerHeight(isItemsContainerTop) {
+        getItemsContainerHeight(isItemsContainerAbove) {
             const DROPDOWN_MARGIN = 16;
+            const SELECTOR_MODAL = '.modal[aria-modal=true]';
+            const documentElementHeight = document.documentElement.getBoundingClientRect().height;
```
lucasOsti, line 137: You could use `doc` insted `document`

lucasOsti (changes requested):

src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
konradoboza: It wasn't there in the first place, but why not. Changed in https://github.com/ezsystems/ezplatform-admin-ui/pull/1709/commits/afdbc260465a858e2604d4245bde0cb4c8b6eead.

lucasOsti (approved):

tomaszszopinski (approved):
QA approved on Ibexa DXP 3.2.4 with diff & patch.

PR #1725, opened 2021-03-17
EZP-32395: Sort out Locations tab info on visibility

barw4:
| Question      | Answer
| ------------- | ---
| Tickets       | https://issues.ibexa.co/browse/EZP-32395
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

`supplyIsLocationInvisible` method is not really necessary, we can easily check if the `Location` is hidden in the twig itself + it shows invalid data as it only accounts for a single `Location`.  A small performance increase is an upside as well.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/content/locationview.html.twig
+++ b/src/bundle/Resources/views/content/locationview.html.twig
@@ -63,6 +63,7 @@
                 </div>
                 <div class="panel panel-primary">
                     <div class="panel-body">
+                        {# 'is_location_visible': location.invisible - param deprecated since EZP-32395 #}
                         {{ ez_platform_tabs('location-view', {
                             'content': content,
                             'location': location,
@@ -73,6 +74,7 @@
                             'system_urls_pagination_params': system_urls_pagination_params,
                             'roles_pagination_params': roles_pagination_params,
                             'policies_pagination_params': policies_pagination_params,
+                            'is_location_visible': location.invisible,
                         }, '@ezdesign/parts/tab/locationview.html.twig') }}
 
                         {% if contentType.isContainer %}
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/views/content/locationview.html.twig
```diff
@@ -73,7 +73,6 @@
                             'system_urls_pagination_params': system_urls_pagination_params,
                             'roles_pagination_params': roles_pagination_params,
                             'policies_pagination_params': policies_pagination_params,
-                            'is_location_visible': is_location_visible,
```
adamwojs, line 76: Technically this is BC break, so let leave this variable (with deprecation node) and pass here location.invisible

src/bundle/Resources/views/content/tab/locations/tab.html.twig
```diff
@@ -51,7 +51,7 @@
                                     {% if not can_hide[location.id] %} disabled="disabled"{% endif %}
                                     value="{{ location.id }}"/>
                             </label>
-                            {% if location.contentInfo.isHidden or not is_location_visible %}
+                            {% if location.invisible or location.hidden %}
```
adamwojs, line 54: According to https://github.com/ezsystems/ezplatform-kernel/blob/94937f8db831eb4229686f6cfde92e052a658b11/eZ/Publish/API/Repository/Values/Content/Location.php#L23:
> Indicates that the Location is not visible, being either marked as hidden itself, or implicitly hidden by its Content or an ancestor Location

`location.invisible` should be enough :wink:

adamwojs (approved):

damianz5 (approved):

tomaszszopinski (approved):
QA approved on eZPlatform 2.5 with patch.

lserwatka:
Could you merge it up?

barw4:
Merged:
into `2.2`: https://github.com/ezsystems/ezplatform-admin-ui/commit/3b8389856a40ea094a7a912d2e53c827e2ff19b3
into `master`: https://github.com/ezsystems/ezplatform-admin-ui/commit/40d47bb8d4af8287491b543eac7113ea56342478

PR #1738, opened 2021-04-16
IBX-148: Skipped deleted Relation fieldtypes when generating Relations tab

barw4:
| Question      | Answer
| ------------- | ---
| Tickets       | https://issues.ibexa.co/browse/IBX-148
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

As the title states.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/UI/Dataset/RelationListDataset.php
+++ b/src/lib/UI/Dataset/RelationListDataset.php
@@ -54,6 +54,10 @@ final class RelationListDataset
             $this->contentService->loadRelations($versionInfo)
         );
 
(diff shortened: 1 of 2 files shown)


webhdx (changes requested):
The real fix should remove all relations when removing the fieldtype. Hiding the relations from UI will be confusing since you would still get them in other contexts but you won't be able to see where they are coming from.

barw4:
Closed in favor of https://github.com/ezsystems/ezpublish-kernel/pull/3099

PR #1745, opened 2021-04-27
EZP-32400: Fixed displaying anchors for tables

konradoboza:
| Question      | Answer
| ------------- | ---
| Tickets       | [EZP-32400](https://issues.ibexa.co/browse/EZP-32400)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

SVG icon representing anchor is placed outside table by the AlloyEditor, hence is not saved alongside with the content. Therefore the icon is not presented in the edit mode. 

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
@@ -1,5 +1,7 @@
 (function(global) {
     const eZ = (global.eZ = global.eZ || {});
+    const TABLE_TAG_NAME = 'table';
+    const SVG_TAG_NAME = 'svg';
     const HTML_NODE = 1;
     const TEXT_NODE = 3;
     const notInitializeElements = ['strong', 'em', 'u', 'sup', 'sub', 's'];
@@ -170,8 +172,14 @@
 
         clearAnchor(element) {
             const icon = element.querySelector('.ez-icon--anchor');
+            const elementPreviousSibling = element.previousSibling;
+            const isTableWithAnchor =
+                element.tagName.toLowerCase() === TABLE_TAG_NAME &&
+                elementPreviousSibling.tagName.toLowerCase() === SVG_TAG_NAME;
 
-            if (icon) {
+            if (isTableWithAnchor) {
+                elementPreviousSibling.remove();
+            } else if (icon) {
                 icon.remove();
             } else {
                 element.classList.remove('ez-has-anchor');
--- a/src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
@@ -175,6 +175,7 @@
             const elementPreviousSibling = element.previousSibling;
             const isTableWithAnchor =
                 element.tagName.toLowerCase() === TABLE_TAG_NAME &&
+                elementPreviousSibling &&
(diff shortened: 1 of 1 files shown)


dew326 (approved):

adamwojs (approved):

src/bundle/Resources/public/js/scripts/fieldType/base/base-rich-text.js
```diff
@@ -170,8 +172,14 @@
 
         clearAnchor(element) {
             const icon = element.querySelector('.ez-icon--anchor');
-
-            if (icon) {
+            const elementPreviousSibling = element.previousSibling;
+            const isTableWithAnchor =
+                element.tagName.toLowerCase() === TABLE_TAG_NAME &&
+                elementPreviousSibling.tagName.toLowerCase() === SVG_TAG_NAME;
```
Steveb-p, line 178: `element.previousSibling` can be `null`, which would cause the script to crash due to method call on an undefined.

konradoboza: `clearAnchor` is called within the following context:
```
documentFragment.querySelectorAll('.ez-has-anchor').forEach(this.clearAnchor);
```
If there are no elements having this class, there will be no method call, so no potential null for the `element` variable I suppose.

Steveb-p: I'm not confident that it's an assumption that we can make (i.e. that the context won't change or the HTML elements are not changed). I'd rather have an additional check that we don't get `null` by chance.

And it's actually fairly easy to do by using the [Optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining). It has the same compatibility as object destructuring, so we should be safe to use it?

```suggestion
            const elementPreviousSibling = element.previousSibling;
            const isTableWithAnchor =
                element.tagName.toLowerCase() === TABLE_TAG_NAME &&
                elementPreviousSibling?.tagName.toLowerCase() === SVG_TAG_NAME;
```

dew326: This assumption here is ok as we have the BC and the HTML cannot really change especially in the LTS version, so this is nice to have but not mandatory in that case.
One thing, I would not like to introduce optional chaining in the old LTS code.

konradoboza: We might introduce that in the PR targetting v3: https://github.com/ezsystems/ezplatform-richtext/pull/182. WDYT @dew326?

Steveb-p: You can also just check the object:

```suggestion
            const elementPreviousSibling = element.previousSibling;
            const isTableWithAnchor =
                element.tagName.toLowerCase() === TABLE_TAG_NAME &&
                elementPreviousSibling && 
                elementPreviousSibling.tagName.toLowerCase() === SVG_TAG_NAME;
```

dew326: @konradoboza yep, optional chaining sounds better for 3.x

Steveb-p (approved):

bogusez (approved):

PR #1746, opened 2021-04-28
Boosted performance of NodeFactory

mateuszbieniek:
The `load-subtree` endpoint can become very slow in big projects when there are multiple Content Tree nodes expanded.
With every new tree expanded, the load time grows significantly. 

Reasons behind the slowness:
1. Counting child number for every single visible node using `SearchService::findLocations` with `$searchQuery->performCount = true;`
2. Initialization of Content Proxy objects - when for example 1000 nodes are displayed in the tree, every single node is loaded separately from SPI.

This PR optimizes `NodeFactory` to improve the overall response times for `load-subtree` endpoint by limiting `countSubitems` action to only those Nodes, that ContentType is a container (reducing unnecessary `countSubitems` calls), by relying on `$location->getContentInfo();` instead of `$location->getContent` when possible (`ContentInfo` is already initialised, therefore no additional call to SPI is made) and loading all Content in bulk instead of one-by-one initialisation.

The mentioned changes result in 60-70% response time reduction for `load-subtree` in large projects (10s -> 3-4s).

The `countSubitems` method can still be further optimized introducing further improvements in performance by using Aggregation, but it will be subject to a separate PR.

### DIFF
--- a/src/lib/UI/Module/ContentTree/NodeFactory.php
+++ b/src/lib/UI/Module/ContentTree/NodeFactory.php
@@ -8,6 +8,7 @@ declare(strict_types=1);
 
 namespace EzSystems\EzPlatformAdminUi\UI\Module\ContentTree;
 
+use ArrayObject;
 use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\Exceptions\NotImplementedException;
 use eZ\Publish\API\Repository\SearchService;
@@ -45,12 +46,6 @@ final class NodeFactory
     /** @var \eZ\Publish\Core\MVC\ConfigResolverInterface */
     private $configResolver;
 
-    /** @var \eZ\Publish\API\Repository\Values\Content\ContentInfo[] */
-    private $contentInfoList;
-
-    /** @var \eZ\Publish\API\Repository\Values\Content\Content[] */
-    private $contentByIds = [];
-
     public function __construct(
         ContentService $contentService,
         SearchService $searchService,
@@ -76,12 +71,10 @@ final class NodeFactory
         ?string $sortClause = null,
         string $sortOrder = Query::SORT_ASC
     ): Node {
-        $node = $this->buildNode($location, $loadSubtreeRequestNode, $loadChildren, $depth, $sortClause, $sortOrder);
-        $contentList = $this->contentService->loadContentListByContentInfo($this->contentInfoList);
-        foreach ($contentList as $content) {
-            $this->contentByIds[$content->id] = $content;
-        }
-        $this->supplyTranslatedContentName($node);
+        $contentInfoList = new ArrayObject();
+        $node = $this->buildNode($location, $loadSubtreeRequestNode, $loadChildren, $depth, $sortClause, $sortOrder, $contentInfoList);
+        $contentById = $this->contentService->loadContentListByContentInfo((array) $contentInfoList);
+        $this->supplyTranslatedContentName($node, $contentById);
 
         return $node;
     }
@@ -231,6 +224,8 @@ final class NodeFactory
     }
 
     /**
+     * @param \eZ\Publish\API\Repository\Values\Content\ContentInfo[] $contentInfoList
+     *
      * @return \EzSystems\EzPlatformAdminUi\REST\Value\ContentTree\Node
      *
      * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
@@ -243,12 +238,13 @@ final class NodeFactory
         bool $loadChildren = false,
         int $depth = 0,
         ?string $sortClause = null,
-        string $sortOrder = Query::SORT_ASC
+        string $sortOrder = Query::SORT_ASC,
+        ArrayObject $contentInfoList = null
     ): Node {
         $contentInfo = $location->getContentInfo();
         $contentId = $location->contentId;
         if (!isset($this->contentInfoList[$contentId])) {
-            $this->contentInfoList[$contentId] = $contentInfo;
+            $contentInfoList[$contentId] = $contentInfo;
         }
 
         // Top Level Location (id = 1) does not have a Content Type
@@ -276,7 +272,10 @@ final class NodeFactory
                     $childLocation,
                     $childLoadSubtreeRequestNode,
                     null !== $childLoadSubtreeRequestNode,
-                    $depth + 1
+                    $depth + 1,
+                    null,
+                    Query::SORT_ASC,
+                    $contentInfoList
                 );
             }
         } else {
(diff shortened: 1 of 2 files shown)


src/lib/UI/Module/ContentTree/NodeFactory.php
```diff
@@ -255,4 +222,92 @@ private function getSortClauses(
...
+
+    /**
+     * @param \eZ\Publish\API\Repository\Values\Content\ContentInfo[] $contentInfoList
+     *
+     * @return \EzSystems\EzPlatformAdminUi\REST\Value\ContentTree\Node
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
+     */
```
adamwojs, line 234:
```suggestion
    /**
     * @param \eZ\Publish\API\Repository\Values\Content\ContentInfo[] $contentInfoList
     *
     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
     */
```

src/lib/UI/Module/ContentTree/NodeFactory.php
```diff
@@ -255,4 +222,92 @@ private function getSortClauses(
...
+            $limit,
+            $totalChildrenCount,
+            $children
+        );
+    }
+
+    /**
+     * @param \EzSystems\EzPlatformAdminUi\REST\Value\ContentTree\Node $node
+     * @param \eZ\Publish\API\Repository\Values\Content\Content[] $contentById
+     */
```
adamwojs, line 305:
```suggestion
    /**
     * @param \eZ\Publish\API\Repository\Values\Content\Content[] $contentById
     */
```

src/lib/UI/Module/ContentTree/NodeFactory.php
```diff
@@ -255,4 +222,92 @@ private function getSortClauses(
...
+            if ($contentType && $contentType->isContainer) {
+                $totalChildrenCount = $this->countSubitems($location);
+            }
+        }
+
+        return new Node(
+            $depth,
+            $location->id,
+            $location->contentId,
+            '',
```
adamwojs, line 292:
```suggestion
            '', // node name will be provided later by `supplyTranslatedContentName` method
```

adamwojs (approved):

src/lib/UI/Module/ContentTree/NodeFactory.php
```diff
@@ -255,4 +222,92 @@ private function getSortClauses(
...
+        bool $loadChildren = false,
+        int $depth = 0,
+        ?string $sortClause = null,
+        string $sortOrder = Query::SORT_ASC,
+        ArrayObject $contentInfoList = null
+    ): Node {
+        $contentInfo = $location->getContentInfo();
+        $contentId = $location->contentId;
+        if (!isset($this->contentInfoList[$contentId])) {
+            $contentInfoList[$contentId] = $contentInfo;
```
Steveb-p, line 247: `$contentInfoList` might not be initialized at this point.

You probably want to add...
```php
$contentInfoList = $contentInfoList ?? new ArrayObject();
```
...at the beginning of this method.

Nattfarinn (approved):

src/lib/UI/Module/ContentTree/NodeFactory.php
```diff
@@ -255,4 +222,89 @@ private function getSortClauses(
...
+            ? $contentInfo->getContentType()
+            : null;
+
+        $limit = $this->resolveLoadLimit($loadSubtreeRequestNode);
+        $offset = null !== $loadSubtreeRequestNode
+            ? $loadSubtreeRequestNode->offset
+            : 0;
+
+        $children = [];
+        if ($depth < $this->getSetting('tree_max_depth') && $loadChildren) {
```
Steveb-p, line 259: Checking against boolean value goes first, as it's a simpler and faster operation.

```suggestion
        if ($loadChildren && $depth < $this->getSetting('tree_max_depth')) {
```

It's a copy from original though I see, so you can leave as is if you prefer.

ciastektk (approved):

mateuszdebinski (approved):

damianz5 (approved):

ViniTou (approved):

mikadamczyk (approved):

webhdx (approved):

src/lib/UI/Module/ContentTree/NodeFactory.php
```diff
@@ -255,4 +222,89 @@ private function getSortClauses(
             return []; // rely on storage engine default sorting
         }
     }
+
+    /**
+     * @param \eZ\Publish\API\Repository\Values\Content\ContentInfo[] $contentInfoList
```
Steveb-p, line 227: Please provide an explanation here why the object reference is used instead of a normal array. At first glance it's not obvious that it **has** to be a reference because it's supposed to be secondary "return" that is used to load Content in bulk after the call.

Also possible rename the variable to better reflect what it's purpose is.

mateuszbieniek: Do you have any suggestions for the variable name? I thought that `$contentInfoList` is self-explanatory.

Steveb-p: These are the `ContentInfo`'s that have uninitialized `Content`, right?

`$uninitializedContentInfoList`?

You could also use `&array` instead of `ArrayObject` to denote the need to use an array reference here.

The thing is just `$contentInfoList` is not really correct because you only place `ContentInfo` objects there if their `Content` objects were not loaded if I'm correct, hence why I'd like to ask for that variable to be renamed.

Steveb-p: If so, the method declaration would be:
```php
private function buildNode(
        Location $location,
        array &$uninitializedContentInfoList,
        ?LoadSubtreeRequestNode $loadSubtreeRequestNode = null,
        bool $loadChildren = false,
        int $depth = 0,
        ?string $sortClause = null,
        string $sortOrder = Query::SORT_ASC
): Node {

Comment (mateuszbieniek) [reply src/lib/UI/Module/ContentTree/NodeFactory.php]: Ok, that's fine with me - thanks for explaining @Steveb-p :slightly_smiling_face:

File: src/lib/UI/Module/ContentTree/NodeFactory.php [hunk 7]
```diff
@@ -8,6 +8,8 @@
 
 namespace EzSystems\EzPlatformAdminUi\UI\Module\ContentTree;
 
+use ArrayObject;
```
Steveb-p, line 11: This import is unnecessary after changes.

mnocon (approved):
Tested the behaviour of Content Tree for, including:
1) hidden/visible Content items (Content or Location)
2) Translated Content Items
3) Items with children, items without them

I didn't notice any functional change in Content Tree, for the data I've generated the speedup was not as big (~10%) but I guesss it depends on the data (my Content Items were 99% containers).

QA Approved

mateuszbieniek:
Merged up:
`2.3`: https://github.com/ezsystems/ezplatform-admin-ui/commit/40b9f0d62c9c0a32706ab70cbc638e06743d72a7
`master:` https://github.com/ezsystems/ezplatform-admin-ui/commit/6bbc4645c6df83af3819970a229bc3ce65702e90

PR #1790, opened 2021-06-30
IBX-673 Moving some jobs from Travis to GitHub actions in ezplatform-admin-ui

kacper-wieczorek-ibexa:
https://issues.ibexa.co/browse/IBX-673

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,49 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+      - '[0-9]+.[0-9]+'
+  pull_request: ~
+
+env:
+  COMPOSER_ROOT_VERSION: 1.5.x-dev
+
+jobs:
+  tests:
+    name: Tests
+    runs-on: "ubuntu-20.04"
+    timeout-minutes: 10
+
+    strategy:
+      fail-fast: false
+      matrix:
+        php:
+          - '7.3'
+          - '7.4'
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup PHP Action
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          coverage: none
+          extensions: pdo_sqlite, gd
+          tools: cs2pr
+
+      - uses: "ramsey/composer-install@v1"
+        with:
+          dependency-versions: "highest"
+          composer-options: "--prefer-dist --no-progress"
+
+      - name: Setup problem matchers for PHPUnit
+        run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
+
+      - name: Run code style check
+        run: composer run-script check-cs -- --format=checkstyle | cs2pr
+
+      - name: Run test suite
+        run: composer run-script --timeout=600 test
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,13 +21,6 @@ branches:
 
 matrix:
   include:
-    - name: "Code Style Check"
-      env: CHECK_CS=1
-    - name: "Unit tests"
-      env: PHPUNIT_CONFIG='phpunit.xml'
-    - name: "Unit tests (PHP 7.4)"
-      env: PHPUNIT_CONFIG='phpunit.xml'
-      php: 7.4
     - name: "Admin-UI tests on Clean P"
       env:
         - COMPOSE_FILE="doc/docker/base-dev.yml:doc/docker/selenium.yml"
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+on:
+  push:
+    branches:
+      - master
+      - '[0-9]+.[0-9]+'
+  pull_request: ~
+
+env:
+  COMPOSER_ROOT_VERSION: 1.5.x-dev
```
Steveb-p, line 11: This should not be necessary and `branch-alias` should be used instead:

https://getcomposer.org/doc/articles/troubleshooting.md#dependencies-on-the-root-package

(afaik GH actions do not use shallow clones, or at the very least not shallow enough)

.travis.yml
```diff
@@ -21,13 +21,13 @@ branches:
...
+    #    - name: "Code Style Check"
+    #      env: CHECK_CS=1
+    #    - name: "Unit tests"
+    #      env: PHPUNIT_CONFIG='phpunit.xml'
+    #    - name: "Unit tests (PHP 7.4)"
+    #      env: PHPUNIT_CONFIG='phpunit.xml'
+    #      php: 7.4
     - name: "Admin-UI tests on Clean P"
       env:
         - COMPOSE_FILE="doc/docker/base-dev.yml:doc/docker/selenium.yml"
```
Steveb-p, line 33: By the way, apparently `docker-compose` is available (built-in) in GH actions @mnocon. So if you ever would like to transfer behat / selenium tests there as well, there is an option.

mnocon: Yes, we plan to switch the selenium tests to Github Actions as well, but this will be a bigger task and I haven't investigated yet what's the best way to do it.

Good to know that docker-compose is available 👍

.github/workflows/ci.yaml
mnocon: The default fetch-depth is 1 (from https://github.com/actions/checkout), Composer was unable to resolve dependencies because of cyclic dependency between AdminUI and ez-support-tools and that's why it was added.

As far as I know Composer should create an implicit branch-alias for branch named like a version (1.5), if I understand correctly your suggestion we could try to add an explicit alias, something like: https://github.com/ezsystems/ezplatform-admin-ui/blob/1.5/composer.json#L46
```
        "branch-alias": {
            "dev-tmp_ci_branch": "1.5.x-dev",
            "dev-master": "1.5.x-dev",
            "dev-1.5": "1.5.x-dev"
        }
```

But I don't know if it works if the whole history is not fetched - definitely something we can check.

Steveb-p: Ok, didn't know that. If it's necessary for something, then :+1:

Although it makes it rather cumbersome to switch between versions, as you'd need to update both composer.json and CI declaration files. If any solution could be found that is contained within `composer.json` file, I think it should be prefered.

Steveb-p (approved):
:+1: despite `COMPOSER_ROOT_VERSION` usage.

As said in comment, if we can find something else, then great. If not... well, too bad.

.travis.yml
```diff
@@ -21,13 +21,13 @@ branches:
...
 matrix:
   include:
-    - name: "Code Style Check"
-      env: CHECK_CS=1
-    - name: "Unit tests"
-      env: PHPUNIT_CONFIG='phpunit.xml'
-    - name: "Unit tests (PHP 7.4)"
-      env: PHPUNIT_CONFIG='phpunit.xml'
-      php: 7.4
+    #    - name: "Code Style Check"
```
mnocon, line 24: Please remove commented code

composer.json
```diff
@@ -37,13 +37,18 @@
...
+        "test": "phpunit -c phpunit.xml"
     },
     "extra": {
         "_ezplatform_branch_for_behat_tests": "2.5",
         "branch-alias": {
             "dev-tmp_ci_branch": "1.5.x-dev",
             "dev-master": "1.5.x-dev"
         }
+    },
+    "config": {
```
mnocon, line 51: Why this change? IMHO we should not install with source by default.

.travis.yml
```diff
@@ -21,13 +21,7 @@ branches:
...
 matrix:
   include:
-    - name: "Code Style Check"
-      env: CHECK_CS=1
-    - name: "Unit tests"
-      env: PHPUNIT_CONFIG='phpunit.xml'
-    - name: "Unit tests (PHP 7.4)"
-      env: PHPUNIT_CONFIG='phpunit.xml'
-      php: 7.4
+
```
mnocon, line 24: Please remove empty line

.travis.yml
```diff
@@ -55,18 +49,15 @@ before_install:
   - echo "memory_limit=-1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
 
 install:
-  # Install packages if needed
-  - if [ "${CHECK_CS}" == "1" -o "${PHPUNIT_CONFIG}" != "" ]; then travis_retry composer install --prefer-dist --no-interaction --no-suggest ; fi
-  # Prepare whole environment if needed
-  - if [ "${BEHAT_OPTS}" != "" ]; then ./.travis/prepare_ezplatform.sh ${INSTALL_EZ_INSTALL_TYPE}; fi
+
```
mnocon, line 52: Please remove empty line

.travis.yml
```diff
@@ -55,18 +49,15 @@ before_install:
...
 
 before_script:
-   - if [ "${SETUP_BEHAT_OPTS}" != "" ] ; then cd "$HOME/build/ezplatform"; docker-compose exec --user www-data app sh -c "bin/ezbehat $SETUP_BEHAT_OPTS" ; fi
+  - if [ "${SETUP_BEHAT_OPTS}" != "" ] ; then cd "$HOME/build/ezplatform"; docker-compose exec --user www-data app sh -c "bin/ezbehat $SETUP_BEHAT_OPTS" ;fi
 
 script:
-  - if [ "${CHECK_CS}" == "1" ] ; then ./vendor/bin/php-cs-fixer fix -v --dry-run --show-progress=estimating ; fi
-  - if [ "${PHPUNIT_CONFIG}" != '' ]; then ./vendor/bin/phpunit -c "${PHPUNIT_CONFIG}"; fi
-  - if [ "${BEHAT_OPTS}" != "" ]; then cd "$HOME/build/ezplatform"; docker-compose exec --user www-data app sh -c "bin/ezbehat ${BEHAT_OPTS}" ; fi
+
```
mnocon, line 59: Please remove empty line

.travis.yml
```diff
@@ -77,4 +68,5 @@ after_failure:
   - docker ps -s
 
 after_script:
-  - if [ "${BEHAT_OPTS}" != "" ] ; then bin/ezreport ; fi
+
```
mnocon, line 71: Please remove empty line

composer.json
```diff
@@ -46,4 +48,5 @@
             "dev-master": "1.5.x-dev"
         }
     }
+
```
mnocon, line 51: Please remove empty line

mnocon (commented):
Looks good, just some nitpicks and we're good to go 😄

mnocon (approved):
Looks good, thanks!

micszo (approved):

PR #1830, opened 2021-07-27
IBX-375: Fixed setting 2nd menu level active

webhdx:
| Question      | Answer
| ------------- | ---
| Tickets       | https://issues.ibexa.co/browse/IBX-375
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

When working on IBX-375 we noticed there is a problem with 2nd level menu. Even though you set `extras.routes` on first level element and it matches, the second level menu won't be displayed unless there is a match on a child. In some cases you still want to display 2nd level menu, for example Page Builder submenu when there is no siteaccess match.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/parts/menu/top_menu_2nd_level.html.twig
+++ b/src/bundle/Resources/views/parts/menu/top_menu_2nd_level.html.twig
@@ -6,7 +6,7 @@
     <div class="tab-content mx-4">
         {% for child in currentItem.children %}
             <div role="tabpanel"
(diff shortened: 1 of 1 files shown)


ViniTou (approved):

adamwojs (approved):

micszo (approved):
Tested with https://github.com/ezsystems/ezplatform-page-builder/pull/797.

PR #1835, opened 2021-08-03
IBX-792: Implemented Field Types Palette

adamwojs:
| Question      | Answer
| ------------- | ---
| Tickets       | https://issues.ibexa.co/browse/IBX-792
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | ?

#### REST endpoints for content type editing

TBD.

#### `ibexa_field_group_name` filter

`ibexa_field_group_name` filter allows to render human readable name of given field group. Usage

```twig
{{ field.group | ibexa_field_group_name }}
```

#### TODO:

- [X] Create endpoint adding field defintions to content type
- [x] Create endpoint removing field definition from content type
- [x] Create endpoint to reorder field definitions
- [x] Remove inlined CSS / JS
- [x] Implement Twig filter for field groups rendering
- [ ] Clean up existing forms
- [x] Move to new namespaces

#### Checklist:
- [ ] Ready for Code Review


tomaszszopinski:
@adamwojs  there seems to be a problem with notifications regarding CT creation; when a new CT is created (with its own custom name) successfully, then notification appears but with default CT name.

Zrzut ekranu 2021-08-24 o 14 33 10

src/bundle/Resources/views/themes/admin/content_type/part/field_definition_form.html.twig
```diff
@@ -0,0 +1,56 @@
+{% trans_default_domain 'content_type' %}
+
+{% form_theme form '@ezdesign/ui/form_fields.html.twig' %}
+
+{% set value = form.vars.value %}
+{% set name = value.names[language_code] ?? value.names[content_type.mainLanguageCode] ?? ''  %}
```
GrabowskiM, line 6: this line seems strange, maybe it would be better to put in in two lines, instead of double shorthand?

src/bundle/Resources/views/themes/admin/content_type/part/field_definition_form.html.twig
```diff
@@ -0,0 +1,56 @@
...
+
+{%- embed "@ezdesign/ui/component/collapse.html.twig" with {
+    'body_id': form.vars.id ~ '_collapse',
+    'header_label': name,
+    'is_expand': false,
+    'is_draggable': true,
+    'container_class': 'ibexa-collapse--field-definition ' ~ 'ibexa-field-definition-' ~ form.identifier.vars.value,
```
GrabowskiM, line 13:
```suggestion
    'container_class': 'ibexa-collapse--field-definition ibexa-field-definition-' ~ form.identifier.vars.value,
```
Does it work by the way? No dots before classes?

src/bundle/Resources/views/themes/admin/content_type/part/field_definition_form.html.twig
```diff
@@ -0,0 +1,56 @@
...
+    'data_attribute': {
+        'data-is-new-field': "true",
```
GrabowskiM, line 15: singleqoute

src/bundle/Resources/views/themes/admin/ui/component/collapse.html.twig
```diff
@@ -0,0 +1,59 @@
+{% import '@ezdesign/ui/component/macros.html.twig' as macros %}
+
+<div
+    class="ibexa-collapse {{ is_hidden|default(false) ? 'ibexa-collapse--hidden' }} {{ container_class|default('') }}"
+    draggable="{{ is_draggable|default(false) ? 'true' : 'false' }}"
```
GrabowskiM, line 5: Isn't enough?
```suggestion
    draggable="{{ is_draggable|default(false) }}"
```

GrabowskiM (changes requested):

src/bundle/Resources/public/js/scripts/admin.contenttype.edit.js
```diff
@@ -0,0 +1,306 @@
...
+                container: dropdownContainer,
+            });
+
+            dropdown.init();
+        });
+    };
+    const generateRequest = (action, bodyData, languageCode) => {
+        const { actionName, method, contentType } = endpoints[action];
+        const { contentTypeGroupId, contentTypeId } = sectionsNode.dataset;
+        const baseURL = '/api/ezp/v2/contenttypegroup/_contentTypeGroupId_/contenttype/_contentTypeId_/_actionName_';
```
dew326, line 101:
```suggestion
        const baseURL = `/api/ezp/v2/contenttypegroup/${contentTypeGroupId}/contenttype/${contentTypeId}/${actionName}`;
```

dew326: Seems like those endpoints are exposed, it should be done by Routing

dew326 (changes requested):

src/bundle/Resources/public/js/scripts/admin.contenttype.edit.js
```diff
@@ -0,0 +1,306 @@
...
+            doc.body.dispatchEvent(
+                new CustomEvent('ibexa-drop-field-definition', {
+                    detail: { fieldSelector: `.ibexa-field-definition-${fieldNode.dataset.identifierValue}` },
+                })
+            );
+        }
+
+        const dropdowns = fieldNode.querySelectorAll('.ibexa-dropdown');
+
+        dropdowns.forEach((dropdownContainer) => {
```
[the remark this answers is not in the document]
lucasOsti: Yes it is, because after drop we create new node

tischsoic (changes requested):

src/bundle/Resources/public/js/scripts/admin.anchor.navigation.js
```diff
@@ -49,9 +49,12 @@
 
                 return position > start && position < end;
             });
-            const activeSectionId = activeSection.dataset.anchorSectionId;
 
-            showSection(activeSectionId);
+            if (activeSection) {
+                const activeSectionId = activeSection.dataset.anchorSectionId;
+
+                showSection(activeSectionId);
```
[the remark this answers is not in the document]
GrabowskiM: In such case I would leave it in previous form, destructing only one thing and changing its name is over the top imo. :)

src/bundle/Resources/views/themes/admin/ui/edit_header.html.twig
```diff
@@ -0,0 +1,52 @@
...
+                    <label class="ibexa-label ibexa-label--small ibexa-edit-header__action-name">{{ action_name|default('') }}
+                        {{ action_object|default('') }}
+                        {{ content_type_name|default('') }}
+                    </label>
+                </div>
+            </div>
+            <div class="col-6 ibexa-edit-header__column ibexa-context-menu-wrapper">
+                <div class="ibexa-edit-header__context-actions">
+                    {% if (description is defined and description|length) or content is defined %}
+                        <div class="ibexa-edit-header__tooltip" title="{% include '@ezdesign/ui/edit_header_tooltip.html.twig' %}">
```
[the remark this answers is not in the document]
lucasOsti: Solved: discussed privately

src/bundle/Resources/views/themes/admin/content_type/available_field_types.html.twig
```diff
@@ -0,0 +1,20 @@
...
+                </svg>
+                <svg class="ibexa-icon ibexa-icon--small ibexa-icon--field-icon">
+                    <use xlink:href="{{ ibexa_icon_path('dropdown') }}"></use>
+                </svg>
+                <div class="ibexa-label ibexa-label--medium">{{ item.name }}</div>
+            </li>
+        {% endfor %}
+    </ul>
+</div>
+
```
GrabowskiM, line 20: line too much

src/bundle/Resources/views/themes/admin/content_type/field_definitions_empty_group.html.twig
```diff
@@ -0,0 +1,9 @@
+<div class="ibexa-field-definitions-empty-group {% if field_defintions|length %}ibexa-field-definitions-empty-group--hidden{% endif %}">
+    <img
+        class="ibexa-field-definitions-empty-group__image"
+        src="{{ asset('bundles/ezplatformadminui/img/empty-field-definition-group.svg') }}"
+        alt="{{ 'content_type.view.edit.empty_group_content'|trans|desc('Empty group, please drag some field') }}"
+    >
+    <label class="ibexa-field-definitions-empty-group__label">{{ 'content_type.view.edit.empty_group_label'|trans|desc('Your section is empty') }}</label>
+    <h3 class="ibexa-field-definitions-empty-group__title">{{ 'content_type.view.edit.empty_group_content'|trans|desc('Drag and drop field type') }}</h2>
+</div>
```
GrabowskiM, line 9: new line

GrabowskiM (changes requested):

src/bundle/Resources/views/themes/admin/content/create/create.html.twig
```diff
@@ -1,4 +1,4 @@
-{% extends '@ezdesign/content/edit/base.html.twig' %}
+{% extends '@ezdesign/content/edit_base.html.twig'  %}
```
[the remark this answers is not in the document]
tischsoic: There is already 1 space between `'` and `%`.

src/bundle/Resources/views/themes/admin/ui/edit_header.html.twig
```diff
@@ -0,0 +1,52 @@
+<div class="ibexa-edit-header">
+    <div class="container">
+        <div class="row ibexa-edit-header__row ibexa-edit-header__row--top">
+            <div class="col-5 offset-1 ibexa-edit-header__column ibexa-edit-header__column--main">
+                <div class="ibexa-edit-header__action-name-container">
+                    {% if icon_name %}
+                        <svg class="ibexa-icon ibexa-icon--small">
+                            <use xlink:href="{{ ibexa_icon_path(icon_name|lower) }}"></use>
```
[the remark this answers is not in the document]
lucasOsti: yes we need, because sometimes the icon name is passed as a variable e.g. `icon_name: content_type.name, ` (content -> create)

src/bundle/Resources/public/js/scripts/admin.contenttype.edit.js
```diff
@@ -0,0 +1,346 @@
...
+            .then(eZ.helpers.request.getTextFromResponse)
+            .then(() => {
+                collapseNode.remove();
+                afterChangeGroup();
+            })
+            .catch(eZ.helpers.notification.showErrorNotification);
+    };
+
+    class FieldDefinitionDraggable extends eZ.core.Draggable {
+        constructor(config) {
```
GrabowskiM, line 252: Is this constructor needed?

src/bundle/Resources/public/js/scripts/admin.contenttype.edit.js
```diff
@@ -0,0 +1,346 @@
...
+
+    doc.querySelectorAll('.ibexa-collapse__extra-action-button--remove-field-definitions').forEach((removeFieldDefinitionsButton) => {
+        removeFieldDefinitionsButton.addEventListener('click', removeField, false);
+    });
+    doc.querySelectorAll('.ibexa-collapse__extra-action-button--remove-field-definitions-group').forEach(
+        (removeFieldDefinitionsGroupButton) => {
+            removeFieldDefinitionsGroupButton.addEventListener('click', removeFieldsGroup, false);
+        }
+    );
+
```
GrabowskiM, line 298: non-consisten new line

src/bundle/Resources/public/js/scripts/admin.contenttype.edit.js
```diff
@@ -0,0 +1,346 @@
...
+            (event) => {
+                currentDraggedItem = event.currentTarget;
+                sourceContainer = currentDraggedItem.parentNode;
+                currentDraggedItem.classList.add('ibexa-available-field-types__field--is-dragged-out');
+            },
+            false
+        );
+        availableField.addEventListener(
+            'dragend',
+            (event) => {
```
GrabowskiM, line 311: argument not needed

GrabowskiM (changes requested):

src/bundle/Resources/public/js/scripts/admin.contenttype.edit.js
```diff
@@ -0,0 +1,346 @@
...
+    const TIMEOUT_REMOVE_PLACEHOLDERS = 1500;
+    let targetContainer = null;
+    let sourceContainer = null;
+    let currentDraggedItem = null;
+    let draggedItemPosition = null;
+    const draggableGroups = [];
+    const token = doc.querySelector('meta[name="CSRF-Token"]').content;
+    const siteaccess = document.querySelector('meta[name="SiteAccess"]').content;
+    const sectionsNode = doc.querySelector('.ibexa-content-type-edit__sections');
+    const searchFiledInput = doc.querySelector('.ibexa-available-field-types__search');
```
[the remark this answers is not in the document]
lucasOsti: nope, `searchFieldInput` :)

src/bundle/Resources/public/js/scripts/admin.contenttype.edit.js
lucasOsti: solved in private conversation

GrabowskiM (approved):

src/bundle/Resources/public/js/scripts/admin.contenttype.edit.js
```diff
@@ -0,0 +1,341 @@
...
+                collapseNode.querySelectorAll('.ibexa-collapse--field-definition').forEach((fieldDefinition) => {
+                    fieldDefinition.remove();
+                });
+                afterChangeGroup();
+            })
+            .catch(eZ.helpers.notification.showErrorNotification);
+    };
+    const removeField = (event) => {
+        const collapseNode = event.currentTarget.closest('.ibexa-collapse');
+        const itemIdentifierToDelete = collapseNode.dataset.fieldDefinitionIdentifier;
```
tischsoic, line 234: nitpick:
```suggestion
        const itemToDeleteIdentifier = collapseNode.dataset.fieldDefinitionIdentifier;
```

src/bundle/Resources/public/scss/_custom.scss
```diff
@@ -420,13 +420,15 @@ $ibexa-text-font-size-small: calculateRem(12px);
...
 $ibexa-btn-focus-box-shadow: 0 0 0 calculateRem(3px) rgba($ibexa-color-primary, 0.2);
+$ibexa-btn-focus-box-shadow-info: 0 0 0 calculateRem(3px) rgba($ibexa-color-info, 0.2);
 $ibexa-btn-hover-box-shadow: 0 calculateRem(22px) calculateRem(24px) 0 rgba($ibexa-color-danger, 0.2);
 $ibexa-btn-hover-box-shadow-dark: 0 calculateRem(22px) calculateRem(24px) 0 rgba($ibexa-color-dark, 0.2);
 $ibexa-btn-hover-box-shadow-info: 0 calculateRem(10px) calculateRem(24px) 0 rgba($ibexa-color-info, 0.5);
 $ibexa-edit-header-box-shadow: calculateRem(4px) calculateRem(22px) calculateRem(47px) 0 rgba($ibexa-color-info, 0.15);
 $ibexa-edit-content-box-shadow: calculateRem(4px) calculateRem(35px) calculateRem(47px) 0 rgba($ibexa-color-info, 0.15);
 $ibexa-anchor-btn-hover-box-shadow: 0 0 calculateRem(14px) 0 rgba($ibexa-color-black, 0.05);
 $ibexa-context-menu-box-shadow: 0 calculateRem(22px) calculateRem(24px) 0 rgba($ibexa-color-info, 0.15);
+$ibexa-content-type-edit-filed-shadow: calculateRem(4px) calculateRem(22px) calculateRem(19px) calculateRem(-10px) rgba($ibexa-color-info, 0.05);
```
tischsoic, line 431: typo?
```suggestion
$ibexa-content-type-edit-field-shadow: calculateRem(4px) calculateRem(22px) calculateRem(19px) calculateRem(-10px) rgba($ibexa-color-info, 0.05);
```

tischsoic: Should we be adding such specific styles to `_customs.scss`?

src/bundle/Resources/public/js/scripts/core/collapse.js
```diff
@@ -0,0 +1,21 @@
+(function (global, doc, eZ) {
+    doc.querySelectorAll('.ibexa-collapse').forEach((collapseNode) => {
+        const toggleButton = collapseNode.querySelector('.ibexa-btn--toggle');
```
tischsoic, line 3: I remember that we discussed in private, that we should change this?

tischsoic (changes requested):

src/bundle/Resources/public/js/scripts/admin.contenttype.edit.js
```diff
@@ -0,0 +1,341 @@
...
+            },
+            body: JSON.stringify(bodyData),
+        });
+    };
+    const afterChangeGroup = () => {
+        const submitBtn = doc.querySelector('.ibexa-content-type-edit__publish-content-type');
+        const numberOfFieldsDefinition = doc.querySelectorAll('.ibexa-collapse--field-definition').length;
+
+        doc.querySelectorAll('.ibexa-collapse--field-definitions-group').forEach((group) => {
+            const numberOfFields = group.querySelectorAll('.ibexa-collapse--field-definition').length;
```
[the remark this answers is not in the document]
lucasOsti: `doc.querySelectorAll('.ibexa-collapse--field-definition').length` "count" all fields and is using for submit button and `group.querySelectorAll('.ibexa-collapse--field-definition').length` "count" fields in group and is using for hide/show placeholders in group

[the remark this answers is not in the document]
tischsoic: Yes, I saw this, but the type of items is the same in both cases, both queries are querying for fieldsDefinitions. This is what I was trying to say. :)

dew326 (approved):

tischsoic (approved):

Steveb-p:
Tests fixed. Changes in behavior for the `ContentTypeData` class were not properly reflected in tests.

Considering that the behavior had changed, someone with more experience with those parts of code **should definitely take a look at it**. The changes are not made to a class in `Contract` namespace, but it might break one's implementation of forms, so it might cause a "soft" BC break.

:point_up: Summoning @alongosz (due to architecture design) & @webhdx (you modified those files in 2019) :mage_man: 

Please, for humanity, resolve the discussions that are no longer relevant. Thanks in advance.

Steveb-p (dismissed):
Blocking merge until the PHP code gets reviewed.

src/bundle/Controller/ContentTypeController.php
```diff
@@ -523,9 +530,49 @@ public function updateAction(
...
         ]);
     }
 
+    public function addFieldDefinitionFormAction(
+        Request $request,
+        ContentTypeGroup $group,
+        ContentTypeDraft $contentTypeDraft,
+        string $fieldDefinitionIdentifier,
+        ?Language $language = null,
+        ?Language $baseLanguage
```
Steveb-p, line 543: Incorrect order of optional & mandatory arguments

```suggestion
        ?Language $language = null,
        ?Language $baseLanguage = null
```

mikadamczyk, line 543:
```suggestion
        ?Language $baseLanguage = null
```

src/bundle/Controller/ContentTypeController.php
```diff
@@ -523,9 +530,49 @@ public function updateAction(
...
+    ): Response {
+        $this->denyAccessUnlessGranted(new Attribute('class', 'update'));
+
+        if (!$language) {
+            $language = $this->getDefaultLanguage($contentTypeDraft);
+        }
```
Steveb-p, line 549:
```suggestion
        $language ??= $this->getDefaultLanguage($contentTypeDraft);
```

src/lib/Behat/Page/ContentTypeUpdatePage.php
```diff
@@ -8,103 +8,95 @@
...
+        $listElement = $this->getHTMLPage()
+            ->findAll($availableFieldLabel)
+            ->getByCriterion(new ElementTextCriterion($fieldName));
+        $listElement->mouseOver();
+
+        $fieldPosition = array_search(
+                $fieldName,
+                $this->getHTMLPage()->findAll($this->getLocator('availableFieldLabelList'))->mapBy(new ElementTextMapper()),
+                true
+            ) + 1; // CSS selectors are 1-indexed
```
Steveb-p, line 66: Nitpick: wrong indentation.

src/lib/Behat/Page/ContentTypeUpdatePage.php
```diff
@@ -8,103 +8,95 @@
...
                 new ElementExistsCondition($this->getHTMLPage(), $this->getLocator('fieldDefinitionContainer'))
             );
-        $this->getFieldDefinition($fieldName)->assert()->isVisible();
-        $this->notification->verifyIsLoaded();
-        $this->notification->verifyAlertSuccess();
-        $this->notification->closeAlert();
     }
 
-    private function getFieldDefinition($fieldName): ElementInterface
+    public function clickAddButton()
```
Steveb-p, line 81: Missing return type

```suggestion
    public function clickAddButton(): void
```

src/lib/Behat/Page/ContentTypeUpdatePage.php
```diff
@@ -8,103 +8,95 @@
...
+    }
 
-        return $this->getHTMLPage()
-            ->findAll($this->getLocator('fieldDefinitionContainer'))
-            ->filter(function (ElementInterface $element) use ($fieldTypeIdentifier) {
-                return false !== strpos($element->find($this->getLocator('fieldDefinitionName'))->getText(), $fieldTypeIdentifier);
-            })
-            ->first()
-        ;
+    public function selectContentTypeCategory(string $categoryName)
```
Steveb-p, line 93:
```suggestion
    public function selectContentTypeCategory(string $categoryName): void
```

src/bundle/Resources/config/routing.yaml
```diff
@@ -389,6 +389,18 @@ ezplatform.content_type.remove_translation:
     defaults:
         _controller: 'EzSystems\EzPlatformAdminUiBundle\Controller\ContentTypeController::removeTranslationAction'
 
+ibexa.content_type.field_definition_form:
+    path: /contenttypegroup/{contentTypeGroupId}/contenttype/{contentTypeId}/field_definition_form/{fieldDefinitionIdentifier}/{toLanguageCode}/{fromLanguageCode}
+    methods: ['GET']
+    options:
+        expose: true
+    defaults:
+        _controller: 'EzSystems\EzPlatformAdminUiBundle\Controller\ContentTypeController::addFieldDefinitionFormAction'
```
mikadamczyk, line 398:
```suggestion
    controller: 'EzSystems\EzPlatformAdminUiBundle\Controller\ContentTypeController::addFieldDefinitionFormAction'
    defaults:
```

src/lib/Form/Data/ContentTypeData.php
```diff
@@ -23,11 +22,8 @@ class ContentTypeData extends ContentTypeUpdateStruct implements NewnessCheckabl
      */
     use NewnessChecker;
 
-    /** @var \eZ\Publish\API\Repository\Values\ContentType\ContentTypeDraft */
-    protected $contentTypeDraft;
-
-    /** @var \EzSystems\EzPlatformAdminUi\Form\Data\FieldDefinitionData[] */
-    protected $fieldDefinitionsData = [];
+    /** @var \EzSystems\EzPlatformAdminUi\Form\Data\FieldDefinitionData[][] */
+    public $fieldDefinitionsData = [];
```
Steveb-p, line 26: This has potential for disaster. I've checked and updated tests that relied on the fact that the array was one dimensional. Once grouping is introduced it becomes a lot harder to follow.

If possible, both grouped and ungrouped variants should be reachable. For example, grouped variant could be returned from a new method. Grouping relies on `FieldDefinitionData` properties (`fieldGroup`, `identifier`) anyway, so it can be done on demand in my opinion.

Additionally, why is the property changed to `public` visibility?

src/lib/Form/Type/ContentType/ContentTypeUpdateType.php
```diff
@@ -102,24 +102,18 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+                'mainLanguageCode' => $options['mainLanguageCode'],
             ])
+//            ->add('fieldDefinitionsData', CollectionType::class, [
+//                'entry_type' => FieldDefinitionType::class,
+//                'entry_options' => ['languageCode' => $options['languageCode'], 'mainLanguageCode' => $options['mainLanguageCode']],
+//                'label' => /** @Desc("Content Field definitions") */ 'content_type.field_definitions_data',
+//                'allow_add' => true,
+//                'allow_delete' => true,
+//                'prototype' => false,
+//            ])
```
mikadamczyk, line 116:
```suggestion
```

Steveb-p, line 116: Leftover comment?

src/lib/Form/Type/ContentType/FieldDefinitionsCollectionType.php
```diff
@@ -0,0 +1,63 @@
...
+            ]);
+        }
+    }
+
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver
+            ->setDefaults([
+                'translation_domain' => 'content_type',
+                'mainLanguageCode' => null,
```
mikadamczyk, line 57:
```suggestion
                'main_language_code' => null,
```

src/lib/REST/Value/ContentType/FieldDefinitionCreate.php
```diff
@@ -0,0 +1,30 @@
...
+    /** @var string|null */
+    public $fieldTypeIdentifier;
+
+    /** @var string|null */
+    public $fieldGroupIdentifier;
+
+    /** @var int|null */
+    public $position;
+
+    public function __construct(?string $fieldTypeIdentifier, ?string $fieldGroupIdentifier, ?int $position = null)
```
mikadamczyk, line 24:
```suggestion
    public function __construct(?string $fieldTypeIdentifier = null, ?string $fieldGroupIdentifier = null, ?int $position = null)
```

Steveb-p: @mikadamczyk in this case I believe it's expected that this argument is always explicitly passed, even if `null`.

src/bundle/Controller/ContentTypeController.php
Steveb-p: Since it's not available until PHP 7.4, this has to be dropped.

src/lib/Form/Data/ContentTypeData.php
Steveb-p: I've analyzed the Ibexa Commerce code and usage of this property. While I'm not okay with the fact that we're doing it like this (basically changing how the property, which already exposed, consists of) considering the alternative (introducing collection object) I'm ok with it.

It has to be public for the form component to interact with the array correctly when setting new values.

src/lib/Form/Type/ContentType/ContentTypeUpdateType.php
Steveb-p: I've removed this comment.

src/bundle/Controller/ContentTypeController.php
Steveb-p: Added missing default value (since it's after a different optional argument)

src/lib/Form/Type/ContentType/FieldDefinitionsCollectionType.php
Steveb-p: :-1: because it's an option that is passed from the parent. If it weren't so, it would be ok to change it, but because of this if we were to use snake_case, we would have two different variants in use.

src/lib/REST/Value/ContentType/FieldDefinitionCreate.php
Steveb-p: @mikadamczyk yeah, it always has a value passed to it. I'd prefer to be explicit here, since we're talking about field type identifier & field group identifier.

src/bundle/Resources/config/routing.yaml
Steveb-p: Fixed.

mikadamczyk (approved):

src/bundle/Controller/FieldDefinitionController.php
```diff
@@ -0,0 +1,164 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformAdminUiBundle\Controller;
```
ViniTou, line 9: This needs new NS in all places.

ViniTou (changes requested):

src/lib/Form/Type/ContentType/FieldDefinitionsCollectionType.php
webhdx: Then I would just proceed and change the option name in the parent as well. It kinda looks bad if we use snake case for options, contradicts with what Symfony does.

webhdx (approved):

src/lib/Form/Type/ContentType/FieldDefinitionsCollectionType.php
Steveb-p: Should we deprecate it (with `trigger_error`) in `3.x` then, and silently convert to snake_case?

This PR targets `4.x`, so a small PR would be going first for this specifically?

Or do we ignore it, treating it as an internal thing?

alongosz: > Should we deprecate it (with `trigger_error`) in `3.x` then, and silently convert to snake_case?
> 
> This PR targets `4.x`, so a small PR would be going first for this specifically?
> 
> Or do we ignore it, treating it as an internal thing?

Depends it we documented extending and relying on those options.

src/lib/Tests/Form/Data/ContentTypeDataTest.php
```diff
@@ -23,52 +23,99 @@ public function testContentTypeDraft()
 
     public function testFieldDefinitionData()
```
alongosz, line 24:
```suggestion
    public function testFieldDefinitionData(): void
```

src/lib/Tests/Form/Data/ContentTypeDataTest.php
```diff
@@ -23,52 +23,99 @@ public function testContentTypeDraft()
...
+
+        self::assertSame([
+            'field_group__alpha.identifier__alpha' => $fieldDef1,
+            'field_group__bravo.identifier__bravo' => $fieldDef2,
+            'field_group__charlie.identifier__charlie' => $fieldDef3,
+            'field_group__delta.identifier__delta' => $fieldDef4,
+        ], iterator_to_array($data->getFlatFieldDefinitionsData()));
     }
 
     public function testSortFieldDefinitions()
```
alongosz, line 69:
```suggestion
    public function testSortFieldDefinitions(): void
```

src/lib/Tests/REST/Input/ContentType/FieldDefinitionCreateTest.php
```diff
@@ -0,0 +1,62 @@
...
+    private $parser;
+
+    protected function setUp(): void
+    {
+        $this->parser = new FieldDefinitionCreate();
+    }
+
+    public function testValidInput(): void
+    {
+        $this->assertEquals(
```
alongosz, line 29:
```suggestion
        self::assertEquals(
```

src/lib/Tests/REST/Input/ContentType/FieldDefinitionCreateTest.php
```diff
@@ -0,0 +1,62 @@
...
+            new FieldDefinitionCreateValue('ezstring', null),
+            $this->parser->parse(
+                [
+                    'fieldTypeIdentifier' => 'ezstring',
+                ],
+                $this->createMock(ParsingDispatcher::class)
+            )
+        );
+
+        $this->assertEquals(
```
alongosz, line 39:
```suggestion
        self::assertEquals(
```

src/lib/Tests/REST/Input/ContentType/FieldDefinitionDeleteTest.php
```diff
@@ -0,0 +1,50 @@
...
+    private $parser;
+
+    protected function setUp(): void
+    {
+        $this->parser = new FieldDefinitionDelete();
+    }
+
+    public function testValidInput(): void
+    {
+        $this->assertEquals(
```
alongosz, line 29:
```suggestion
        self::assertEquals(
```

src/lib/Tests/REST/Input/ContentType/FieldDefinitionReorderTest.php
```diff
@@ -0,0 +1,50 @@
...
+    private $parser;
+
+    protected function setUp(): void
+    {
+        $this->parser = new FieldDefinitionReorder();
+    }
+
+    public function testValidInput(): void
+    {
+        $this->assertEquals(
```
alongosz, line 29:
```suggestion
        self::assertEquals(
```

src/lib/UI/Module/FieldTypeToolbar/FieldTypeToolbarFactory.php
```diff
@@ -0,0 +1,65 @@
...
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformAdminUi\UI\Module\FieldTypeToolbar;
+
+use eZ\Publish\Core\FieldType\FieldTypeRegistry;
+use EzSystems\EzPlatformAdminUi\UI\Module\FieldTypeToolbar\Values\FieldTypeToolbar;
+use EzSystems\EzPlatformAdminUi\UI\Module\FieldTypeToolbar\Values\FieldTypeToolbarItem;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+final class FieldTypeToolbarFactory
```
alongosz, line 16: is it internal?

alongosz (changes requested):
From my side mostly cosmetic changes. What stands out though is usage of legacy namespaces, but that was already mentioned by other Reviewers

src/bundle/Controller/FieldDefinitionController.php
Steveb-p: Namespaces introduced.

src/lib/Form/Type/ContentType/FieldDefinitionsCollectionType.php
Steveb-p: @alongosz I've looked through our docs, and any mentions of `mainLanguageCode` are in context of Content objects and their structs, not their forms. So in theory we could simply replace `mainLanguageCode` option with `main_language_code` whenever this form type is used, treating it as internal.

Is that ok with you guys? @alongosz @webhdx  @mikadamczyk ?

alongosz: > @alongosz I've looked through our docs, and any mentions of `mainLanguageCode` are in context of Content objects and their structs, not their forms. So in theory we could simply replace `mainLanguageCode` option with `main_language_code` whenever this form type is used, treating it as internal.
> 
> Is that ok with you guys? @alongosz @webhdx @mikadamczyk ?

No objections from my side, but this is mostly up to the AdminUI Maintainers ;)

mikadamczyk: I am ok with this

Steveb-p: Considering how those forms are created, I'm not certain that I should go forward with this change anymore:

image

Unfortunately that's also not the only instance that `"mainLanguageCode"` is used as an option. To maintain consistency I'd leave it as-is, unless you'd like that option with automatic conversion - although argument can be made that it should be done completely separately, and with all instances in consideration.

Therefore, out of scope?

src/lib/Tests/REST/Input/ContentType/FieldDefinitionReorderTest.php
Steveb-p: Fixed in 4ac54ec2.

src/lib/Tests/REST/Input/ContentType/FieldDefinitionDeleteTest.php
Steveb-p: Fixed in 4ac54ec2.

src/lib/Tests/REST/Input/ContentType/FieldDefinitionCreateTest.php
Steveb-p: Fixed in 4ac54ec2.

src/lib/Tests/REST/Input/ContentType/FieldDefinitionCreateTest.php
Steveb-p: Fixed in 4ac54ec2.

src/lib/Tests/Form/Data/ContentTypeDataTest.php
Steveb-p: Fixed in e6a2a544.

src/lib/Tests/Form/Data/ContentTypeDataTest.php
Steveb-p: Technically was left as-is probably because it was already present. Considering that tests are not covered by BC promises, I'm adjusting this file.

Fixed in e6a2a544.

src/bundle/Controller/ContentTypeController.php
```diff
@@ -523,9 +530,49 @@ public function updateAction(
...
+        foreach ($form['fieldDefinitionsData'] as $fieldDefinitionsGroupForm) {
+            if (isset($fieldDefinitionsGroupForm[$fieldDefinitionIdentifier])) {
+                return $this->render('@ezdesign/content_type/part/field_definition_form.html.twig', [
+                    'form' => $fieldDefinitionsGroupForm[$fieldDefinitionIdentifier]->createView(),
+                    'content_type_group' => $group,
+                    'content_type' => $contentTypeDraft,
+                    'language_code' => $baseLanguage ? $baseLanguage->languageCode : $language->languageCode,
+                    'is_translation' => $contentTypeDraftData->mainLanguageCode !== $contentTypeDraftData->languageCode,
+                ]);
+            }
```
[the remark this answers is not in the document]
Steveb-p: Fixed in 7ef1f67f.

src/bundle/Controller/ContentTypeController.php
```diff
@@ -523,9 +530,49 @@ public function updateAction(
...
+        Request $request,
+        ContentTypeGroup $group,
+        ContentTypeDraft $contentTypeDraft,
+        string $fieldDefinitionIdentifier,
+        ?Language $language = null,
+        ?Language $baseLanguage = null
+    ): Response {
+        $this->denyAccessUnlessGranted(new Attribute('class', 'update'));
+
+        if (!$language) {
```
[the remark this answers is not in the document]
Steveb-p: Fixed in d04ac188.

Are we prefering yoda style conditions btw?

src/lib/UI/Module/FieldTypeToolbar/FieldTypeToolbarFactory.php
Steveb-p: It seems it's only used for `ContentTypeController`, to provide parameters to view.

So I guess I'll mark it as `@internal` unless otherwise. @adamwojs might want to answer this once he's back. Maybe... :sweat_smile: 

Fixed in 1959048b.

composer.json
```diff
@@ -7,11 +7,15 @@
...
           "EzSystems\\EzPlatformAdminUi\\": "src/lib/",
+          "Ibexa\\Bundle\\AdminUi\\": "src/bundle/",
           "Ibexa\\AdminUi\\": "src/lib/"
         }
     },
     "autoload-dev": {
-        "psr-4": { "EzSystems\\EzPlatformAdminUi\\Tests\\": "src/lib/Tests" }
+        "psr-4": {
+            "EzSystems\\EzPlatformAdminUi\\Tests\\": "src/lib/Tests",
+            "Ibexa\\AdminUi\\Tests\\": "src/lib/Tests"
```
alongosz, line 17: This does not follow [package structure](https://doc.ibexa.co/en/master/community_resources/package_structure/) requirements.
```suggestion
            "Ibexa\\Tests\\AdminUi\\": "tests/lib"
```

alongosz (changes requested):

composer.json
Steveb-p: Fixed in ae4faf3e.

Includes added tests directory to code style configuration & phpunit.xml adjustment.

src/bundle/Resources/config/services/controllers.yaml
```diff
@@ -1,7 +1,7 @@
 services:
     EzSystems\EzPlatformAdminUiBundle\Controller\:
         resource: "../../Controller/*"
-        exclude: "../../Controller/{Controller}"
+        exclude: "../../Controller/{FieldDefinitionController.php}"
```
alongosz, line 4: TBH I don't like this workaround. I'd rather go for expanding this definition into specific services. It feels like explicitness is more clear and understandable.

This is remark rather for the future, for now let's leave as-is, since it was given as one of the options.

cc @Nattfarinn

alongosz (approved):

src/bundle/Resources/config/services/controllers.yaml
Steveb-p: Agreed, but the PR is big as it is, and from what I've seen I'd need to add most of the classes that reside in the subdirectory, which is around ~30 I believe?

Steveb-p:
@ViniTou new namespaces have been introduced. You'd like to review the code again, or should I dismiss your review?

ViniTou (approved):

mnocon (approved):

PR #1836, opened 2021-08-04
IBX-823: Enable/Disable validation form before send workflow review

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-823](https://issues.ibexa.co/browse/IBX-823)
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.content.edit.js
+++ b/src/bundle/Resources/public/js/scripts/admin.content.edit.js
@@ -70,6 +70,12 @@
 
         event.preventDefault();
 
+        if (isFormValid(btn)) {
+            btn.click();
+        }
+    };
+
+    const isFormValid = (btn) => {
         const validators = eZ.fieldTypeValidators;
         const validationResults = validators.map(getValidationResults);
         const isFormValid = validationResults.every((result) => result.isValid);
@@ -77,34 +83,34 @@
 
         if (isFormValid) {
             btn.dataset.isFormValid = 1;
-            // for some reason trying to fire click event inside the event handler flow is impossible
-            // the following line breaks the flow so it's possible to fire click event on a button again.
-            window.setTimeout(() => btn.click(), 0);
-        } else {
-            btn.dataset.validatorsWithErrors = Array.from(
-                validationResults
-                    .filter((result) => !result.isValid)
-                    .reduce((total, result) => {
-                        total.add(result.validatorName);
 
-                        return total;
-                    }, new Set())
-            ).join();
-
-            fields.forEach((field) => field.removeAttribute('id'));
-
-            doc.querySelectorAll('.ez-tabs__nav-item').forEach((navItem) => {
-                navItem.classList.remove('ez-tabs__nav-item--invalid');
-            });
-
-            invalidTabs.forEach((invalidInputs) => {
-                invalidInputs.forEach((invalidInputKey) => {
-                    doc.querySelector(`#item-${invalidInputKey}`).classList.add('ez-tabs__nav-item--invalid');
-                });
-            });
-
-            focusOnFirstError();
+            return true;
         }
+        btn.dataset.validatorsWithErrors = Array.from(
+            validationResults
+                .filter((result) => !result.isValid)
+                .reduce((total, result) => {
+                    total.add(result.validatorName);
+
+                    return total;
+                }, new Set())
+        ).join();
+
+        fields.forEach((field) => field.removeAttribute('id'));
+
+        doc.querySelectorAll('.ez-tabs__nav-item').forEach((navItem) => {
+            navItem.classList.remove('ez-tabs__nav-item--invalid');
+        });
+
+        invalidTabs.forEach((invalidInputs) => {
+            invalidInputs.forEach((invalidInputKey) => {
+                doc.querySelector(`#item-${invalidInputKey}`).classList.add('ez-tabs__nav-item--invalid');
+            });
+        });
+
+        focusOnFirstError();
+
+        return false;
     };
 
     const isAutosaveEnabled = () => {
@@ -166,6 +172,6 @@
 
     menuButtonsToValidate.forEach((btn) => {
         btn.dataset.isFormValid = 0;
-        btn.addEventListener('click', clickHandler, false);
+        btn.addEventListener('click', isFormValid, false);
     });
 })(window, window.document, window.eZ, window.Translator);
--- a/src/bundle/Resources/public/js/scripts/admin.content.edit.js
+++ b/src/bundle/Resources/public/js/scripts/admin.content.edit.js
@@ -75,10 +75,12 @@
         }
     };
     const validateHandler = (event) => {
+        event.preventDefault();
+
         const btn = event.currentTarget;
+
         btn.dataset.isFormValid = 0;
 
-        event.preventDefault();
         isFormValid(btn);
     };
     const isFormValid = (btn) => {
@@ -92,6 +94,7 @@
 
             return true;
         }
+
         btn.dataset.validatorsWithErrors = Array.from(
             validationResults
                 .filter((result) => !result.isValid)
--- a/src/bundle/Resources/public/js/scripts/admin.content.edit.js
+++ b/src/bundle/Resources/public/js/scripts/admin.content.edit.js
@@ -71,7 +71,7 @@
         event.preventDefault();
 
         if (isFormValid(btn)) {
-            btn.click();
+            window.setTimeout(() => btn.click(), 0);
         }
     };
 
--- a/src/bundle/Resources/public/js/scripts/admin.content.edit.js
+++ b/src/bundle/Resources/public/js/scripts/admin.content.edit.js
@@ -74,7 +74,13 @@
             window.setTimeout(() => btn.click(), 0);
         }
     };
+    const validateHandler = (event) => {
+        const btn = event.currentTarget;
+        btn.dataset.isFormValid = 0;
 
+        event.preventDefault();
+        isFormValid(btn);
+    };
     const isFormValid = (btn) => {
         const validators = eZ.fieldTypeValidators;
         const validationResults = validators.map(getValidationResults);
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/public/js/scripts/admin.content.edit.js
```diff
@@ -148,6 +150,35 @@
...
+        const applyButton = doc.querySelector('[name="ezplatform_content_forms_content_edit[workflow][apply]"]');
+        if (workflowEventListener === true) {
+            return;
+        }
+
+        applyButton.dataset.isFormValid = 0;
+        workflowEventListener = true;
+        applyButton.addEventListener('click', clickHandler, false);
+    }
+
```
GrabowskiM, line 163: unnecessary new lines between variables (arrow functions) declarations

src/bundle/Resources/public/js/scripts/admin.content.edit.js
```diff
@@ -148,6 +150,35 @@
         }, eZ.adminUiConfig.autosave.interval);
     }
 
+    const addClickWorkflowHandler = (event) => {
+        const applyButton = doc.querySelector('[name="ezplatform_content_forms_content_edit[workflow][apply]"]');
+        if (workflowEventListener === true) {
```
GrabowskiM, line 155: new line before if

GrabowskiM, line 155:
```suggestion
        if (workflowEventListener) {
```

src/bundle/Resources/public/js/scripts/admin.content.edit.js
```diff
@@ -148,6 +150,35 @@
...
+        }
+
+        applyButton.dataset.isFormValid = 0;
+        workflowEventListener = true;
+        applyButton.addEventListener('click', clickHandler, false);
+    }
+
+    const removeClickWorkflowHandler = (event) => {
+        const applyButton = doc.querySelector('[name="ezplatform_content_forms_content_edit[workflow][apply]"]');
+        if (workflowEventListener === false) {
```
GrabowskiM, line 166: new line before if

GrabowskiM, line 166:
```suggestion
        if (!workflowEventListener) {
```

GrabowskiM (changes requested):

src/bundle/Resources/public/js/scripts/admin.content.edit.js
```diff
@@ -148,6 +150,36 @@
         }, eZ.adminUiConfig.autosave.interval);
     }
```
lucasOsti, line 152:
```suggestion

```

lucasOsti (changes requested):

GrabowskiM (approved):

lucasOsti (approved):

src/bundle/Resources/public/js/scripts/admin.content.edit.js
```diff
@@ -23,6 +23,8 @@
     ];
     const form = doc.querySelector('.ez-form-validate');
     const submitBtns = form.querySelectorAll('[type="submit"]:not([formnovalidate])');
+    const buttonsMenuWorkflow = doc.querySelectorAll('button[id^="workflow__apply__"]');
```
dew326, line 26:
```suggestion
    const menuWorkflowButtons = doc.querySelectorAll('button[id^="workflow__apply__"]');
```

src/bundle/Resources/public/js/scripts/admin.content.edit.js
```diff
@@ -147,6 +149,35 @@
                 });
         }, eZ.adminUiConfig.autosave.interval);
     }
+    const addClickWorkflowHandler = (event) => {
```
dew326, line 152:
```suggestion
    const addClickWorkflowHandler = () => {
```

src/bundle/Resources/public/js/scripts/admin.content.edit.js
```diff
@@ -147,6 +149,35 @@
...
+
+        if (workflowEventListener) {
+            return;
+        }
+
+        applyButton.dataset.isFormValid = 0;
+        workflowEventListener = true;
+        applyButton.addEventListener('click', clickHandler, false);
+    }
+    const removeClickWorkflowHandler = (event) => {
```
dew326, line 163:
```suggestion
    const removeClickWorkflowHandler = () => {
```

src/bundle/Resources/public/js/scripts/admin.content.edit.js
```diff
@@ -147,6 +149,35 @@
...
+
+        if (!workflowEventListener) {
+            return;
+        }
+
+        workflowEventListener = false;
+        applyButton.removeEventListener('click', clickHandler, false);
+    }
+
+    buttonsMenuWorkflow.forEach((buttonMenuWorkflow) => {
```
dew326, line 174: I think the event listener will depend only on last item from the array. I mean if the lest item has property validate it will be attached if not if will be removed, or am I missing something?

dew326 (changes requested):
Few minor issues and some questions but the biggest problem is that admin-ui has no dependency on workflow so code related to workflow cannot be in admin-ui.

src/bundle/Resources/public/js/scripts/admin.content.edit.js
```diff
@@ -147,6 +149,35 @@
...
     }
+    const addClickWorkflowHandler = (event) => {
+        const applyButton = doc.querySelector('[name="ezplatform_content_forms_content_edit[workflow][apply]"]');
+
+        if (workflowEventListener) {
+            return;
+        }
+
+        applyButton.dataset.isFormValid = 0;
+        workflowEventListener = true;
```
tischsoic, line 160: Actually, I wonder whether we need this additional condition because you cannot add the same event listener twice with `addEventListener`.
ref. https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#other_notes

Also, to quote MDN: "calling removeEventListener() with arguments that do not identify any currently registered EventListener on the EventTarget has no effect."
ref. https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#notes

Because as far as I can see, you add the same `clickHandler` regardless the workflow button? I think that this logic can be simplified.
But maybe I don't understand something, or I overlooked something, please let me know. 🙂

tischsoic (changes requested):

src/bundle/Resources/public/js/scripts/admin.content.edit.js
tischsoic: I didn't see Darek's comments before review.

src/bundle/Resources/public/js/scripts/admin.content.edit.js
mateuszdebinski: addClickWorkflowHandler and removeClickWorkflowHandler  it is added only once for a given workflow button in the right menu (there can be several) and for "applyButton" event listener clickHandler is added once or removed, therefore there is a flag that checks whether it has already been added or not before adding/removing this EventListener

src/bundle/Resources/public/js/scripts/admin.content.edit.js
```diff
@@ -23,6 +23,9 @@
     ];
     const form = doc.querySelector('.ez-form-validate');
     const submitBtns = form.querySelectorAll('[type="submit"]:not([formnovalidate])');
+    const menuWorkflowButtons = doc.querySelectorAll('button[id^="workflow__apply__"]');
```
dew326, line 26: Still the same issue, workflow code cannot be in admin-ui

dew326 (changes requested):

src/bundle/Resources/public/js/scripts/admin.content.edit.js
mateuszdebinski: I working on this, I change this on WiP

src/bundle/Resources/public/js/scripts/admin.content.edit.js
```diff
@@ -69,43 +70,54 @@
...
+        if (isFormValid(btn)) {
+            window.setTimeout(() => btn.click(), 0);
+        }
+    };
+    const validateHandler = (event) => {
+        const btn = event.currentTarget;
+        btn.dataset.isFormValid = 0;
+
+        event.preventDefault();
+        isFormValid(btn);
```
dew326, line 82:
```suggestion
        event.preventDefault();
        
        const btn = event.currentTarget;
        
        btn.dataset.isFormValid = 0;
        
        isFormValid(btn);
```

src/bundle/Resources/public/js/scripts/admin.content.edit.js
```diff
@@ -69,43 +70,54 @@
...
-                    .filter((result) => !result.isValid)
-                    .reduce((total, result) => {
-                        total.add(result.validatorName);
 
-                        return total;
-                    }, new Set())
-            ).join();
+            return true;
+        }
+        btn.dataset.validatorsWithErrors = Array.from(
```
dew326, line 95:
```suggestion

        btn.dataset.validatorsWithErrors = Array.from(
```

dew326 (changes requested):

src/bundle/Resources/public/js/scripts/admin.content.edit.js
```diff
@@ -69,43 +70,57 @@
 
         event.preventDefault();
 
+        if (isFormValid(btn)) {
+            window.setTimeout(() => btn.click(), 0);
```
dew326, line 74: I would keep the comment with explanations why this has to be in timeout
Also could you change `window` to `global`.

dew326 (changes requested):

src/bundle/Resources/public/js/scripts/admin.content.edit.js
mateuszdebinski: done

dew326 (approved):

tischsoic (approved):

bogusez (approved):

PR #1847, opened 2021-08-11
IBX-620: Redesigned ezauthor

dew326:
| Question      | Answer
| ------------- | ---
| Tickets       | https://issues.ibexa.co/browse/IBX-620
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no
| License       | [GPL-2.0](https://github.com/ezsystems/ezplatform-admin-ui/blob/master/LICENSE)

#### Checklist:
- [x] Coding standards (`$ composer fix-cs`)
- [x] Ready for Code Review

PR #1880, opened 2021-09-01
IBX-808: As an Editor, I want to be able to reset starting location for ezobjectrelation FieldType

konradoboza:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-808](https://issues.ibexa.co/browse/IBX-808)
| Bugfix?      | no
| Improvement? | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Excerpt from JIRA:
>Currently, after selecting "Starting Location" in ezobjectrelation settings, there is no way to reset it. This might get problematic when, for some reason, the initial content is not visible in UDW due to tree_root parameter usage or associated user permissions.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.contenttype.relation.default.location.js
+++ b/src/bundle/Resources/public/js/scripts/admin.contenttype.relation.default.location.js
@@ -1,5 +1,6 @@
 (function (global, doc, React, ReactDOM) {
-    const btns = doc.querySelectorAll('.btn--udw-relation-default-location');
+    const resetStartingLocationBtns = doc.querySelectorAll('.ez-reset-starting-location');
+    const udwBtns = doc.querySelectorAll('.btn--udw-relation-default-location');
     const udwContainer = doc.getElementById('react-udw');
     const token = doc.querySelector('meta[name="CSRF-Token"]').content;
     const siteaccess = doc.querySelector('meta[name="SiteAccess"]').content;
@@ -11,6 +12,11 @@
         const locationName = items[0].ContentInfo.Content.TranslatedName;
         const objectRelationListSettingsWrapper = btn.closest('.ezobjectrelationlist-settings');
         const objectRelationSettingsWrapper = btn.closest('.ezobjectrelation-settings');
+
+        toggleResetStartingLocationBtn(
+            btn.parentNode.querySelector('.ez-reset-starting-location'),
+            true
+        );
 
         if (objectRelationListSettingsWrapper) {
             objectRelationListSettingsWrapper.querySelector(btn.dataset.relationRootInputSelector).value = locationId;
@@ -35,6 +41,21 @@
             restInfo: { token, siteaccess }
         }, config)), udwContainer);
     };
+    const toggleResetStartingLocationBtn = (button, isEnabled) => {
+        isEnabled ? button.removeAttribute('disabled') : button.setAttribute('disabled', true);
+    };
+    const resetStartingLocation = (event) => {
+        const button = event.target.closest('button');
+        const buttonDataset = button.dataset;
+        const relationRootInputSelector = buttonDataset.relationRootInputSelector;
+        const relationSelectedRootNameSelector = buttonDataset.relationSelectedRootNameSelector;
 
-    btns.forEach(btn => btn.addEventListener('click', openUDW, false));
+        doc.querySelector(relationRootInputSelector).value = '';
+        doc.querySelector(relationSelectedRootNameSelector).innerHTML = '';
+
+        toggleResetStartingLocationBtn(button, false);
+    };
+
+    udwBtns.forEach(btn => btn.addEventListener('click', openUDW, false));
+    resetStartingLocationBtns.forEach(btn => btn.addEventListener('click', resetStartingLocation, false));
 })(window, window.document, window.React, window.ReactDOM);
--- a/src/bundle/Resources/public/js/scripts/admin.contenttype.relation.default.location.js
+++ b/src/bundle/Resources/public/js/scripts/admin.contenttype.relation.default.location.js
@@ -44,7 +44,7 @@
     };
     const toggleResetStartingLocationBtn = (button, isEnabled) => {
         if (isEnabled) {
-            button.removeAttribute('disabled')
+            button.removeAttribute('disabled');
         } else {
             button.setAttribute('disabled', true);
         }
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/public/js/scripts/admin.contenttype.relation.default.location.js
```diff
@@ -1,5 +1,6 @@
 (function (global, doc, React, ReactDOM) {
-    const btns = doc.querySelectorAll('.btn--udw-relation-default-location');
+    const resetStartingLocationBtns = doc.querySelectorAll('.ez-reset-starting-location');
```
dew326, line 2:
```suggestion
    const resetStartingLocationBtns = doc.querySelectorAll('.ez-btn--reset-starting-location');
```

src/bundle/Resources/public/js/scripts/admin.contenttype.relation.default.location.js
```diff
@@ -35,6 +41,21 @@
             restInfo: { token, siteaccess }
         }, config)), udwContainer);
     };
+    const toggleResetStartingLocationBtn = (button, isEnabled) => {
+        isEnabled ? button.removeAttribute('disabled') : button.setAttribute('disabled', true);
+    };
+    const resetStartingLocation = (event) => {
+        const button = event.target.closest('button');
```
dew326, line 48: Your button should in `event.currentTarget`

src/bundle/Resources/public/js/scripts/admin.contenttype.relation.default.location.js
```diff
@@ -35,6 +41,21 @@
...
+        const buttonDataset = button.dataset;
+        const relationRootInputSelector = buttonDataset.relationRootInputSelector;
+        const relationSelectedRootNameSelector = buttonDataset.relationSelectedRootNameSelector;
```
dew326, line 51:
```suggestion
        const { relationRootInputSelector, relationSelectedRootNameSelector } = button.dataset;
```

src/bundle/Resources/public/js/scripts/admin.contenttype.relation.default.location.js
```diff
@@ -35,6 +41,21 @@
             restInfo: { token, siteaccess }
         }, config)), udwContainer);
     };
+    const toggleResetStartingLocationBtn = (button, isEnabled) => {
+        isEnabled ? button.removeAttribute('disabled') : button.setAttribute('disabled', true);
```
dew326, line 45: We are using thernary more for assignment, not for invoking functions.
I would see it should something like this:
```js
    const methodName = isEnabled ? 'removeAttribute' : 'setAttribute';
    
    button[methodName]('disabled', 'disabled');
```

src/bundle/Resources/views/admin/content_type/field_types.html.twig
```diff
@@ -142,6 +142,19 @@
             >{{ "field_definition.ezobjectrelation.selection_root_udw_button"
                 |trans({}, "ezrepoforms_content_type")
                 |desc("Select item") }}</button>
+            <button type="button"
```
dew326, line 145:
```suggestion
            <button 
                type="button"
```

src/bundle/Resources/views/admin/content_type/field_types.html.twig
```diff
@@ -174,6 +187,19 @@
             >{{ "field_definition.ezobjectrelationlist.selection_root_udw_button"
                 |trans({}, "ezrepoforms_content_type")
                 |desc("Select location") }}</button>
+            <button type="button"
```
dew326, line 190:
```suggestion
            <button 
                type="button"
```

dew326 (changes requested):

src/bundle/Resources/public/js/scripts/admin.contenttype.relation.default.location.js
tischsoic: Actually, `toggleAttribute` was already used by Darek in admin-ui:
https://github.com/ezsystems/ezplatform-admin-ui/pull/1847/files#diff-9530ed0ec9743cd8231648df7312b0eaf421d4b5375c3217fab6249b5a3b1c54R109

konradoboza: Not sure about `toggleAttribute` but please note, that we are dealing with `1.5` AdminUi code, so I think it would be good to stay aligned and eventually improve this part starting from `2.3` onwards.

dew326: `toggleAttribute` I wouldn't use as this PR targets 2.5.

So either the way I suggested or the `regular if`. Both are fine for me. If @tischsoic hates this so much you can use the regular if.

src/bundle/Resources/public/js/scripts/admin.contenttype.relation.default.location.js
```diff
@@ -1,5 +1,7 @@
 (function (global, doc, React, ReactDOM) {
-    const btns = doc.querySelectorAll('.btn--udw-relation-default-location');
+    const RESET_STARTING_LOCATION_BTN_SELECTOR = '.ez-btn--reset-starting-location';
```
dew326, line 2:
```suggestion
    const SELECTOR_RESET_STARTING_LOCATION_BTN = '.ez-btn--reset-starting-location';
```

src/bundle/Resources/public/js/scripts/admin.contenttype.relation.default.location.js
```diff
@@ -35,6 +42,23 @@
...
+    const toggleResetStartingLocationBtn = (button, isEnabled) => {
+        if (isEnabled) {
+            button.removeAttribute('disabled')
+        } else {
+            button.setAttribute('disabled', true);
+        }
+    };
+    const resetStartingLocation = (event) => {
+        const button = event.currentTarget;
+        const {relationRootInputSelector, relationSelectedRootNameSelector} = button.dataset;
```
dew326, line 54:
```suggestion
        const { relationRootInputSelector, relationSelectedRootNameSelector } = button.dataset;
```

dew326 (changes requested):

src/bundle/Resources/public/js/scripts/admin.contenttype.relation.default.location.js
```diff
@@ -35,6 +42,23 @@
             restInfo: { token, siteaccess }
         }, config)), udwContainer);
     };
+    const toggleResetStartingLocationBtn = (button, isEnabled) => {
+        if (isEnabled) {
+            button.removeAttribute('disabled')
```
dew326, line 47:
```suggestion
            button.removeAttribute('disabled');
```

dew326 (changes requested):

GrabowskiM (approved):

dew326 (approved):

lucasOsti (approved):

kacper-wieczorek-ibexa (approved):
Qa -approved - Function 'Reset start location' at 'ezobjectrelation' is added and works properly for v2.5.23.

mikadamczyk:
Could you merge it up @konradoboza ?

PR #1893, opened 2021-09-09
IBX-976: Disabled transition for side menu after reload page

lucasOsti:
| Question      | Answer
| ------------- | ---
| Tickets       | https://issues.ibexa.co/browse/IBX-976
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/helpers/cookies.helper.js
+++ b/src/bundle/Resources/public/js/scripts/helpers/cookies.helper.js
@@ -1,0 +1,27 @@
+(function (global, doc, eZ) {
+    const setCookie = (name, value, maxAgeDays = 356, path = '/') => {
+        const maxAge = maxAgeDays * 24 * 60 * 60;
+
+        doc.cookie = `${name}=${value};max-age=${maxAge};path=${path}`;
+    };
+    const getCookie = (name) => {
+        const cookieName = name + '=';
+        const decodedCookie = decodeURIComponent(doc.cookie);
+        const cookiesArray = decodedCookie.split(';');
+
+        for (index in cookiesArray) {
+            const cookieString = cookiesArray[index].trim();
+
+            if (cookieString.indexOf(cookieName) === 0) {
+                return cookieString.split('=')[1];
+            }
+        }
+
+        return '';
+    };
+
+    eZ.addConfig('helpers.cookies', {
+        getCookie,
+        setCookie,
+    });
+})(window, window.document, window.eZ);
--- a/src/bundle/Resources/public/js/scripts/helpers/cookies.helper.js
+++ b/src/bundle/Resources/public/js/scripts/helpers/cookies.helper.js
@@ -1,23 +1,21 @@
-(function (global, doc, eZ) {
+(function(global, doc, eZ) {
     const setCookie = (name, value, maxAgeDays = 356, path = '/') => {
         const maxAge = maxAgeDays * 24 * 60 * 60;
 
         doc.cookie = `${name}=${value};max-age=${maxAge};path=${path}`;
     };
     const getCookie = (name) => {
-        const cookieName = name + '=';
         const decodedCookie = decodeURIComponent(doc.cookie);
         const cookiesArray = decodedCookie.split(';');
 
-        for (index in cookiesArray) {
-            const cookieString = cookiesArray[index].trim();
+        const cookieValue = cookiesArray.find((cookie) => {
+            const cookieString = cookie.trim();
+            const seachingString = `${name}=`;
 
-            if (cookieString.indexOf(cookieName) === 0) {
-                return cookieString.split('=')[1];
-            }
-        }
+            return cookieString.indexOf(seachingString) === 0;
+        });
 
-        return '';
+        return cookieValue ? cookieValue.split('=')[1] : null;
     };
 
     eZ.addConfig('helpers.cookies', {
--- a/src/bundle/Resources/public/js/scripts/sidebar/main.menu.js
+++ b/src/bundle/Resources/public/js/scripts/sidebar/main.menu.js
@@ -1,4 +1,4 @@
-(function (global, doc, eZ, localStorage) {
+(function(global, doc, eZ, localStorage) {
     const HEADER_HEIGHT = 72;
     const RESIZER_WIDTH = 10;
     const TOGGLER_HEIGHT = 48;
--- a/src/bundle/Resources/public/js/scripts/sidebar/main.menu.js
+++ b/src/bundle/Resources/public/js/scripts/sidebar/main.menu.js
@@ -1,4 +1,4 @@
-(function(global, doc, eZ, localStorage) {
+(function (global, doc, eZ, localStorage) {
     const HEADER_HEIGHT = 72;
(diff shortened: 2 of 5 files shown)


src/bundle/Resources/public/js/scripts/helpers/cookies.helper.js
```diff
@@ -0,0 +1,33 @@
+(function (global, doc, eZ) {
+    const setCookie = (name, value, expiryDays = 356, path = '/') => {
+        const date = new Date();
+        const expiryDaysInMiliseconds = expiryDays * 24 * 60 * 60 * 1000;
+
+        date.setTime(date.getTime() + expiryDaysInMiliseconds);
+        doc.cookie = `${name}=${value};expires=${date.toUTCString()};path=${path}`;
```
dew326, line 7: Maybe it would be better to use `max-age` instead of `expires`?

src/bundle/Resources/public/js/scripts/helpers/cookies.helper.js
```diff
@@ -0,0 +1,33 @@
...
+    };
+    const getCookie = (name) => {
+        let cookieName = name + '=';
```
dew326, line 10:
```suggestion
        const cookieName = name + '=';
```

src/bundle/Resources/public/js/scripts/helpers/cookies.helper.js
```diff
@@ -0,0 +1,33 @@
...
+        let decodedCookie = decodeURIComponent(doc.cookie);
```
dew326, line 11:
```suggestion
        const decodedCookie = decodeURIComponent(doc.cookie);
```

src/bundle/Resources/public/js/scripts/helpers/cookies.helper.js
```diff
@@ -0,0 +1,33 @@
...
+        let cookiesArray = decodedCookie.split(';');
```
dew326, line 12:
```suggestion
        const cookiesArray = decodedCookie.split(';');
```

src/bundle/Resources/public/js/scripts/helpers/cookies.helper.js
```diff
@@ -0,0 +1,33 @@
...
+        let cookieValue = null;
+
+        cookiesArray.forEach((cookie) => {
+            cookie = cookie.trim();
```
dew326, line 16: here is a lot of assignment for `cookie` which is function param, I don't think this is best practice.

src/bundle/Resources/views/themes/admin/ui/menu/main_2nd_level.html.twig
```diff
@@ -1,6 +1,11 @@
 {% extends '@ezdesign/ui/menu/main_base.html.twig' %}
 
 {% block root %}
+    {% set resizerWidth = 10 %}
```
dew326, line 4: I think it should snake not camelCase in twig.

dew326 (changes requested):

src/bundle/Resources/public/js/scripts/helpers/cookies.helper.js
```diff
@@ -0,0 +1,27 @@
...
+        const maxAge = maxAgeDays * 24 * 60 * 60;
+
+        doc.cookie = `${name}=${value};max-age=${maxAge};path=${path}`;
+    };
+    const getCookie = (name) => {
+        const cookieName = name + '=';
+        const decodedCookie = decodeURIComponent(doc.cookie);
+        const cookiesArray = decodedCookie.split(';');
+
+        for (index in cookiesArray) {
```
dew326, line 12: I think `for ... of` would be better.

dew326 (changes requested):

src/bundle/Resources/public/js/scripts/helpers/cookies.helper.js
GrabowskiM: Maybe even change it all to `.find`
```
const cookieValue = cookiesArray.find((cookie) => {
  const cookieString = cookie.trim();
  
  return cookieString.indexOf(cookieName) === 0;
});

return cookieValue ? cookieValue.split('=')[1] : '';
```
or sth like this

GrabowskiM (changes requested):

src/bundle/Resources/public/js/scripts/helpers/cookies.helper.js
```diff
@@ -0,0 +1,27 @@
...
+            const cookieString = cookiesArray[index].trim();
+
+            if (cookieString.indexOf(cookieName) === 0) {
+                return cookieString.split('=')[1];
+            }
+        }
+
+        return '';
```
tischsoic, line 20: I think that no value should be represented by `null`.

src/bundle/Resources/public/js/scripts/helpers/cookies.helper.js
```diff
@@ -0,0 +1,27 @@
+(function (global, doc, eZ) {
+    const setCookie = (name, value, maxAgeDays = 356, path = '/') => {
+        const maxAge = maxAgeDays * 24 * 60 * 60;
+
+        doc.cookie = `${name}=${value};max-age=${maxAge};path=${path}`;
+    };
+    const getCookie = (name) => {
+        const cookieName = name + '=';
```
tischsoic, line 8: nitpick: this is actually not a cookie name, but cookie name with equal sign.
Personally, because it is very short and straightforward, I would prefer to remove this and just add `` `${name}=` `` inside `cookieString.indexOf`.

src/bundle/Resources/public/js/scripts/sidebar/main.menu.js
```diff
@@ -1,4 +1,4 @@
-(function(global, doc, eZ, localStorage) {
+(function (global, doc, eZ, localStorage) {
```
tischsoic, line 1:
```suggestion
(function(global, doc, eZ, localStorage) {
```

src/bundle/Resources/views/themes/admin/ui/menu/main_2nd_level.html.twig
```diff
@@ -1,23 +1,37 @@
 {% extends '@ezdesign/ui/menu/main_base.html.twig' %}
 
 {% block root %}
-    {% set extraClass = 'ibexa-main-menu__navbar--hidden' %}
+    {% set resizer_width = 10 %}
+    {% set second_level_manual_resize_min_width = 80 %}
+    {% set second_level_menu_width = app.request.cookies.get('second_menu_width') %}
+    {% set second_menu_width_style = '' %}
+    {% set second_menu_list_widt_style = '' %}
```
tischsoic, line 8: typo
```suggestion
    {% set second_menu_list_width_style = '' %}
```

tischsoic (changes requested):

dew326 (approved):

GrabowskiM (approved):

tischsoic (approved):

PR #1901, opened 2021-09-15
IBX-1013: Fixed top header and disabled animation for scrolling side menu

lucasOsti:
| Question      | Answer
| ------------- | ---
| Tickets       | https://issues.ibexa.co/browse/IBX-1013
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | no
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.content.tree.js
+++ b/src/bundle/Resources/public/js/scripts/admin.content.tree.js
@@ -94,6 +94,5 @@
     contentTreeContainer.addEventListener('transitioncancel', handleContainerTransitionStop, false);
     contentTreeContainer.addEventListener('transitionend', handleContainerTransitionStop, false);
 
-    doc.addEventListener('scroll', handleViewportChange, { capture: false, passive: true });
     window.addEventListener('resize', handleViewportChange, { capture: false, passive: true });
 })(window, window.document, window.React, window.ReactDOM, window.eZ, window.localStorage);
--- a/src/bundle/Resources/public/scss/_main-container.scss
+++ b/src/bundle/Resources/public/scss/_main-container.scss
@@ -2,13 +2,14 @@
     display: flex;
     flex-wrap: nowrap;
     overflow: hidden;
-    min-height: calc(100vh - #{calculateRem(200px)});
+    height: calc(100vh - #{calculateRem(73px)});
 
     &__content-column {
         display: flex;
         flex-direction: column;
         width: 100%;
-        overflow: hidden;
+        height: 100%;
+        overflow-y: auto;
     }
 
     &--without-anchor-menu-items {
--- a/src/bundle/Resources/public/scss/_main-menu.scss
+++ b/src/bundle/Resources/public/scss/_main-menu.scss
@@ -28,8 +28,8 @@
         }
 
         &--second-level {
+            position: relative;
             width: calculateRem(220px);
-            overflow-x: hidden;
             padding-bottom: calculateRem(32px);
 
             .ibexa-main-menu {
@@ -114,8 +114,8 @@
 
         .ibexa-popup-menu {
             width: calculateRem(250px);
-            top: calculateRem(90px);
-            left: calculateRem(120px);
+            top: calculateRem(12px);
+            left: calculateRem(48px);
         }
     }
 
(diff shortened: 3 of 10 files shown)


dew326 (approved):

src/bundle/Resources/public/js/scripts/admin.content.tree.js
```diff
@@ -94,6 +95,6 @@
     contentTreeContainer.addEventListener('transitioncancel', handleContainerTransitionStop, false);
     contentTreeContainer.addEventListener('transitionend', handleContainerTransitionStop, false);
 
-    doc.addEventListener('scroll', handleViewportChange, { capture: false, passive: true });
+    contentColumnContainer.addEventListener('scroll', handleViewportChange, { capture: false, passive: true });
```
tischsoic, line 98: Is this still needed? Won't ContentTree have the same height regardless of scroll position?

src/bundle/Resources/public/scss/_main-container.scss
```diff
@@ -2,13 +2,14 @@
     display: flex;
     flex-wrap: nowrap;
     overflow: hidden;
-    min-height: calc(100vh - #{calculateRem(200px)});
+    height: calc(100vh);
```
tischsoic, line 5:
```suggestion
    height: 100vh;
```

src/bundle/Resources/public/scss/_main-menu.scss
```diff
@@ -9,22 +9,17 @@
     }
 
     display: flex;
-    height: 100%;
-    min-height: calc(100vh - #{calculateRem(72px)});
-    background-color: $ibexa-color-black;
+    height: calc(100vh - #{calculateRem(72px)});
```
tischsoic, line 12: I just wonder: do we need this?

src/bundle/Resources/public/scss/_main-menu.scss
```diff
@@ -183,6 +160,10 @@
 
     &__item-icon-column {
         padding: calculateRem(10px);
+
+        .ibexa-icon {
+            fill: $ibexa-color-white;
+        }
```
tischsoic, line 166: Nitpick: I don't know if you like this approach, but sometimes I add `fill` to the parent and the icon just inherits it.
```suggestion
        fill: $ibexa-color-white;
```

GrabowskiM (approved):

src/bundle/Resources/public/scss/_main-menu.scss
lucasOsti: in my opinion yes because it is 100vh - the height of the header

src/bundle/Resources/public/scss/_main-menu.scss
lucasOsti: I prefer to set the style to a specific node

tischsoic (approved):

src/bundle/Resources/public/scss/_main-menu.scss
```diff
@@ -33,14 +28,13 @@
 
         &--second-level {
             width: calculateRem(220px);
-            background-color: rgba($ibexa-color-dark, 0.8);
-            border-left: calculateRem(1px) solid $ibexa-color-info-800;
-            transition: all $ibexa-admin-transition-duration;
+            overflow-x: hidden;
+            overflow-y: auto;
```
GrabowskiM, line 32: Little nitpick/useless knowledge - if one of the overflows is `hidden` and the second one is `visible` (which is default value) the second one will be automatically converted to `auto` ;)

ref: https://www.brunildo.org/test/Overflowxy2.html

GrabowskiM (approved):

src/bundle/Resources/public/scss/_main-menu.scss
tischsoic: Hmm, so basically, no need to state `overflow-y: auto;` explicitly?

tischsoic (approved):

GrabowskiM (approved):

src/bundle/Resources/public/scss/_main-menu.scss
dew326: Why not use `ibexa-icon--light`?

dew326 (approved):

piotrbisaga (approved):
QA approved.
Checked on Ibexa Commerce 4.0.x-dev on Chrome & Firefox

PR #1903, opened 2021-09-16
IBX-940: Dropdown in tab is partly hidden

GrabowskiM:
| Question      | Answer
| ------------- | ---
| Tickets       | https://issues.ibexa.co/browse/IBX-940
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

https://github.com/ezsystems/ezcommerce-shop/pull/350
https://github.com/ezsystems/ezcommerce-admin-ui/pull/107
https://github.com/ezsystems/ezplatform-form-builder/pull/306

#### Checklist:
- [x] Ready for Code Review


src/bundle/Resources/public/js/scripts/core/dropdown.js
```diff
@@ -1,94 +1,86 @@
...
-                `<li
-                class="ibexa-dropdown__selected-item"
-                data-value="{{value}}">
-                    {{label}}<span class="${CLASS_REMOVE_SELECTION}"></span>
-            </li>`;
-            this.canSelectOnlyOne = !sourceInput?.multiple;
+            this.container = config.container.classList.contains('ibexa-dropdown') ? config.container : config.container.querySelector('.ibexa-dropdown');
+
+            if (!this.container) {
+                this.error = true;
```
tischsoic, line 12: I don't think it's needed, given we throw the exception below.
It's only used in one place, in the `init` function, to check if the error occurred, but `init` should not be called after the throw in the first place. If someone catches the error and still calls `init` it's 100% his fault.

src/bundle/Resources/public/scss/_dropdown.scss
```diff
@@ -338,4 +357,4 @@
     .ibexa-dropdown {
         margin: calculateRem(5px) calculateRem(8px);
     }
-}
+}
```
tischsoic, line 360: newline

src/bundle/Resources/public/js/scripts/core/dropdown.js
```diff
@@ -151,52 +143,82 @@
...
+            this.container.classList.add('ibexa-dropdown--is-expanding');
+            this.itemsFilterInput
+                .closest('.ibexa-input-text-wrapper')
+                .querySelector('.ibexa-input-text-wrapper__action-btn--clear')
+                .click();
+            this.itemsFilterInput.focus();
+
+            setTimeout(() => {
+                this.container.classList.add('ibexa-dropdown--is-expanded');
+            }, 1);
```
tischsoic, line 196: Is this solution robust? Do we need `setTimeout`?

src/bundle/Resources/public/js/scripts/core/dropdown.js
```diff
@@ -151,52 +143,82 @@
...
+                this.container.classList.remove('ibexa-dropdown--is-expanded');
+                this.hideOptions();
+                this.fireValueChangedEvent();
+            }, 1);
+        }
+
+        expandDropdown() {
+            const viewportHeight = window.innerHeight || document.documentElement.clientHeight;
+            const { left, top, bottom } = this.selectedItemsContainer.getBoundingClientRect();
+            const isItemsContainerAbove = bottom + ITEMS_LIST_MIN_HEIGHT > viewportHeight;
```
tischsoic, line 177: Tiny nitpick: in this place, items container is not above, because it has not been expanded yet.

src/bundle/Resources/views/themes/admin/ui/component/dropdown.html.twig
```diff
@@ -9,14 +9,20 @@
         {{ source|default(null)|raw }}
     </div>
     <div class="ibexa-dropdown__wrapper {{ class|default('') }}">
-        <ul class="ibexa-dropdown__selection-info">
+        <ul
+            class="ibexa-dropdown__selection-info"
+            data-template="{{ include('@ezdesign/ui/component/dropdown_selected_item.html.twig', {
+                value: '{{ value }}',
+                label: '{{ label }}',
+            })|e }}"
```
tischsoic, line 17: not sure, but:
```suggestion
            })|e('html_attr') }}"
```

tischsoic (changes requested):

lucasOsti (approved):

src/bundle/Resources/public/js/scripts/core/dropdown.js
GrabowskiM: Yes, as we need first to add class that will change display to block, we can't run both changing display to block and changing scale to 1 in same frame, as in such case transition won't be visible

dew326 (approved):

tischsoic (approved):

bogusez (approved):

PR #1911, opened 2021-09-20
IBX-964: Added offset to timestamp

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-964](https://issues.ibexa.co/browse/IBX-964)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   |/no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
@@ -73,13 +73,9 @@
             return;
         }
 
-        const selectedDateWithUserTimezone = eZ.helpers.timezone.convertDateToTimezone(date[0]);
+        const selectedDateWithUserTimezone = eZ.helpers.timezone.convertDateToTimezone(date[0], eZ.adminUiConfig.timezone, true);
 
-        date = new Date(date[0]);
-        date = new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()));
-        const milliseconds = date.getTime() - selectedDateWithUserTimezone.utcOffset() * 60 * 1000;
-
-        sourceInput.value = Math.floor(milliseconds / 1000);
+        sourceInput.value = Math.floor(selectedDateWithUserTimezone.valueOf() / 1000);
         sourceInput.dispatchEvent(event);
     };
     const clearValue = (sourceInput, flatpickrInstance, event) => {
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
@@ -73,10 +73,13 @@
             return;
         }
 
+        const selectedDateWithUserTimezone = eZ.helpers.timezone.convertDateToTimezone(date[0]);
+
         date = new Date(date[0]);
         date = new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()));
+        const microTime = date.getTime() - (selectedDateWithUserTimezone.utcOffset() * 60 * 1000);
 
-        sourceInput.value = Math.floor(date.getTime() / 1000);
+        sourceInput.value = Math.floor(microTime / 1000);
         sourceInput.dispatchEvent(event);
     };
     const clearValue = (sourceInput, flatpickrInstance, event) => {
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
@@ -77,9 +77,9 @@
 
         date = new Date(date[0]);
         date = new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()));
-        const microTime = date.getTime() - selectedDateWithUserTimezone.utcOffset() * 60 * 1000;
+        const milliseconds = date.getTime() - selectedDateWithUserTimezone.utcOffset() * 60 * 1000;
 
-        sourceInput.value = Math.floor(microTime / 1000);
+        sourceInput.value = Math.floor(milliseconds / 1000);
         sourceInput.dispatchEvent(event);
     };
     const clearValue = (sourceInput, flatpickrInstance, event) => {
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
@@ -77,7 +77,7 @@
 
         date = new Date(date[0]);
         date = new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()));
-        const microTime = date.getTime() - (selectedDateWithUserTimezone.utcOffset() * 60 * 1000);
+        const microTime = date.getTime() - selectedDateWithUserTimezone.utcOffset() * 60 * 1000;
 
         sourceInput.value = Math.floor(microTime / 1000);
         sourceInput.dispatchEvent(event);


src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
```diff
@@ -73,10 +73,13 @@
             return;
         }
 
+        const selectedDateWithUserTimezone = eZ.helpers.timezone.convertDateToTimezone(date[0]);
+
         date = new Date(date[0]);
         date = new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()));
+        const microTime = date.getTime() - (selectedDateWithUserTimezone.utcOffset() * 60 * 1000);
```
Steveb-p, line 80: Operator precedence makes it unecessary.
```suggestion
        const microTime = date.getTime() - selectedDateWithUserTimezone.utcOffset() * 60 * 1000;
```

src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
```diff
@@ -73,10 +73,13 @@
             return;
         }
 
+        const selectedDateWithUserTimezone = eZ.helpers.timezone.convertDateToTimezone(date[0]);
+
         date = new Date(date[0]);
         date = new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()));
+        const microTime = date.getTime() - selectedDateWithUserTimezone.utcOffset() * 60 * 1000;
```
dew326, line 80: Im not sure if this is a micro time. This is with the precision of milliseconds and microtime is smaller?

mateuszdebinski: yes, you are right, I changed to milliseconds

lucasOsti: Aggred with @dew326, function getTime() returns milliseconds

lucasOsti (approved):

dew326 (approved):

mateuszbieniek (approved):

konradoboza (approved):

GrabowskiM (approved):

src/bundle/Resources/public/js/scripts/fieldType/ezdate.js
```diff
@@ -73,10 +73,13 @@
             return;
         }
 
+        const selectedDateWithUserTimezone = eZ.helpers.timezone.convertDateToTimezone(date[0]);
```
tischsoic, line 76: Just a question: what about passing `userTimezone` to convertDateToTimezone, like in `ezdatetime`?
https://github.com/ezsystems/ezplatform-admin-ui/blob/329b69b980572ab96a726223901fba0094f9bf0b/src/bundle/Resources/public/js/scripts/fieldType/ezdatetime.js#L81
this won't work?

mateuszdebinski: yes, it could be changed as it is in ezdatetime, i push changed, now

tischsoic: If it works, then great!

tischsoic (approved):

bogusez (approved):

dew326:
@mateuszdebinski please merge up.

DoNotToyWithMe:
Hi, the provided solution doesn't fix the issue.

My PHP timezone is America/Sao_Paulo

I created an article, selected September 23, 2021.

Published.

Article preview shows September 23, 2021.

Edited the Article again and now the selected date is  September 22, 2021.

Could you please check the original ticket description where I pointed the changes in IBX-300 caused the issue? When we reverted the changes in:

src/bundle/Resources/public/js/scripts/fieldType/ezdate.js

src/bundle/Resources/views/themes/admin/ui/field_type/preview/content_fields.html.twig

It fixed completely.

DoNotToyWithMe:
I created a new issue https://issues.ibexa.co/browse/IBX-1090

mateuszdebinski:
@DoNotToyWithMe in user settings you must have the same time zone as in php.ini settings, otherwise, there will be discrepancies

DoNotToyWithMe:
@mateuszbieniek originally the date attribute would be set to UTC always, this is why the original code in vendor/ezsystems/ezplatform-admin-ui/src/bundle/Resources/views/themes/admin/ui/field_type/preview/content_fields.html.twig

Converts the  date in the preview template to UTC.

DoNotToyWithMe:
Check the problematic commits:

https://github.com/ezsystems/ezplatform-admin-ui/commit/08ea4f3ff8715baaaa23357e2e159632bbc389c8#diff-03a88b29d0af2169f521b6db3be86937e4791cfa8d844d022acd2a3646927d96

https://github.com/ezsystems/ezplatform-kernel/commit/da07b7825ec67949e8ed2523ca6532976c403e84#diff-4477c92ab138d7846a862f46780dc60c00c2cbb737b6fc3fe1b0b4479f836e05

In eZ/Publish/Core/FieldType/Date/Value.php for example

return new static(new DateTime("@{$timestamp}"))

Was converting the value to UTC, that is expected. Then the preview template does the same, it should be UTC everywhere.

mateuszbieniek:
> @mateuszbieniek originally the date attribute would be set to UTC always, this is why the original code in vendor/ezsystems/ezplatform-admin-ui/src/bundle/Resources/views/themes/admin/ui/field_type/preview/content_fields.html.twig
> 
> Converts the date in the preview template to UTC. Also the js code should do the same.

I think you pinged the wrong Mateusz :slightly_smiling_face:

Steveb-p:
@DoNotToyWithMe unfortunately the problem is more complex than just simply "converting the value to UTC".

The fact is that `new DateTime('YYYY-MM-DD')` and `new DateTime("@{$timestamp}")` generate objects with different timezones. This means that when calling `DateTime::format` you'll get two different outcomes, even if UTC timestamp is the same.

For this reason PHP code (mentioned by you: https://github.com/ezsystems/ezplatform-kernel/commit/da07b7825ec67949e8ed2523ca6532976c403e84#diff-4477c92ab138d7846a862f46780dc60c00c2cbb737b6fc3fe1b0b4479f836e05) was adjusted to always generate objects with the same `DateTimeZone`. This is done so that every time a `DateTime` object is created, it's default time zone for presentation is the same (matching php.ini settings).

Unfortunate side effect of this is that we require an upgrade for users that are on `2.5` version, because at some point in time a frontend solution was introduced that **added an offset to timestamp that was sent to the server**. This means that values that were presented as timestamps were not, in fact, UTC timestamps - **they were UTC timestamps + an offset**.

Therefore, the reason why `DateTime` objects passed into twig templates needed timezone to be specified as `UTC` was exactly that they already contain (incorrectly) a time zone offset. This is clearly wrong and makes it impossible for sites that actually span multiple timezones (and allow users to switch to their own timezone, for example) to work correctly - they will not be able to properly adjust the displayed time. To do so, they would first need to extract the offset and then specify the request timezone.

Unfortunately this means that there might be places in frontend code that add the offset on their own still. I'm pretty sure we haven't located all of these.

A workaround for that is to switch the default timezone for the application (via `date_default_timezone_set` or in php.ini) to UTC. I'm sorry that there is no complete solution yet.

DoNotToyWithMe:
> The fact is that new DateTime('YYYY-MM-DD') and new DateTime("@{$timestamp}") generate objects with different timezones. This means that when calling DateTime::format you'll get two different outcomes, even if UTC timestamp is the same.

The idea is that the $timestamp variable passed to DateTime would be the Date unix timestamp set to be in UTC. DateTime created with that @ has the timezone set to UTC, so format is always going to be the day in UTC. In the frontend code always take that into account, we all now DateTime('YYYY-MM-DD') will create the Date in the current timezone.

The proposed workaround is worse than what we did in one of our projects where we reverted the changes in IBX-300.

DoNotToyWithMe:
@Steveb-p To complement my answer, check the description of:

https://github.com/ezsystems/ezpublish-legacy/commit/3ce2994e031c69e86f7608868d1fba6457c7b05b

EZP-29279: Values of Date Field Type should be handled in UTC only (#…
…1401)

* EZP-29279: ezdate timestamps are now stored in UTC

* Added missing hours, minutes and seconds to Timestamp conversion and add more s p a c e s to the same class code

* Fixup after CR

Steveb-p:
@DoNotToyWithMe while I agree that the current state is unacceptable, I still stand with the notion that the solution presented in IBX-300 is the right one going forward.

### tl;dr It's best to not concern ourselves with timezones at all if possible.

It goes against the previous solution precisely because it was incorrect and introduced the mess we are in right now.

If you look into Symfony code you'll see that when dealing with DateTime forms, they use current server timezone setting by default (even for timestamps!):
https://github.com/symfony/symfony/blob/f32af4683a04f440cda7ad8068d6cd05600d5e0e/src/Symfony/Component/Form/Extension/Core/DataTransformer/BaseDateTimeTransformer.php#L39-L40
https://github.com/symfony/symfony/blob/f32af4683a04f440cda7ad8068d6cd05600d5e0e/src/Symfony/Component/Form/Extension/Core/DataTransformer/DateTimeToTimestampTransformer.php#L67-L73

This makes all Date objects in PHP consistent (unless you actually do work with timezones). Twig templating code in this case may not care about timezones because every object it gets uses the same one.
We do not need to be timezone-aware in Twig precisely because it makes it harder to work with. In case of each Date object you need to control it's source and remember to use or not use a specific timezone declaration. That's error prone and part of the current problem.

### Now, the root of the problem.

**In some instances when creating or updating dates the fact that we were dealing with UTC-based DateTime object was neglected**, which lead to date in question being offset by the difference between server default timezone and UTC. Solutions for this varied - some people adjusted the displayed timezone (that's the `| date('UTC')` approach) while others **added timezone offset to the timestamp**.

While the first option was inconsequential (just date discrepancy if there were other forgotten places), the second one made this a lot worse.

That's because now some timestamps in database have an offset (which they should never have!) while others don't.

That's why, unless your application actually deals with different timezones, DateTime objects should always use the same timezone. This makes it so you don't need to think about timezone conversion anywhere, and that's what IBX-300 aims for.

PR #1930, opened 2021-09-27
IBX-1054: Fixed navigation font size

dew326:
| Question      | Answer
| ------------- | ---
| Tickets       | https://issues.ibexa.co/browse/IBX-1054
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/core/_custom.tooltip.scss
+++ b/src/bundle/Resources/public/scss/core/_custom.tooltip.scss
@@ -91,6 +91,7 @@
                 color: $ibexa-color-dark;
                 background-color: $ibexa-color-info;
                 border-color: transparent;
+                box-shadow: 0 calculateRem(4px) calculateRem(6px) rgba(65, 145, 255, 0.1);
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/public/scss/core/_custom.tooltip.scss
```diff
@@ -91,6 +91,7 @@
                 color: $ibexa-color-dark;
                 background-color: $ibexa-color-info;
                 border-color: transparent;
+                box-shadow: 0 calculateRem(4px) calculateRem(6px) rgba(65, 145, 255, 0.1);
```
GrabowskiM, line 94: I don't think that we want to use this kind of rgba, we should use one of our palette colors (that's `$ibexa-color-info` )

GrabowskiM (changes requested):

GrabowskiM (approved):

lucasOsti (approved):

PR #1934, opened 2021-09-28
[Behat] Removed logic based on exceptions in Table searching

mnocon:
This is a companion PR for https://github.com/ezsystems/BehatBundle/pull/212

Relying on exceptions to control flow of the application is a bad practice, in this case it stops us from using interactive debug in places where this code is used (it's detected as an error and interactive session kicks in).

This code works the same way, but does not rely on exceptions.

#### Checklist:
- [x] Ready for Code Review


bogusez (approved):

tomaszszopinski (approved):

micszo (approved):

Nattfarinn (approved):

Nattfarinn:
@mnocon Please, merge up to `master` and then make a merge up PR to `ibexa/main`.

PR #1935, opened 2021-09-28
Fixed content always being translated using a base translation

mateuszbieniek:
| Question      | Answer
| ------------- | ---
| Tickets       | https://issues.ibexa.co/browse/IBX-1079
| Bug fix?      | yes

Due to a `placeholder` setting set to `false` for `base_language` ChoiceType, `<select>` field with Languages available as a Base Translation always used the first option on the list, when submitted empty.

Changing `placeholder` to `true` resolves the "HTML" issue, but uncovers some underlying issues:
- When BaseLanguage is `null`, `ibexa.content.translate_with_location.proxy` route generates without the last `fromLanguageCode` part, and URL is matched against `ibexa.content.translate_with_location` with exception "Can't find Language `proxy`". This PR fixes this by changing order in `routing.yaml` so in this case `ibexa.content.translate_with_location.proxy` takes precedense.
- `ContentProxyCreateDraftListener:translate` always sets field values for the struct, therefore even when `fromLanguage` is set to `null`, main language values are used. This is fixed by adding `if ($fromLanguageCode !== null)` check before setting fields in the Struct.

In addition to the original issue, the same fix for translating Content Types is used.

### DIFF
--- a/src/bundle/Resources/config/routing.yaml
+++ b/src/bundle/Resources/config/routing.yaml
@@ -548,20 +548,6 @@ ezplatform.content.preview:
         languageCode: ~
         locationId: ~
 
-ezplatform.content.translate:
-    path: /content/{contentId}/translate/{toLanguageCode}/{fromLanguageCode}
-    methods: ['GET', 'POST']
-    defaults:
-        _controller: 'EzSystems\EzPlatformAdminUiBundle\Controller\ContentEditController::translateAction'
-        fromLanguageCode: ~
-
-ibexa.content.translate_with_location:
-    path: /content/{contentId}/location/{locationId}/translate/{toLanguageCode}/{fromLanguageCode}
-    methods: ['GET', 'POST']
-    defaults:
-        _controller: 'EzSystems\EzPlatformAdminUiBundle\Controller\ContentEditController::translateAction'
-        fromLanguageCode: ~
-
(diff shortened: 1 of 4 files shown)


mikadamczyk (changes requested):
Maybe a good idea will be to add a comment in `routing.yaml` that we are relying on the order of routings. Or, we could exclude "proxy" as a language in requirements. And for 4.0 we should provide a better solution to not rely on routes order or requirements.

mikadamczyk (approved):

Steveb-p (approved):
> And for 4.0 we should provide a better solution to not rely on routes order or requirements.

I second to that.

adamwojs (approved):

adamwojs:
@mateuszbieniek You can move this PR to QA

tomaszszopinski (approved):
QA approved on IbexaDXP 3.3.13-dev.

mateuszbieniek:
pinging @mikadamczyk :)

mikadamczyk:
@mateuszbieniek you can merge up and to ibexa

mateuszbieniek:
merged to ibexa: https://github.com/ibexa/admin-ui/commit/6f36566da0f66bad1657d71660319bcc95b8a6a3

PR #1993, opened 2021-10-28
IBX-1306: Personalization dashboard fixes

tischsoic:
JIRA: https://issues.ibexa.co/browse/IBX-1306


src/bundle/Resources/public/js/scripts/core/base.chart.js
```diff
@@ -42,9 +42,6 @@
         setOptions(options) {
             this.options = {
                 ...defaultOptions,
-                animation: {
-                    onComplete: this.onCompleteAnimationCallback,
```
tischsoic, line 46: Note: it did not work as expected as it was also triggered when animating these tooltips:
Screenshot 2021-11-02 at 11 54 42
It was deleted as there is no need for it.

src/bundle/Resources/public/js/scripts/core/dropdown.js
```diff
@@ -39,8 +39,8 @@
             this.itemsListContainer = this.itemsContainer.querySelector('.ibexa-dropdown__items-list');
             this.itemsFilterInput = this.itemsContainer.querySelector('.ibexa-dropdown__items-filter');
 
-            this.hasDefaultSelection = config.hasDefaultSelection ?? false;
             this.canSelectOnlyOne = !this.sourceInput?.multiple;
+            this.hasDefaultSelection = config.hasDefaultSelection ?? this.canSelectOnlyOne;
```
GrabowskiM, line 43: Here I would put `||` instead of `??`, imo `??` should be used only when second argument is constant

src/bundle/Resources/public/scss/_chart.scss
```diff
@@ -0,0 +1,93 @@
+.ibexa-chart {
+    display: flex;
+    flex-direction: column;
+    height: 100%;
+    max-height: calculateRem(1000px);
+    border: calculateRem(1px) solid $ibexa-color-light;
+    border-radius: $ibexa-border-radius;
+    letter-spacing: calculateRem(0.12px);
```
GrabowskiM, line 8: There's letter-spacing on whole html tag, so it's unnecessary here

src/bundle/Resources/public/scss/_chart.scss
```diff
@@ -0,0 +1,93 @@
...
+    &__time-range,
+    &__time-period {
+        color: $ibexa-color-dark-500;
+        font-size: $ibexa-text-font-size-medium;
+    }
+
+    &__message {
+        position: absolute;
+        z-index: 1;
+        top: calc(50% - 150px);
```
GrabowskiM, line 41: calculateRem

GrabowskiM (changes requested):

lucasOsti (approved):

src/bundle/Resources/public/js/scripts/core/dropdown.js
tischsoic: I wouldn't subject the use of `??` to the type of argument used after it. Whether the last argument is constant or not it does not affect `??` behaviour.

When it comes to discussing behaviour, in your case, given that `config.hasDefaultSelection` was set to `false` it would default to `this.canSelectOnlyOne`. Do we want such behaviour?

dew326: I think the logic here is correct.
Not set -> default
set false -> false

dew326 (approved):

GrabowskiM (approved):

GrabowskiM (approved):

lucasOsti (approved):

PR #2003, opened 2021-11-09
IBX-1015: Added support for subitems grouping in admin-ui menu

damianz5:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-1015](https://issues.ibexa.co/browse/IBX-1015)
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | yes

As a developer I want to group submenu items and change order of items in groups

Screenshot 2021-11-16 at 14 44 39

Screenshot 2021-11-17 at 21 02 32

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/scss/_main-menu.scss
+++ b/src/bundle/Resources/public/scss/_main-menu.scss
@@ -79,7 +79,7 @@
                 }
 
                 .ibexa-main-menu__item {
-                    &--group-name:before {
+                    &--group-name {
                         width: 0;
                     }
                 }
@@ -132,14 +132,7 @@
         &--group-name {
             color: $ibexa-color-light-700;
             font-size: $ibexa-text-font-size-small;
-
-            &:before {
-                content: "";
-                border-top: 1px solid $ibexa-color-info-800;
-                transition: width 0.30s linear;
-                width: 100%;
-                display: block;
-            }
+            border-top: calculateRem(1px) solid $ibexa-color-info-800;
         }
     }
 
--- a/src/bundle/Resources/public/scss/_main-menu.scss
+++ b/src/bundle/Resources/public/scss/_main-menu.scss
@@ -43,6 +43,11 @@
 
                 &__item-text-column {
                     padding: calculateRem(10px);
+
+                    &--group {
+                        color: $ibexa-color-light-700;
+                        font-size: $ibexa-text-font-size-small;
+                    }
                 }
             }
 
@@ -70,6 +75,12 @@
 
                     &__toggler {
                         margin-right: calculateRem(6px);
+                    }
+                }
+
+                .ibexa-main-menu__item {
+                    &--group-name:before {
+                        width: 0;
                     }
                 }
             }
@@ -117,6 +128,19 @@
             top: calculateRem(12px);
             left: calculateRem(48px);
         }
+
+        &--group-name {
+            color: $ibexa-color-light-700;
+            font-size: $ibexa-text-font-size-small;
+
+            &:before {
+                content: "";
+                border-top: 1px solid $ibexa-color-info-800;
+                transition: width 0.30s linear;
+                width: 100%;
+                display: block;
+            }
+        }
     }
 
     &__item-action {
@@ -142,6 +166,10 @@
 
                 &__item-text-column {
                     color: $ibexa-color-info;
+
+                    &--group {
+                        color: $ibexa-color-light-700;
+                    }
                 }
             }
         }
@@ -189,10 +217,10 @@
 
         &--bottom {
             width: calc(100% - #{calculateRem(24px)});
-            margin: 0 auto;          
+            margin: 0 auto;
         }
     }
-    
+
     &__resizer {
         position: absolute;
         top: 0;
--- a/src/bundle/Resources/public/scss/_popup-menu.scss
+++ b/src/bundle/Resources/public/scss/_popup-menu.scss
@@ -21,6 +21,19 @@
         &--hidden {
             display: none;
         }
+
+        &--group-name {
+            flex-direction: column;
+            height: 100%;
+            align-items: initial;
+
+            &:before {
+                content: "";
+                border-top: 1px solid $ibexa-color-info-800;
+                width: 100%;
+                display: block;
+            }
+        }
     }
 
     .ibexa-popup-menu__item-content {
@@ -46,6 +59,21 @@
             opacity: 0.2;
             cursor: default;
         }
+
+        &--group-name {
+            color: $ibexa-color-light-700;
+            font-size: $ibexa-text-font-size-small;
+            cursor: initial;
+
+            &:hover {
+                background-color: initial;
+                color: $ibexa-color-light-700;
+            }
+        }
+    }
+
+    .ibexa-popup-menu__items-list {
+        padding-left: 0;
     }
 
     &--info-neon {
--- a/src/bundle/Resources/public/scss/_popup-menu.scss
+++ b/src/bundle/Resources/public/scss/_popup-menu.scss
@@ -26,13 +26,7 @@
             flex-direction: column;
             height: 100%;
             align-items: initial;
-
-            &:before {
-                content: "";
-                border-top: 1px solid $ibexa-color-info-800;
-                width: 100%;
-                display: block;
-            }
+            border-top: calculateRem(1px) solid $ibexa-color-info-800;
         }
     }
 
--- a/src/bundle/Resources/views/themes/admin/ui/menu/main_2nd_level.html.twig
+++ b/src/bundle/Resources/views/themes/admin/ui/menu/main_2nd_level.html.twig
@@ -42,11 +42,13 @@
 
                             <ul id="{{ child.name }}-popup" class="ibexa-popup-menu ibexa-popup-menu--info-neon ibexa-popup-menu--hidden">
                                 {% for item in child.children %}
+                                    {% set renderItemAsGroup = item.children|length > 0 %}
                                     {{ block('popupItem') }}
                                 {% endfor %}
                             </ul>
                         </li>
                         {% for item in child.children %}
+                            {% set renderItemAsGroup = item.children|length > 0 %}
                             {{ block('item') }}
                         {% endfor %}
                     </ul>
@@ -88,9 +90,15 @@
 
     {% set link_attributes = item.linkAttributes|merge(linkAttributes) %}
 
-    <a{{ knp_menu.attributes(link_attributes) }}>
-        <div class="ibexa-main-menu__item-text-column">{{ block('label') }}</div>
-    </a>
+    {% if renderItemAsGroup|default() %}
+        <div {{ knp_menu.attributes(link_attributes) }}>
+            <span class="ibexa-main-menu__item-text-column ibexa-main-menu__item-text-column--group">{{ block('label') }}</span>
+        </div>
+    {% else %}
+        <a {{ knp_menu.attributes(link_attributes) }}>
+            <span class="ibexa-main-menu__item-text-column">{{ block('label') }}</span>
+        </a>
+    {% endif %}
 {% endblock %}
 
 {% block spanElement %}
--- a/src/bundle/Resources/views/themes/admin/ui/menu/main_base.html.twig
+++ b/src/bundle/Resources/views/themes/admin/ui/menu/main_base.html.twig
@@ -3,8 +3,13 @@
 
 {% block popupItem %}
     {% if item.displayed %}
+        {% set classes = [
+            'ibexa-popup-menu__item',
+            (renderItemAsGroup|default() ? 'ibexa-popup-menu__item--group-name')
+        ] %}
+
         {% set node_attributes = {
-            'class': 'ibexa-popup-menu__item',
+            'class': classes|join(' ')
         } %}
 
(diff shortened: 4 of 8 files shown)


src/bundle/Resources/config/services/menu.yaml
```diff
@@ -174,6 +181,11 @@ services:
         tags:
             - { name: kernel.event_listener, event: ezplatform_admin_ui.menu_configure.main_menu, method: moveAdminToLast, priority: -50 }
 
+    # regroup menu listener
```
mikadamczyk, line 184:
```suggestion
```

Steveb-p:
Any particular reason why the normal KNP Menu nesting was not used here?

damianz5:
BC mostly and current 2nd_level structure

src/lib/Menu/Admin/GroupManipulator.php
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Sorts children based on their group, orderNumber and position.
+     *
+     * @param string $parentName
+     * @param \Knp\Menu\ItemInterface[] $children
+     */
+    private function sort(string $parentName, array &$children): void
+    {
+        $c = [];
```
mikadamczyk, line 57: Could you add a more descriptive variable name here?

src/lib/Menu/Admin/GroupManipulator.php
```diff
@@ -0,0 +1,77 @@
...
+
+            // check if group was registered for specified parent
+            if (!array_key_exists($groupName, $this->groupsOrderByParentItem[$parentName])) {
+                $child->setExtra('group', null);
+                $groupName = self::DEFAULT_GROUP_KEY;
+            }
+
+            $groupOrder = $this->groupsOrderByParentItem[$parentName][$groupName] ?? self::DEFAULT_GROUP_ORDER;
+            $itemOrder = $child->getExtra('orderNumber', 0);
+            $c[$name] = ['g' => $groupOrder, 'o' => $itemOrder, 'i' => $i++];
```
mikadamczyk, line 70: Same for array indexes

mikadamczyk (commented):
Could you provide test coverage for new classes?

Steveb-p (commented):
Personally I don't think it's the right direction - KNP Menu already provides a menu structure with subelements. This not only provides a custom grouping based on arbitrary item attribute (circumventing any features that KNP Menu already has, like detecting that a child is `active` and parent should be "an active ancestor").

This is added code that only increases the maintenance. I fail to see why normal structure could not be used.

alongosz (commented):
Other Reviewers have good points, I don't see anything more code-wise.

The remark I would have is that your PR title and commit message should be prefixed with IBX jira issue, followed by colon. Otherwise it won't appear in the autogenerated change log. Not to mention that commits should be formed in past simple ;)

src/bundle/Resources/config/services/menu.yaml
damianz5: No need for this class in reorganized code

damianz5:
Thank you for all reviews, the code has been reorganized and now uses the KNP Menu nesting. Please review again

src/bundle/Resources/views/themes/admin/ui/menu/main_2nd_level.html.twig
```diff
@@ -87,9 +98,10 @@
     {% endif %}
 
     {% set link_attributes = item.linkAttributes|merge(linkAttributes) %}
+    {% set group_name_class = renderItemAsGroup is defined and renderItemAsGroup ? 'ibexa-main-menu__item-text-group' %}
```
dew326, line 101: This class should be something like `ibexa-main-menu__item-text-column--group`

src/bundle/Resources/views/themes/admin/ui/menu/main_2nd_level.html.twig
```diff
@@ -87,9 +98,10 @@
...
 
     <a{{ knp_menu.attributes(link_attributes) }}>
```
dew326, line 103: If this will be a group label should this be still an `<a>` tag? I think it shouldn't be a link

dew326 (changes requested):

src/lib/Menu/MainMenuBuilder.php
```diff
@@ -247,7 +249,7 @@ public function createStructure(array $options): ItemInterface
         $contentMenuItems = $this->getContentMenuItems();
         $adminMenuItems = $this->getAdminMenuItems();
 
-        $menu[self::ITEM_CONTENT]->setChildren($contentMenuItems);
+        array_map([$menu[self::ITEM_CONTENT], 'addChild'], $contentMenuItems);
```
Steveb-p, line 252: Personally I'd pass the content menu object into `getContentMenuItems()` method, and `addChild` there. But that's my personal preference.

Steveb-p (approved):

src/bundle/Resources/views/themes/admin/ui/menu/main_2nd_level.html.twig
Steveb-p: It can still lead to a valid route, although it usually doesn't (otherwise it's difficult to create a proper accordion with open/close).

IMO do not render it as an `<a>` element, regardless of route - at least for now.

adamwojs:
Code style needs to be fixed here

src/lib/Menu/MainMenuBuilder.php
```diff
@@ -175,27 +176,28 @@ public function createStructure(array $options): ItemInterface
         /** @var \Knp\Menu\ItemInterface|\Knp\Menu\ItemInterface[] $menu */
         $menu = $this->factory->createItem('root');
 
-        $menu->addChild($this->factory->createItem(self::ITEM_DASHBOARD, [
-                'route' => 'ezplatform.dashboard',
+        $menu->addChild($this->factory->createItem(self::ITEM_CONTENT, [
```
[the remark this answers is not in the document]
damianz5: +1 for suggested change

for side note: those are different factories `\Knp\Menu\MenuFactory` and `\EzSystems\EzPlatformAdminUi\Menu\MenuItemFactory`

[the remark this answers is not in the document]
Steveb-p: @damianz5 makes sense.

The issue of using wrong factory when using `addChild()` without an already constructed `Item` can be circumvented by slightly changing `EzSystems\EzPlatformAdminUi\Menu\MenuItemFactory::createItem`:
```diff
- return $this->factory->createItem($name, array_merge_recursive($defaults, $options));
+ $item = $this->factory->createItem($name, array_merge_recursive($defaults, $options));
+ $item->setFactory($this);
+
+ return $item;
```

This would ensure that when calling `$item->addChild()` our implementation of the MenuFactory is called, not the default one.

src/lib/Menu/MainMenuBuilder.php
```diff
@@ -173,95 +174,85 @@ public function createStructure(array $options): ItemInterface
...
-                ],
-                'extras' => [
-                    'separate' => true,
-                    'bottom_item' => true,
-                    'icon' => 'settings-block',
-                    'orderNumber' => 140,
-                ],
-            ]
-        ));
+        $menu->addChild($adminMenu = $this->createMenuItem(self::ITEM_ADMIN, [
```
Steveb-p, line 202: `ItemInterface::addChild()` returns the added child. You can assign from the returned value, instead of inlining the assignment (which is usually frowned upon, because it's not "immediately" visible)

```suggestion
        $adminMenu = $menu->addChild($this->createMenuItem(self::ITEM_ADMIN, [
```

src/bundle/Resources/views/themes/admin/ui/menu/main_2nd_level.html.twig
```diff
@@ -88,9 +99,15 @@
...
 
-    <a{{ knp_menu.attributes(link_attributes) }}>
-        <div class="ibexa-main-menu__item-text-column">{{ block('label') }}</div>
-    </a>
+    {% if renderItemAsGroup is defined and renderItemAsGroup %}
+        <span{{ knp_menu.attributes(link_attributes) }}>
+            <div class="ibexa-main-menu__item-text-column ibexa-main-menu__item-text-column--group">{{ block('label') }}</div>
+        </span>
+    {% else %}
+        <a{{ knp_menu.attributes(link_attributes) }}>
```
adamwojs, line 107:
```suggestion
        <a {{ knp_menu.attributes(link_attributes) }}>
```

src/bundle/Resources/views/themes/admin/ui/menu/main_2nd_level.html.twig
```diff
@@ -88,9 +99,15 @@
 
     {% set link_attributes = item.linkAttributes|merge(linkAttributes) %}
 
-    <a{{ knp_menu.attributes(link_attributes) }}>
-        <div class="ibexa-main-menu__item-text-column">{{ block('label') }}</div>
-    </a>
+    {% if renderItemAsGroup is defined and renderItemAsGroup %}
+        <span{{ knp_menu.attributes(link_attributes) }}>
```
adamwojs, line 103:
```suggestion
        <span {{ knp_menu.attributes(link_attributes) }}>
```

adamwojs (approved):

Steveb-p (approved):

src/bundle/Resources/views/themes/admin/ui/menu/main_2nd_level.html.twig
```diff
@@ -88,9 +92,15 @@
 
     {% set link_attributes = item.linkAttributes|merge(linkAttributes) %}
 
-    <a{{ knp_menu.attributes(link_attributes) }}>
-        <div class="ibexa-main-menu__item-text-column">{{ block('label') }}</div>
-    </a>
+    {% if renderItemAsGroup is defined and renderItemAsGroup %}
```
mikadamczyk, line 95:
```suggestion
    {% if renderItemAsGroup|default %} 
```
the value will default to an empty string which is evaluated as `false`

mikadamczyk (approved):

src/bundle/Resources/views/themes/admin/ui/menu/main_base.html.twig
```diff
@@ -64,11 +82,20 @@
...
         {%- endif -%}
 
-        <li{{ knp_menu.attributes(attributes) }}>
+        <li {{ knp_menu.attributes(attributes) }}>
             {%- if item.uri is not empty and (not matcher.isCurrent(item) or options.currentAsLink) %}
                 {{ block('linkElement') }}
             {%- else %}
                 {{ block('spanElement') }}
+
+                {% if item.children and secondLevel is defined and secondLevel %}
```
Steveb-p, line 91: It's possible that depth is available here, and could be used instead of `secondLevel` variable. It might suit it better?

src/bundle/Resources/public/scss/_main-menu.scss
```diff
@@ -117,6 +128,19 @@
...
             left: calculateRem(48px);
         }
+
+        &--group-name {
+            color: $ibexa-color-light-700;
+            font-size: $ibexa-text-font-size-small;
+
+            &:before {
+                content: "";
+                border-top: 1px solid $ibexa-color-info-800;
```
dew326, line 138:
```suggestion
                border-top: calculateRem(1px) solid $ibexa-color-info-800;
```

src/bundle/Resources/public/scss/_popup-menu.scss
```diff
@@ -21,6 +21,19 @@
...
         }
+
+        &--group-name {
+            flex-direction: column;
+            height: 100%;
+            align-items: initial;
+
+            &:before {
+                content: "";
+                border-top: 1px solid $ibexa-color-info-800;
```
dew326, line 32:
```suggestion
                border-top: calculateRem(1px) solid $ibexa-color-info-800;
```

src/bundle/Resources/public/scss/_main-menu.scss
```diff
@@ -117,6 +128,19 @@
             top: calculateRem(12px);
             left: calculateRem(48px);
         }
+
+        &--group-name {
+            color: $ibexa-color-light-700;
+            font-size: $ibexa-text-font-size-small;
+
+            &:before {
```
dew326, line 136: Why to add the border-top you have to use `before`? Should be added on the element itself.

src/bundle/Resources/public/scss/_popup-menu.scss
```diff
@@ -21,6 +21,19 @@
         &--hidden {
             display: none;
         }
+
+        &--group-name {
+            flex-direction: column;
+            height: 100%;
+            align-items: initial;
+
+            &:before {
```
dew326, line 30: The same, should be on the element not before.

dew326 (changes requested):

src/bundle/Resources/public/scss/_main-menu.scss
damianz5: it was a fix for border-top animation on menu toggler, I will remove it

dew326 (approved):

PR #2015, opened 2021-11-25
IBX-1514: Refactored 'read' UDW config resolvers

barw4:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-1514](https://jira.ez.no/browse/IBX-1514)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Additional `browse` configuration has been handled in the renamed `ReadAllowedContentTypesTest` class. This is to ensure that browse/read specific UDW actions are not overridden by `ContentRead` resolver which has nothing to do with browsing or searching UDW when not selecting any content (for example selecting a Subtree in Search module, or "Browse" button in main left menu).

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/ui-dev/src/modules/universal-discovery/components/content-type-selector/content.type.selector.js
+++ b/src/bundle/ui-dev/src/modules/universal-discovery/components/content-type-selector/content.type.selector.js
@@ -2,9 +2,11 @@ import React, { useState, useContext } from 'react';
 
 import { createCssClassNames } from '../../../common/helpers/css.class.names';
 import { SelectedContentTypesContext } from '../search/search';
+import { AllowedContentTypesContext } from '../../universal.discovery.module';
 
 const ContentTypeSelector = () => {
     const { contentTypes: contentTypesMap } = window.eZ.adminUiConfig;
+    const allowedContentTypes = useContext(AllowedContentTypesContext);
     const [selectedContentTypes, dispatchSelectedContentTypesAction] = useContext(SelectedContentTypesContext);
     const [collapsedGroups, setCollapsedGroups] = useState(() => {
         return Object.keys(contentTypesMap).reduce((collapsedGroups, contentTypeGroup, index) => {
@@ -28,6 +30,14 @@ const ContentTypeSelector = () => {
     return (
         <div className="ez-content-type-selector c-content-type-selector">
             {Object.entries(contentTypesMap).map(([contentTypeGroup, contentTypes]) => {
+                const isHidden = contentTypes.every(
+                    (contentType) => allowedContentTypes && !allowedContentTypes.includes(contentType.identifier)
+                );
+
+                if (isHidden) {
+                    return null;
+                }
+
                 const groupSelectorClassName = createCssClassNames({
                     'ez-content-type-selector__group': true,
                     'ez-content-type-selector__group--collapsed': collapsedGroups[contentTypeGroup],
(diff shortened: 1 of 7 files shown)


src/bundle/ui-dev/src/modules/universal-discovery/components/content-type-selector/content.type.selector.js
```diff
@@ -2,11 +2,9 @@ import React, { useState, useContext } from 'react';
 
 import { createCssClassNames } from '../../../common/helpers/css.class.names';
 import { SelectedContentTypesContext } from '../search/search';
-import { AllowedContentTypesContext } from '../../universal.discovery.module';
```
dew326: I'm not convinced of this approach.
`AllowedContentTypesContext` stores configuration for UDW which content types are allowed to be chosen from UDW, that's why we filter possible content types in search. And I think it should stay the same.
I think the problem is in the configuration/listeners themselves.

dew326 (changes requested):

mikadamczyk (approved):

src/lib/UniversalDiscovery/Event/Subscriber/ContentCreate.php
```diff
@@ -66,6 +66,10 @@ public static function getSubscribedEvents(): array
      */
     public function onUdwConfigResolve(ConfigResolveEvent $event): void
     {
+        if (in_array($event->getConfigName(), $event::READ_SPECIFIC_CONFIGURATIONS)) {
```
Steveb-p:
```suggestion
        if (in_array($event->getConfigName(), $event::READ_SPECIFIC_CONFIGURATIONS), true) {
```

src/lib/UniversalDiscovery/Event/Subscriber/ReadAllowedContentTypes.php
```diff
@@ -79,7 +79,7 @@ public function onUdwConfigResolve(ConfigResolveEvent $event): void
     {
         $config = $event->getConfig();
 
-        if (!in_array($event->getConfigName(), ['richtext_embed', 'richtext_embed_image'])) {
+        if (!in_array($event->getConfigName(), $event::READ_SPECIFIC_CONFIGURATIONS)) {
```
Steveb-p:
```suggestion
        if (!in_array($event->getConfigName(), $event::READ_SPECIFIC_CONFIGURATIONS), true) {
```

src/lib/UniversalDiscovery/Event/ConfigResolveEvent.php
```diff
@@ -14,6 +14,8 @@ class ConfigResolveEvent extends Event
 {
     public const NAME = 'udw.resolve.config';
 
+    public const READ_SPECIFIC_CONFIGURATIONS = ['richtext_embed', 'richtext_embed_image', 'browse'];
```
Steveb-p: I see this only used in context of checking whether the config name matches or does not match entries in this array.

If that's the case, I think it's more appropriate to introduce an event method instead?

Steveb-p (changes requested):

tischsoic (approved):

konradoboza (approved):

src/lib/UniversalDiscovery/Event/ConfigResolveEvent.php
Steveb-p:
```suggestion
    private const READ_SPECIFIC_CONFIGURATIONS = ['richtext_embed', 'richtext_embed_image', 'browse'];
```
:smiling_imp:

Steveb-p (approved):

pbisaga:
QA approved.
Check on Ibexa Commerce 3.3.x-dev.

mikadamczyk:
@barw4 you could merge it up and to ibexa

barw4:
Merge request to `ibexa/admin-ui/main`: https://github.com/ibexa/admin-ui/pull/98

PR #2022, opened 2021-12-20
[Behat] Added language selection coverage

tomaszszopinski:
JIRA: https://issues.ibexa.co/browse/IBX-1481
Related PRs: https://github.com/ezsystems/BehatBundle/pull/241 & https://github.com/ezsystems/ezplatform-workflow/pull/247
Successful Travis build: https://github.com/ezsystems/ezplatform-page-builder/pull/920

This PR adds option of selecting a language in which a certain CT is created in.

Zrzut ekranu 2021-12-22 o 11 47 56

### DIFF
--- a/src/lib/Behat/BrowserContext/ContentViewContext.php
+++ b/src/lib/Behat/BrowserContext/ContentViewContext.php
@@ -49,10 +49,11 @@ class ContentViewContext implements Context
 
     /**
      * @Given I start creating a new Content :contentType
+     * @Given I start creating a new Content :contentType in :language language
      */
-    public function startCreatingContent(string $contentType): void
+    public function startCreatingContent(string $contentType, string $language = null): void
     {
-        $this->contentViewPage->startCreatingContent($contentType);
+        $this->contentViewPage->startCreatingContent($contentType, $language);
     }
 
     /**
--- a/src/lib/Behat/Page/ContentUpdateItemPage.php
+++ b/src/lib/Behat/Page/ContentUpdateItemPage.php
@@ -93,6 +93,7 @@ class ContentUpdateItemPage extends Page
             new VisibleCSSLocator('closeButton', '.ez-content-edit-container__close'),
             new VisibleCSSLocator('fieldLabel', '.ez-field-edit__label-wrapper label.ez-field-edit__label, .ez-field-edit__label-wrapper legend, .ez-card > .card-body > div > div > legend'),
             new VisibleCSSLocator('nthField', '.ez-field-edit:nth-of-type(%s)'),
+            new VisibleCSSLocator('activeNthField', '.active .ez-field-edit:nth-of-type(%s)'),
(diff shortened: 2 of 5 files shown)


src/lib/Behat/Page/ContentViewPage.php
```diff
@@ -123,6 +123,14 @@ public function startCreatingContent(string $contentTypeName)
         $this->contentTypePicker->select($contentTypeName);
     }
 
+    public function startCreatingContentWithSelectedLanguage(string $contentTypeName, string $language)
```
mnocon: I'd do this by adding an optional parameter to the `startCreatingContent` method, so that it would look like this:
```
    public function startCreatingContent(string $contentTypeName, string $language = null)
    {
        $this->rightMenu->clickButton('Create');
        $this->contentTypePicker->verifyIsLoaded();
        if ($language !== null) {
              $this->contentTypePicker->selectLanguage($language);
        }
        $this->contentTypePicker->select($contentTypeName);
    }
```

src/lib/Behat/Page/ContentUpdateItemPage.php
```diff
@@ -163,4 +164,10 @@ public function switchToFieldGroup(string $tabName): void
             ->getByCriterion(new ElementTextCriterion($tabName))
             ->click();
     }
+
+    public function verifyFieldCannotBeEditedDueToLimitation(string $fieldName)
+    {
+       $fieldLocator = new VisibleCSSLocator('', sprintf($this->getLocator('activeNthField')->getSelector(), $this->getFieldPosition($fieldName)));
+       Assert::assertTrue($this->getHTMLPage()->find($fieldLocator)->hasClass('ez-field-edit--disabled'));
```
mnocon: I thought we talked about adding it to ElementAssert, so that we could use it like this:
```
$this->getHTMLPage()->find($fieldLocator)->assert()->hasClass('ez-field-edit--disabled')
```

We should avoid adding PHPUnit Assertions without custom error messages, otherwise this would fail with "Failed asserting that False is True", which is very cryptic.

Adding a new assertion to ElementAssert allows us to add a better default message.

src/lib/Behat/BrowserContext/ContentViewContext.php
```diff
@@ -55,6 +55,14 @@ public function startCreatingContent(string $contentType): void
         $this->contentViewPage->startCreatingContent($contentType);
     }
 
+    /**
+     * @Given I start creating a new Content :contentType with selected :language language
```
mnocon: With the optional language argument this would look like this:
```
    /**
     * @Given I start creating a new Content :contentType
     * @Given I start creating a new Content :contentType in :language language
     */
    public function startCreatingContent(string $contentType, string $language = null): void
    {
        $this->contentViewPage->startCreatingContent($contentType, $language);
    }
```
(I've changed `with selected :language language` to `in :language language`, because for me "in German language" sounds better)

mnocon (approved):

micszo (approved):
👍🏻

PR #2023, opened 2021-12-20
IBX-1552 As a QA I want spike test coverage for deep links

pbisaga:
| Question      | Answer
| ------------- | ---
| Tickets       | https://issues.ibexa.co/browse/IBX-1552
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes/no
| Doc needed?   | no

#### Checklist:
- [ ] Ready for Code Review

### DIFF
--- a/features/standard/Authentication.feature
+++ b/features/standard/Authentication.feature
@@ -17,3 +17,9 @@ Feature: Verify that Admin Panel is available only for authenticated users
     Given I open Login page in admin SiteAccess
     When I log in as admin with password notpublish
     Then I should be on Login page
+
+  @javascript
+  Scenario: Should be redirected to the same page in back office after login
+    Given I try to open "System Information" page in admin SiteAccess
+    When I log in as admin with password publish
+    Then I should be on "System Information" page


features/standard/Authentication.feature
```diff
@@ -17,3 +17,10 @@ Feature: Verify that Admin Panel is available only for authenticated users
     Given I open Login page in admin SiteAccess
     When I log in as admin with password notpublish
     Then I should be on Login page
+
+  @javascript 
+  Scenario: Should be redirected to the same page in back office after relogin
```
mnocon: I don't think the application behaves like that and it's a different Scenario than what's in https://issues.ibexa.co/browse/EZP-26238

I'd start with the ticket description:
```
For example, in eZ Publish 5.4.7, if you point your browser to:
http://example.com/ezdemo_site_admin/Tests/Article-1
...you will get to that page after login. In eZ Studio that no longer happens, you always get re-directed to the dashboard after login.
```

which we then could convert into a Scenario like this:
```
Given I open back office deep link page X
And I login as <USER>
Then I should be on Page X
```
Which we will have to adjust and reuse our existing Steps.

I think that there are two issues with the current Scenario:
1) It uses logout, which https://issues.ibexa.co/browse/EZP-26238 does not mention at all - imho it's best to focus on the initial Scenario first
2) It uses logout, but it never logs in before that - which seems incorrect

micszo (approved):
LGTM. 🙂

bogusez (approved):

mnocon (approved):

tomaszszopinski (approved):

PR #2028, opened 2022-01-18
IBX-1948: Log exceptions from admin-ui

vidarl:
| Question      | Answer
| ------------- | ---
| Tickets       | https://issues.ibexa.co/browse/IBX-1948
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes/no
| Doc needed?   | no

Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/config/services/events.yaml
+++ b/src/bundle/Resources/config/services/events.yaml
@@ -3,6 +3,8 @@ services:
         autowire: true
         autoconfigure: true
         public: false
+        bind:
+            string $logLevel: 'error'
 
     EzSystems\EzPlatformAdminUi\EventListener\:
         resource: "../../../lib/EventListener/*"
@@ -38,6 +40,7 @@ services:
             $entrypointLookupCollection: '@webpack_encore.entrypoint_lookup_collection'
         tags:
             - { name: kernel.event_listener, event: kernel.exception }
+            - { name: monolog.logger, channel: ibexa.admin }
 
     EzSystems\EzPlatformAdminUi\EventListener\MenuPermissionsListener:
         tags:
--- a/src/lib/EventListener/AdminExceptionListener.php
+++ b/src/lib/EventListener/AdminExceptionListener.php
@@ -11,6 +11,8 @@ namespace EzSystems\EzPlatformAdminUi\EventListener;
 use eZ\Publish\Core\MVC\Symfony\SiteAccess;
 use EzSystems\EzPlatformAdminUi\Notification\NotificationHandlerInterface;
 use EzSystems\EzPlatformAdminUiBundle\EzPlatformAdminUiBundle;
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerAwareTrait;
 use SplFileInfo;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\HttpFoundation\Response;
@@ -22,8 +24,10 @@ use Throwable;
 use Twig\Environment;
 use Twig\Error\RuntimeError;
 
-class AdminExceptionListener
+class AdminExceptionListener implements LoggerAwareInterface
 {
+    use LoggerAwareTrait;
+
     /** @var \EzSystems\EzPlatformAdminUi\Notification\NotificationHandlerInterface */
     protected $notificationHandler;
 
@@ -45,6 +49,9 @@ class AdminExceptionListener
     /** @var string */
     protected $kernelEnvironment;
 
+    /** @var \Psr\Log\LogLevel::* */
+    private $logLevel;
+
     /**
      * @param \Twig\Environment $twig
      * @param \EzSystems\EzPlatformAdminUi\Notification\NotificationHandlerInterface $notificationHandler
@@ -53,6 +60,7 @@ class AdminExceptionListener
      * @param array $siteAccessGroups
      * @param string $kernelRootDir
      * @param string $kernelEnvironment
+     * @param \Psr\Log\LogLevel::* $logLevel
      */
     public function __construct(
         Environment $twig,
@@ -61,7 +69,8 @@ class AdminExceptionListener
         EntrypointLookupCollectionInterface $entrypointLookupCollection,
         array $siteAccessGroups,
         string $kernelProjectDir,
-        string $kernelEnvironment
+        string $kernelEnvironment,
+        string $logLevel
     ) {
         $this->twig = $twig;
         $this->notificationHandler = $notificationHandler;
@@ -70,6 +79,7 @@ class AdminExceptionListener
         $this->siteAccessGroups = $siteAccessGroups;
         $this->rootDir = $kernelProjectDir;
         $this->kernelEnvironment = $kernelEnvironment;
+        $this->logLevel = $logLevel;
     }
 
     /**
@@ -99,6 +109,9 @@ class AdminExceptionListener
 
         // map exception to UI notification
         $this->notificationHandler->error(/** @Ignore */ $this->getNotificationMessage($exception));
+        $this->logger->log($this->logLevel, $exception->getMessage(), [
+            'exception' => $exception,
+        ]);
 
         if ($exception instanceof RuntimeError) {
             // If exception is coming from the template where encore already


src/lib/EventListener/AdminExceptionListener.php
```diff
@@ -45,6 +46,9 @@ class AdminExceptionListener
     /** @var string */
     protected $kernelEnvironment;
 
+    /** @var LoggerInterface */
+    protected $loggerInterface;
```
Steveb-p: Use `LoggerAwareTrait` & `LoggerAwareInterface` instead.

This will make it so this service will be automatically injected with the right logger by framework (there is a built-in compiler pass for that).

src/lib/EventListener/AdminExceptionListener.php
```diff
@@ -99,6 +106,7 @@ public function onKernelException(GetResponseForExceptionEvent $event)
 
         // map exception to UI notification
         $this->notificationHandler->error($this->getNotificationMessage($exception));
+        $this->loggerInterface->critical($exception->getMessage());
```
Steveb-p: Add exception object to context, so it can be properly recovered by tools like Sentry or DataDog, when their loggers are added to the chain.
```suggestion
        $this->loggerInterface->critical($exception->getMessage(), [
            'exception' => $e,
        ]);
```

Additionally, I'm not 100% certain that it is a critical level error all the time. In most cases these are reserved for PHP failures, syntax issues, and similar problems when the application straight up fails to launch - and immediate administrative action is required.

```suggestion
        $this->loggerInterface->error($exception->getMessage(), [
            'exception' => $e,
        ]);
```

adamwojs:
@vidarl could you please address @Steveb-p feedback ?

vidarl:
Hi @adamwojs . Yes, sorry about the delay. I'll have a look next week

src/lib/EventListener/AdminExceptionListener.php
vidarl: Fixed but `setLogger` was not called automatically and I didn't find that compiler pass that was supposed to do that. So I had to call `setLogger` explicitly in the service definition ref https://github.com/ezsystems/ezplatform-admin-ui/pull/2028/files#diff-0859788ec9d4b46a3c24ac17d11e648aecab1a297456cf1ae6084ffdef817e15R111-112

src/lib/EventListener/AdminExceptionListener.php
vidarl: I added the exception object to the context as suggested

vidarl: Regarding loglevel please see my comment in the PR description regarding merging
In principle, I agree to that log level should be error. 
However, the monolog config for 2.5 says that only messages with critical and alert levels are sent to the logs.
So if you insist, I can change log level to error but the end result will then still be that admin-ui errors will still not end up in the log in prod out-of-the-box

vidarl: @Steveb-p ^

vidarl:
> @vidarl could you please address @Steveb-p feedback ?

@adamwojs , I have done so, but I have some comments. Please see the unresolved threads.

src/lib/EventListener/AdminExceptionListener.php
```diff
@@ -45,6 +47,8 @@ class AdminExceptionListener
     /** @var string */
     protected $kernelEnvironment;
 
+    use LoggerAwareTrait;
```
adamwojs: Could you please move `use LoggerAwareTrait;` to the beginning of the class ?

src/lib/EventListener/AdminExceptionListener.php
Steveb-p: `setLogger` is called when service is tagged with `monolog.logger`.

https://symfony.com/doc/current/logging.html#using-a-logger-inside-a-service

> If your application uses [service autoconfiguration](https://symfony.com/doc/current/service_container.html#services-autoconfigure), any service whose class implements Psr\Log\LoggerAwareInterface will receive a call to its method setLogger() with the default logger service passed as a service.
>
> If you want to use in your own services a pre-configured logger which uses a specific channel (app by default), you can either [autowire monolog channels](https://symfony.com/doc/current/logging/channels_handlers.html#monolog-autowire-channels) or use the monolog.logger tag with the channel property as explained in the [Dependency Injection reference](https://symfony.com/doc/current/reference/dic_tags.html#dic_tags-monolog).

src/lib/EventListener/AdminExceptionListener.php
Steveb-p: @vidarl I would therefore propose a different approach, one that is present in the `Symfony\Component\HttpKernel\EventListener\ErrorListener::logKernelException`. 

Instead of hard coding a log level, we can control it using any sort of logic. Even the basic property with log level injected via constructor would do - although a full blown class map as done in Symfony default listener would be neat.

And by the way, the whole reason we even need to do this logging is because our listener is on the same priority as default, aforementioned `Symfony\Component\HttpKernel\EventListener\ErrorListener::logKernelException`. If it were lower, then we could simply re-use default Symfony listener for logging and adjust the class map as needed.

_EDIT: This means that since we `stopPropagation` of the event (as we handle the response), this means that default logging listener is not reached._

See excerpt from `debug:event` with `kernel.exception`.

```shell
php bin/console debug:event kernel.exception

Registered Listeners for "kernel.exception" Event
=================================================

 ------- --------------------------------------------------------------------------------------------------------- ---------- 
  Order   Callable                                                                                                  Priority  
 ------- --------------------------------------------------------------------------------------------------------- ---------- 
  #1      BabDev\PagerfantaBundle\EventListener\ConvertNotValidMaxPerPageToNotFoundListener::onKernelException()    512       
  #2      BabDev\PagerfantaBundle\EventListener\ConvertNotValidCurrentPageToNotFoundListener::onKernelException()   512       
  #3      Ibexa\Workflow\Event\Subscriber\VersionLockedExceptionListener::onVersionLocked()                         255       
  #4      Ibexa\Personalization\Event\Listener\InvalidInstallationKeyListener::onInvalidInstallationKey()           255       
  #5      Ibexa\Bundle\Rest\EventListener\ResponseListener::onKernelExceptionView()                                 20        
  #6      Ibexa\Bundle\Core\EventListener\ExceptionListener::onKernelException()                                    10        
  #7      Ibexa\HttpCache\EventSubscriber\HiddenLocationExceptionSubscriber::tagHiddenLocationExceptionResponse()   10        
  #8      Ibexa\PageBuilder\Event\Subscriber\ExceptionSubscriber::onKernelException()                               10        
  #9      Symfony\WebpackEncoreBundle\EventListener\ExceptionListener::onKernelException()                          0         
  #10     Ibexa\AdminUi\EventListener\AdminExceptionListener::onKernelException()                                   0         
  #11     Symfony\Component\HttpKernel\EventListener\ErrorListener::logKernelException()                            0         
  #12     Symfony\Component\HttpKernel\EventListener\ProfilerListener::onKernelException()                          0         
  #13     Symfony\Bundle\SwiftmailerBundle\EventListener\EmailSenderListener::onException()                         0         
  #14     FOS\HttpCacheBundle\EventListener\InvalidationListener::onKernelException()                               0         
  #15     Ibexa\Bundle\Core\EventListener\BackgroundIndexingTerminateListener::reindex()                            0         
  #16     Symfony\Component\HttpKernel\EventListener\RouterListener::onKernelException()                            -64       
  #17     Symfony\Component\HttpKernel\EventListener\ErrorListener::onKernelException()                             -128      
 ------- --------------------------------------------------------------------------------------------------------- ----------
 ```
 
 ```shell
 php bin/console debug:container exception_listener --show-arguments

Information for Service "exception_listener"
============================================

 ---------------- ---------------------------------------------------------- 
  Option           Value                                                     
 ---------------- ---------------------------------------------------------- 
  Service ID       exception_listener                                        
  Class            Symfony\Component\HttpKernel\EventListener\ErrorListener  
  Tags             kernel.event_subscriber                                   
                   monolog.logger (channel: request)                         
                   container.hot_path                                        
                   container.hot_path                                        
  Public           no                                                        
  Synthetic        no                                                        
  Lazy             no                                                        
  Shared           yes                                                       
  Abstract         no                                                        
  Autowired        no                                                        
  Autoconfigured   no                                                        
  Arguments        error_controller                                          
                   Service(monolog.logger.request)                           
                   1                                                         
                   Array (0 element(s))                                      
 ---------------- ----------------------------------------------------------
 ```

Comment (vidarl) [reply src/lib/EventListener/AdminExceptionListener.php]: @Steveb-p 
The purpose of the mapping done in `Symfony\Component\HttpKernel\EventListener\ErrorListener::logKernelException` is to have some logic so that some exceptions (decided by classname of the exception) are logged with log level X while others are logged with log level Y. I am not sure that is what we need in this use case

> Instead of hard coding a log level, we can control it using any sort of logic. Even the basic property with log level injected via constructor would do

So, I inject for instance "$logLevel" as parameter in the constructor, then we [bind](https://symfony.com/blog/new-in-symfony-3-4-local-service-binding) that value to "critical" in 2.5, but "error" in 3.x and later. Is this what you mean ?

Comment (vidarl) [reply src/lib/EventListener/AdminExceptionListener.php]: @Steveb-p  : still waiting for feedback on the question above.

Anyway, now that 2.5 is no longer maintained, I have rebased this to 2.3 ( Ibexa DXP 3.3 ).
As mentioned I think default log level should be error in 3.3 ( vs critical in 2.5 ).

But if you still want, we can make some logic around that as soon as you clarify my question asked in https://github.com/ezsystems/ezplatform-admin-ui/pull/2028#discussion_r864796618

File: src/lib/EventListener/AdminExceptionListener.php (hunk 3 above)
Comment (vidarl) [reply src/lib/EventListener/AdminExceptionListener.php]: Done in 9927157fd14c028217cf764721d030303003f378

File: src/lib/EventListener/AdminExceptionListener.php (hunk 2 above)
Comment (Steveb-p) [reply src/lib/EventListener/AdminExceptionListener.php]: Sorry @vidarl, I've missed the original response.

> So, I inject for instance "$logLevel" as parameter in the constructor, then we [bind](https://symfony.com/blog/new-in-symfony-3-4-local-service-binding) that value to "critical" in 2.5, but "error" in 3.x and later. Is this what you mean ?

Pretty much what I'd like to see.

Developers can configure their loggin in multitude of ways. One might treat one error type as critical (and require immediate attention, including notifying the admin in the middle of the night), while another might consider this an error that can be tackled later.

Providing ability to control precisely what is logged and how should be what we strive for. For this reason I'm suggesting passing error level as constructor value (even if it's not configurable via user-friendly configuration) and tagging the service with `monolog.logger` tag with proper channel attribute (so it can be excluded from general application errors).

Comment (vidarl) [reply src/lib/EventListener/AdminExceptionListener.php]: @Steveb-p  : Fixed in https://github.com/ezsystems/ezplatform-admin-ui/pull/2028/commits/c2932886c24944b0fbaf196d1e15d7d51db8d557

ok ?

File: src/lib/EventListener/AdminExceptionListener.php [hunk 4]
```diff
@@ -49,6 +49,9 @@ class AdminExceptionListener implements LoggerAwareInterface
     /** @var string */
     protected $kernelEnvironment;
 
+    /** @var mixed */
```
Steveb-p: I know that PSR does not contain this information in their phpdoc entries, but we should limit these values:

```suggestion
    /** @var \Psr\Log\LogLevel::* */
```

Indicating that we expect these values to be one of the constants defined in that class (which is true).

vidarl: Fixed in https://github.com/ezsystems/ezplatform-admin-ui/pull/2028/commits/7b852c64c76ed70521190e238b475459f36767fc

davidsayre:
Hi, How do I get this commit into my DXP build? I really need the admin UI to log errors becuase it's very difficult to troubleshoot.
Please review this commit and if passed/ merge.

src/lib/EventListener/AdminExceptionListener.php
```diff
@@ -61,7 +69,8 @@ public function __construct(
         EntrypointLookupCollectionInterface $entrypointLookupCollection,
         array $siteAccessGroups,
         string $kernelProjectDir,
-        string $kernelEnvironment
+        string $kernelEnvironment,
+        $logLevel
```
Steveb-p:
```suggestion
        string $logLevel
```

src/lib/EventListener/AdminExceptionListener.php
```diff
@@ -45,6 +49,9 @@ class AdminExceptionListener
     /** @var string */
     protected $kernelEnvironment;
 
+    /** @var \Psr\Log\LogLevel::* */
+    protected $logLevel;
```
Steveb-p: Already exposed in constructor (as an argument). We don't expect extending classes (if any) to access it.
```suggestion
    private $logLevel;
```

src/bundle/Resources/config/services/events.yaml
```diff
@@ -38,6 +40,7 @@ services:
             $entrypointLookupCollection: '@webpack_encore.entrypoint_lookup_collection'
         tags:
             - { name: kernel.event_listener, event: kernel.exception }
+            - { name: monolog.logger}
```
Steveb-p:
```suggestion
            - { name: monolog.logger, channel: ibexa.admin }
```

src/bundle/Resources/config/services/events.yaml
```diff
@@ -3,6 +3,8 @@ services:
         autowire: true
         autoconfigure: true
         public: false
+        bind:
+            $logLevel: 'error'
```
Steveb-p: That is a kinda risky bind. We may want different services to use different errors in different circumstances. It would be better to inject it directly into the relevant service.

Or, in lieu of that and in light of another one of my comments:
```suggestion
            string $logLevel: 'error'
```

Steveb-p:
> Hi, How do I get this commit into my DXP build? I really need the admin UI to log errors becuase it's very difficult to troubleshoot. Please review this commit and if passed/ merge.

@davidsayre Sorry, it kinda went under the radar.

As a workaround, you can create a service in your project, and make it listen to `kernel.exception` events, with priority higher than this listener (which is currently `0` priority, so at minimum `1`). Then [follow normal Symfony rules for injecting logger](https://symfony.com/doc/5.4/logging.html#using-a-logger-inside-a-service) and log the error in the same fashion as is done here.

vidarl:
Everything is resolved. Ready for merge now ? @Steveb-p

Steveb-p (approved):

ViniTou (approved):

webhdx (approved):

adamwojs (approved):

tomaszszopinski (approved):

PR #2030, opened 2022-01-24
IBX-1999: Added pagination to URL Wildcard tab in URL Management

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-1999](https://issues.ibexa.co/browse/IBX-1999)
| Bug fix?      | no
| New feature?  | yes/no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/themes/admin/url_wildcard/list.html.twig
+++ b/src/bundle/Resources/views/themes/admin/url_wildcard/list.html.twig
@@ -73,8 +73,8 @@
                 </tr>
             </thead>
             <tbody>
-                {% if url_wildcards|length %}
-                    {% for url_wildcard in url_wildcards %}
+                {% if pager.currentPageResults|length %}
+                    {% for url_wildcard in pager.currentPageResults %}
                         {% set edit_url = path('ezplatform.url_wildcard.update', {urlWildcardId: url_wildcard.id}) %}
                         <tr>
                             <td class="ez-table__cell ez-table__cell--has-checkbox">
@@ -119,5 +119,15 @@
                 {% endif %}
             </tbody>
         </table>
+    {% if pager.haveToPaginate %}
+        {% include '@ezdesign/ui/pagination.html.twig' with {
+            'pager': pager,
+            'paginaton_params' : pager_options|merge({
+                'routeParams': {
+                    '_fragment': 'ez-tab-link-manager-url-wildcards'
+                }
+            })
+        } %}
+    {% endif %}
     {{ form_end(form) }}
 </section>
--- a/src/bundle/Resources/views/themes/admin/url_wildcard/list.html.twig
+++ b/src/bundle/Resources/views/themes/admin/url_wildcard/list.html.twig
@@ -73,8 +73,16 @@
                 </tr>
             </thead>
             <tbody>
-                {% if pager.currentPageResults|length %}
-                    {% for url_wildcard in pager.currentPageResults %}
+                {% if url_wildcards.count is same as(0) %}
+                    <tr>
+                        <td class="ez-table__cell ez-table__cell--no-content" colspan="5">
+                            <p class="ez-table-no-content mb-0 py-0">
+                                {{ 'url_wildcard.empty'|trans|desc('No URL wildcards.') }}
+                            </p>
+                        </td>
+                    </tr>
+                {% else %}
+                    {% for url_wildcard in url_wildcards %}
                         {% set edit_url = path('ezplatform.url_wildcard.update', {urlWildcardId: url_wildcard.id}) %}
                         <tr>
                             <td class="ez-table__cell ez-table__cell--has-checkbox">
@@ -108,20 +116,12 @@
                             </td>
                         </tr>
                     {% endfor %}
-                {% else %}
-                    <tr>
-                        <td class="ez-table__cell ez-table__cell--no-content" colspan="5">
-                            <p class="ez-table-no-content mb-0 py-0">
-                                {{ 'url_wildcard.empty'|trans|desc('No URL wildcards.') }}
-                            </p>
-                        </td>
-                    </tr>
                 {% endif %}
             </tbody>
         </table>
-    {% if pager.haveToPaginate %}
+    {% if url_wildcards.haveToPaginate %}
         {% include '@ezdesign/ui/pagination.html.twig' with {
-            'pager': pager,
+            'pager': url_wildcards,
             'paginaton_params' : pager_options|merge({
                 'routeParams': {
                     '_fragment': 'ez-tab-link-manager-url-wildcards'
--- a/src/lib/Pagination/Pagerfanta/URLWildcardAdapter.php
+++ b/src/lib/Pagination/Pagerfanta/URLWildcardAdapter.php
@@ -1,0 +1,55 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformAdminUi\Pagination\Pagerfanta;
+
+use eZ\Publish\API\Repository\URLWildcardService;
+use Pagerfanta\Adapter\AdapterInterface;
+
+/**
+ * Pagerfanta adapter for eZ Publish content search.
+ * Will return results as notification list.
+ */
+class URLWildcardAdapter implements AdapterInterface
+{
+    /** @var \eZ\Publish\API\Repository\URLWildcardService */
+    private $urlWildcardService;
+
+    /** @var int */
+    private $nbResults;
+
+    public function __construct(URLWildcardService $urlWildcardService)
+    {
+        $this->urlWildcardService = $urlWildcardService;
+    }
+
+    /**
+     * Returns the number of results.
+     *
+     * @return int the number of results
+     */
+    public function getNbResults(): int
+    {
+        if ($this->nbResults !== null) {
+            return $this->nbResults;
+        }
+
+        return $this->nbResults = $this->urlWildcardService->countAll();
+    }
+
+    /**
+     * Returns a slice of the results.
+     *
+     * @param int $offset the offset
+     * @param int $length the length
+     *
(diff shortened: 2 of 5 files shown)


konradoboza (approved):

Steveb-p (approved):

src/lib/Tab/URLManagement/URLWildcardsTab.php
```diff
@@ -87,8 +97,20 @@ public function getOrder(): int
...
     {
-        $urlWildcards = $this->urlWildcardService->loadAll();
+        $currentPage = $this->requestStack->getCurrentRequest()->query->getInt(
+            self::PAGINATION_PARAM_NAME, 1
+        );
+        $limit = $this->configResolver->getParameter('pagination.url_wildcards');
+
+        $pagerfanta = new Pagerfanta(
+            new ArrayAdapter(
+                $this->urlWildcardService->loadAll()
```
Steveb-p: Did you know that `URLWildcardService::loadAll()` actually has two parameters that in theory should allow proper pagination? :sweat_smile: 

EDIT: I mean we would still need to allow `COUNT` query to be performed for this to make sense in Handler & Gateway, but it's doable.

mateuszdebinski: and yes and no;), because soon there will be PR that will add a search engine to it, so the loadAll function with these parameters will not be used

as for the count function, you're right, I'll add it in a moment :)

Steveb-p (approved):

konradoboza (approved):

src/bundle/Resources/views/themes/admin/url_wildcard/list.html.twig
```diff
@@ -73,8 +73,8 @@
                 </tr>
             </thead>
             <tbody>
-                {% if url_wildcards|length %}
-                    {% for url_wildcard in url_wildcards %}
+                {% if pager.currentPageResults|length %}
+                    {% for url_wildcard in pager.currentPageResults %}
```
adamwojs:
```suggestion
                    {% for url_wildcard in pager %}
```

Pagerfanta objects are iterable.

src/bundle/Resources/views/themes/admin/url_wildcard/list.html.twig
```diff
@@ -73,8 +73,8 @@
                 </tr>
             </thead>
             <tbody>
-                {% if url_wildcards|length %}
-                    {% for url_wildcard in url_wildcards %}
+                {% if pager.currentPageResults|length %}
```
adamwojs:
```suggestion
                {% if pager|length %}
```

Pagerfanta objects are countable. Also you  use getNbResults method.

src/lib/Pagination/Pagerfanta/URLWildcardAdapter.php
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformAdminUi\Pagination\Pagerfanta;
+
+use eZ\Publish\API\Repository\URLWildcardService;
+use Pagerfanta\Adapter\AdapterInterface;
+
+class URLWildcardAdapter implements AdapterInterface
```
adamwojs: A) Unit tests are missing
B) Class could be `final`

src/lib/Pagination/Pagerfanta/URLWildcardAdapter.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformAdminUi\Pagination\Pagerfanta;
```
adamwojs: Legacy namespace

src/lib/Tab/URLManagement/URLWildcardsTab.php
```diff
@@ -103,7 +125,10 @@ public function renderView(array $parameters): string
         $canManageWildcards = $this->permissionResolver->hasAccess('content', 'urltranslator');
 
         return $this->twig->render('@ezdesign/url_wildcard/list.html.twig', [
-            'url_wildcards' => $urlWildcards,
+            'pager' => $pagerfanta,
```
adamwojs: Could you please keep the old parameter name ?

adamwojs (changes requested):

src/bundle/Resources/views/themes/admin/url_wildcard/list.html.twig
mateuszdebinski: changed to `count is same as(0)` like in another tabs in admin-ui

src/lib/Tests/Pagination/Pagerfanta/URLWildcardAdapterTest.php
```diff
@@ -0,0 +1,81 @@
...
+{
+    /** @var \eZ\Publish\API\Repository\Values\Content\UrlWildcard|\PHPUnit\Framework\MockObject\MockObject */
+    private $urlWildcardService;
+
+    protected function setUp(): void
+    {
+        $this->urlWildcardService = $this->createMock(URLWildcardService::class);
+    }
+
+    public function testGetNbResults()
```
Steveb-p:
```suggestion
    public function testGetNbResults(): void
```

src/lib/Tests/Pagination/Pagerfanta/URLWildcardAdapterTest.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        $countAll = 5;
+
+        $this->urlWildcardService
+            ->expects($this->once())
+            ->method('countAll')
+            ->with()
```
Steveb-p: Empty `with`, which in practice is an assertion definition, can & should be skipped.
```suggestion
```

src/lib/Tests/Pagination/Pagerfanta/URLWildcardAdapterTest.php
```diff
@@ -0,0 +1,81 @@
...
+            ->willReturn($countAll);
+
+        $adapter = new URLWildcardAdapter($this->urlWildcardService);
+
+        $this->assertEquals(
+            $countAll,
+            $adapter->getNbResults()
```
Steveb-p, line 37: We could also assert that calling `getNbResults` does **NOT** load the results, i.e. `urlWildcardService::loadAll` method is not called. The reverse is true for tests that follow this one: when loading a result slice, we expect method `urlWildcardService::countAll()` to not be called as well.

Additionally, subsequent calls to `$adapter->getNbResults()` should not call the `urlWildcardService::countAll` method, as seen in your implementation where you store the result in property.
(just call the `$adapter->getNbResults()` a second time here, which will trigger the `self::once()` method call counter assertion if mock is called a second time).

src/lib/Tests/Pagination/Pagerfanta/URLWildcardAdapterTest.php
```diff
@@ -0,0 +1,81 @@
...
+        );
+    }
+
+    public function testGetSlice()
```
Steveb-p:
```suggestion
    public function testGetSlice(): void
```

src/lib/Tests/Pagination/Pagerfanta/URLWildcardAdapterTest.php
```diff
@@ -0,0 +1,81 @@
...
+            ->method('loadAll')
+            ->with($offset, $limit)
+            ->willReturn([$this->urlWildcards()]);
+
+        $adapter = new URLWildcardAdapter($this->urlWildcardService);
+
+        $this->assertEquals(
+            [$this->urlWildcards()],
+            $adapter->getSlice($offset, $limit)
+        );
```
Steveb-p: I understand what you want to do here, but in this case the comparison will simply be wrong (albeit it passes tests because the Generators that are actually the only element of the array are identical).

In actuality, the array being compared is:
```php
[
    <Generator>,
];
```
instead of the expected:
```php
[
    <UrlWildcard>,
    <UrlWildcard>,
];
```

What you can do is unpack the Generator:
```suggestion
        $expectedResult = iterator_to_array($this->urlWildcards());
        $this->urlWildcardService
            ->expects($this->once())
            ->method('loadAll')
            ->with($offset, $limit)
            ->willReturn($expectedResult);

        $adapter = new URLWildcardAdapter($this->urlWildcardService);

        $this->assertEquals(
            $expectedResult,
            $adapter->getSlice($offset, $limit)
        );
```
or drop the Generator (`iterable`) entirely and return a simple array in `urlWildcards` method.

Steveb-p (approved):
Tests should be updated, but otherwise than that :+1:

src/lib/Tests/Pagination/Pagerfanta/URLWildcardAdapterTest.php
```diff
@@ -0,0 +1,86 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzPlatformAdminUi\Tests\Pagination\Pagerfanta;
+
+use eZ\Publish\API\Repository\URLWildcardService;
+use eZ\Publish\API\Repository\Values\Content\URLWildcard;
+use Ibexa\AdminUi\Pagination\Pagerfanta\URLWildcardAdapter;
+use PHPUnit\Framework\TestCase;
+
+class URLWildcardAdapterTest extends TestCase
```
Steveb-p:
```suggestion
final class URLWildcardAdapterTest extends TestCase
```

Steveb-p (approved):

adamwojs (approved):

Steveb-p (approved):

bogusez (approved):

PR #2037, opened 2022-03-23
[Behat] Enabled map location tests and stabilize Richtext

mnocon:
JIRA: https://issues.ibexa.co/browse/IBX-1545

1) The map locations tests have been disabled some time ago because of network stability issues. I think it's time to bring them back.
2) I've increased the timeout and added a direct click into the richtext to avoid the `Exception: Timeout, CSS selector 'styleDropdown': '.ae-toolbar-element' not found in 1 seconds.` issue . Tested in https://github.com/ezsystems/ezplatform-page-builder/pull/939

PR for v4: https://github.com/ibexa/admin-ui/pull/375 (richtext changes are not needed - CKEditor is used there)


micszo (approved):
Thanks!

tomaszszopinski (approved):

bogusez (approved):

PR #2039, opened 2022-04-07
IBX-2275: Refactor UDW configuration subscribers

barw4:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-2275](https://issues.ibexa.co/browse/IBX-2275)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

From the UDW standpoint this might look tricky because usually the swap/move actions are validated using targets like:
`$this->permissionResolver->canUser('content', 'create', $location->getContentInfo(), [$newParentLocation])`, so the outcome can differ for each content. Therefore we cannot lock this permission to `content/create` as `allowedContentTypes` are global for UDW. The check will be done after all in the API.

//Edit Okay, at this point I feel like `content/create` should be used only for `create` configuration. Therefore, solution has been refactored.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/UniversalDiscovery/Event/ConfigResolveEvent.php
+++ b/src/lib/UniversalDiscovery/Event/ConfigResolveEvent.php
@@ -14,16 +14,6 @@ class ConfigResolveEvent extends Event
 {
     public const NAME = 'udw.resolve.config';
 
-    private const READ_SPECIFIC_CONFIGURATIONS = [
-        'richtext_embed',
-        'richtext_embed_image',
-        'browse',
-        'subtree_search',
-        'single',
-        'multiple',
-        'single_container',
-    ];
-
     /** @var string */
     protected $configName;
 
@@ -79,10 +69,5 @@ class ConfigResolveEvent extends Event
     public function setContext(array $context): void
     {
         $this->context = $context;
-    }
-
-    public function isReadOnlyEvent(): bool
-    {
-        return in_array($this->getConfigName(), self::READ_SPECIFIC_CONFIGURATIONS, true);
     }
 }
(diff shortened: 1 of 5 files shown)


src/lib/UniversalDiscovery/Event/ConfigResolveEvent.php
```diff
@@ -21,6 +21,7 @@ class ConfigResolveEvent extends Event
         'subtree_search',
         'single',
         'multiple',
+        'single_container',
```
Nattfarinn: I don't think we should enforce read-only on `single_container` configuration (and we shouldn't enforce it on `single` and `multiple` either). If a feature requires specific treatment it should us own configuration. Now it is global change and affects all places where `single_container` is used.

If there is an issue with move/swap it means they should use own configuration (will it be copy of `single_container` or not) and that configuration could be read-only.

The same applies for the `single` and `multiple` above. If there was an issue with relations, relations should use own configuration instead of global ones (and we shouldn't make global ones read-only).

Nattfarinn (changes requested):

ViniTou:
@barw4 
It looks like related CI failure here.

barw4:
After discussion with @Nattfarinn we decided to use `ContentCreate` class to use its validation only for `create` UDW configuration and use `ReadAllowedContentTypes` for **every single** configuration.

The tests had to be adapted as well:
- we allow CotF only for `create` - https://github.com/ezsystems/ezplatform-admin-ui/pull/2039/files#diff-df396750fe1d0e35705176f608089e33c78fce98808343821b960d4b6aadbe51R49
- `ReadAllowedContentTypes` would be validated **always**, because we must have `content/read` permission policy every time we want to use UDW (it doesn't make sense otherwise), so we assume that `ReadAllowedContentTypes` affects **all** configurations - therefore https://github.com/ezsystems/ezplatform-admin-ui/pull/2039/files#diff-af3417d4c261ddc7d6e6ab5a92feefefb976de7b801466358ab190e6112cf5d8R56.

ViniTou (approved):

Nattfarinn (approved):

tomaszszopinski (approved):

PR #2047, opened 2022-05-23
IBX-857: Added call invalidateSize on each map when tab clicked and if tab contains any map

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-857](https://issues.ibexa.co/browse/IBX-857)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Because the map is not visible immediately (display:none;) the map can't compute the correct position, therefore the call invalidateSize(true); has been added for maps when changing the tab, thanks to which it is able to calculate the correct position

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezgmaplocation.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezgmaplocation.js
@@ -1,4 +1,4 @@
-(function(global, doc, eZ, Leaflet) {
+(function(global, doc, eZ, Leaflet, $) {
     const SELECTOR_FIELD = '.ez-field-edit--ezgmaplocation';
     const SELECTOR_ADDRESS_INPUT = '.ez-data-source__field--address .ez-data-source__input';
     const SELECTOR_LAT_FIELD = '.ez-data-source__field--latitude';
@@ -18,6 +18,7 @@
     const VALIDATE_LONGITUDE = 'validateLongitude';
     const VALIDATE_LATITUDE = 'validateLatitude';
     const VALIDATE_ADDRESS = 'validateAddress';
+    const maps = [];
 
     class EzGMapLocationValidator extends eZ.BaseFieldValidator {
         /**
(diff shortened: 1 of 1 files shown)


dew326 (approved):

src/bundle/Resources/public/js/scripts/fieldType/ezgmaplocation.js
```diff
@@ -618,5 +620,14 @@
         }
     });
 
+    $('.ez-tabs a').on('shown.bs.tab', (event) => {
+        const id = event.target.getAttribute('href');
+        const element = doc.querySelector(id);
+
+        if (element.querySelectorAll(SELECTOR_FIELD).length > 0) {
+            maps.forEach(map => map.invalidateSize(true));
```
tischsoic: code formatting (Prettier)
```suggestion
            maps.forEach((map) => map.invalidateSize(true));
```

src/bundle/Resources/public/js/scripts/fieldType/ezgmaplocation.js
```diff
@@ -618,5 +620,14 @@
         }
     });
 
+    $('.ez-tabs a').on('shown.bs.tab', (event) => {
+        const id = event.target.getAttribute('href');
+        const element = doc.querySelector(id);
```
tischsoic:
```suggestion
        const tabPaneSelector = event.target.getAttribute('href');
        const tabPane = doc.querySelector(tabPaneSelector);
```

src/bundle/Resources/public/js/scripts/fieldType/ezgmaplocation.js
```diff
@@ -618,5 +620,14 @@
         }
     });
 
+    $('.ez-tabs a').on('shown.bs.tab', (event) => {
```
tischsoic: We avoid using elements selectors and tend to use CSS classes instead:
```suggestion
    $('.ez-tabs .nav-link').on('shown.bs.tab', (event) => {
```

tischsoic (changes requested):

OstafinL (approved):

tischsoic (approved):

GrabowskiM (approved):

bogusez (approved):

PR #2061, opened 2022-08-12
IBX-2559: Added new style to ez-extra-actions__content class to allow scrolling language list

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-2559](https://issues.ibexa.co/browse/IBX-2559)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Added new class to scrolling language lists

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/themes/admin/content/widget/content_edit.html.twig
+++ b/src/bundle/Resources/views/themes/admin/content/widget/content_edit.html.twig
@@ -2,7 +2,7 @@
 
 <div class="ez-extra-actions ez-extra-actions--edit ez-extra-actions--hidden {% if form.language.vars.choices|length == 1 %} ez-extra-actions--prevent-show {% endif %}" data-actions="edit">
     <div class="ez-extra-actions__header">{{ 'content.edit.select_language'|trans|desc('Select translation') }} ({{ form.language.vars.choices|length }})</div>
-    <div class="ez-extra-actions__content">
+    <div class="ez-extra-actions__content ez-extra-actions__content__scroller">
         {{ form_start(form, { 'action': path('ezplatform.content.edit') }) }}
         {{ form_widget(form.content_info, { 'attr': {
             'hidden': 'hidden',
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/views/themes/admin/content/widget/content_edit.html.twig
```diff
@@ -2,7 +2,7 @@
 
 <div class="ez-extra-actions ez-extra-actions--edit ez-extra-actions--hidden {% if form.language.vars.choices|length == 1 %} ez-extra-actions--prevent-show {% endif %}" data-actions="edit">
     <div class="ez-extra-actions__header">{{ 'content.edit.select_language'|trans|desc('Select translation') }} ({{ form.language.vars.choices|length }})</div>
-    <div class="ez-extra-actions__content">
+    <div class="ez-extra-actions__content ez-extra-actions__content__scroller">
```
OstafinL: Why not just?
```suggestion
    <div class="ez-extra-actions__content">
```
You have selector in https://github.com/ezsystems/ezplatform-admin-ui/blob/b048b14e7ecc8af1d62813c47aa4b63f2f4eb579/src/bundle/Resources/public/scss/_extra-actions.scss#L54

mateuszdebinski: there is more than one place where this class is used, so I didn't want to change it everywhere. However, if there is no problem with it, I will move code to __content selector

dew326: BEM element cannot be parent and child at the same time. In that case, you should use a modifier.
`ez-extra-actions__content--scroller`

edit: maybe in other places there is the same problem and it should be done globally on __content as Łukasz suggested? But this is something to check.

GrabowskiM (approved):
Now PR title isn't correct, as you don't add any new class

OstafinL (approved):

bogusez (approved):

PR #2063, opened 2022-08-17
IBX-1648: Added allow_extra_fields to allow remove image with extra form fields

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-1648](https://issues.ibexa.co/browse/IBX-1648)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Added allow_extra_fields to allow send a form with extra data like confirm that "I understand the consequences of this action." even if that field is no need anymore and move to trash item

#### Checklist:
- [x] Ready for Code Review


adamwojs (changes requested):
IMHO scenario described in JIRA issue is invalid. Have you discussed/confirmed what is desired behavior with ENGINEERING/PM team?

PR #2065, opened 2022-08-22
IBX-3681 - Added selectedLanguage update on location change in UDW

mateuszbieniek:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-3681](https://issues.ibexa.co/browse/IBX-3681)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

When a location is clicked inside UDW the languageSelected is set by
```
    const [selectedLanguage, setSelectedLanguage] = useState(preselectedLanguage || firstLanguageCode);
```
Unfortunately, `useState` only sets the value of the hook and won't update it with the next elements clicked, which can end up in a different Language used by Content on the Fly, when the default element in the Language Selected form is the one that user wants to use (so no `onClick` event).
The change introduced ensures that `languageSelected` const has a proper value set.

### DIFF
--- a/src/bundle/ui-dev/src/modules/universal-discovery/components/content-create-widget/content.create.widget.js
+++ b/src/bundle/ui-dev/src/modules/universal-discovery/components/content-create-widget/content.create.widget.js
@@ -78,6 +78,9 @@ const ContentCreateWidget = () => {
     });
 
     useEffect(() => {
+        setSelectedLanguage(preselectedLanguage || firstLanguageCode)
+    }, [preselectedLanguage, firstLanguageCode]);
+    useEffect(() => {
         window.eZ.helpers.tooltips.parse(refContentTree.current);
     }, []);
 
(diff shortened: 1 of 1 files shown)


src/bundle/ui-dev/src/modules/universal-discovery/components/content-create-widget/content.create.widget.js
```diff
@@ -77,6 +77,9 @@ const ContentCreateWidget = () => {
         'c-content-create--hidden': !createContentVisible,
     });
 
+    useEffect(() => {
+        setSelectedLanguage(preselectedLanguage || firstLanguageCode)
+    }, [preselectedLanguage, firstLanguageCode]);
```
GrabowskiM, line 82:
```suggestion
    }, [preselectedLanguage, firstLanguageCode]);
    
```

GrabowskiM (changes requested):

GrabowskiM (approved):

OstafinL (approved):

tomaszszopinski:
After patch appliance original error is fixed, but 404 error appears when Article CT is opened (step 8 in Jira)

https://user-images.githubusercontent.com/59650405/186133281-fc7361b3-a77c-4128-8b4b-414da66ead63.mov

.

tomaszszopinski (approved):

mateuszbieniek:
> After patch appliance original error is fixed, but 404 error appears when Article CT is opened (step 8 in Jira)
> 2065_1.mov
> 
> .

Issue not related to the fix:
https://issues.ibexa.co/browse/IBX-3713
The fix only makes it easier to surface.

PR #2074, opened 2022-10-25
IBX-1648: Added redirect when the trash form has any error

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| Tickets       | https://issues.ibexa.co/browse/IBX-1648
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   |no

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Controller/LocationController.php
+++ b/src/bundle/Controller/LocationController.php
@@ -354,10 +354,10 @@ class LocationController extends Controller
             if ($result instanceof Response) {
                 return $result;
             }
-    
+
             $data = $form->getData();
             $location = $data->getLocation();
-    
+
             return $this->redirectToLocation($location);
         }
 
--- a/src/bundle/Controller/LocationController.php
+++ b/src/bundle/Controller/LocationController.php
@@ -357,11 +357,10 @@ class LocationController extends Controller
 
             $data = $form->getData();
             $location = $data->getLocation();
-            
+
             if ($location !== null) {
                 return $this->redirectToLocation($location);
             }
-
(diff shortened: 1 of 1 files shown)


micszo:
@mateuszdebinski does the second commit intentionally contain a different JIRA id?

mateuszdebinski:
> @mateuszdebinski does the second commit intentionally contain a different JIRA id?

Thanks, corrected, it's wasn't intentional

mikadamczyk (approved):

src/bundle/Controller/LocationController.php
```diff
@@ -354,6 +354,11 @@ public function trashAction(Request $request): Response
             if ($result instanceof Response) {
                 return $result;
             }
+
+            $data = $form->getData();
+            $location = $data->getLocation();
+
+            return $this->redirectToLocation($location);
```
ViniTou: What if `$location` is null?

mateuszdebinski: Is it at all possible for `$location` to be empty when operating on location?

but of course, I can add 'if' if it doesn't find a location then move it to the trash as it is now

mateuszdebinski: Ok never mind I checked and it's possible. Added condition to check if the location is not null

ViniTou (approved):

konradoboza (approved):

bogusez (approved):

PR #2077, opened 2022-11-14
IBX-4078: Moment.js localization

barw4:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-4078](https://issues.ibexa.co/browse/IBX-4078)
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

The language will be picked up from the User's Back Office settings.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/admin.format.date.js
+++ b/src/bundle/Resources/public/js/scripts/admin.format.date.js
@@ -1,4 +1,8 @@
-(function(moment) {
+(function(moment, eZ) {
+    const { backOfficeLanguage } = eZ.adminUiConfig;
+
+    moment.locale(backOfficeLanguage);
+
     /*
         ([yqLdDeEcaZ])\1* -> find any pattern of one or repeated one of these characters
         or
@@ -134,4 +138,4 @@
 
         return this.format(form);
     };
-})(window.moment);
+})(window.moment, window.eZ);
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/public/js/scripts/admin.format.date.js
```diff
@@ -1,4 +1,8 @@
-(function(moment) {
+(function(moment, eZ) {
+    const { backOfficeLanguage } = eZ.adminUiConfig;
+
+    moment.locale(backOfficeLanguage);
```
tischsoic, line 4: Not sure, but maybe this should be set in the header in `src/bundle/Resources/views/themes/admin/ui/layout.html.twig` so that it will be set globally?

barw4: @tischsoic Aren't we kinda setting it globally? This is the same file that adds additional methods to moment.js and is fired after loading dom. All `window.moment` calls we are utilizing will have such locale assigned.

tischsoic: Ok, you are right, this file seems to fit this code. :)

tischsoic (approved):

dew326 (approved):

bogusez (approved):

PR #2078, opened 2022-11-16
IBX-3814: Added checks if the image is not SVG to prevent adding alternativeText

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-3814](https://issues.ibexa.co/browse/IBX-3814)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Related to https://github.com/ezsystems/ezplatform-rest/pull/98

Currently, the response from rest API has been changed to add information about fields that are of the ezimage type, then we verify whether the mapped field is in the new field with rest API, which will allow us to add the alternativeText field more dynamically

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/ui-dev/src/modules/multi-file-upload/services/multi.file.upload.service.js
+++ b/src/bundle/ui-dev/src/modules/multi-file-upload/services/multi.file.upload.service.js
@@ -147,7 +147,7 @@ const prepareStruct = ({ parentInfo, config, languageCode }, data) => {
                 data: data.fileReader.result.replace(/^.*;base64,/, ''),
             };
 
-            if (data.file.type.startsWith('image/')) {
+            if (data.file.type.startsWith('image/') && !data.file.type.startsWith('image/svg')) {
                 fileValue.alternativeText = data.file.name;
             }
 
--- a/src/bundle/ui-dev/src/modules/multi-file-upload/services/multi.file.upload.service.js
+++ b/src/bundle/ui-dev/src/modules/multi-file-upload/services/multi.file.upload.service.js
@@ -147,7 +147,9 @@ const prepareStruct = ({ parentInfo, config, languageCode }, data) => {
                 data: data.fileReader.result.replace(/^.*;base64,/, ''),
             };
 
-            if (data.file.type.startsWith('image/') && !data.file.type.startsWith('image/svg')) {
+            const contentType = response.ContentTypeInfoList.ContentType[0];
+
+            if (contentType.imageFields.identifier.includes(mapping.contentFieldIdentifier)) {
                 fileValue.alternativeText = data.file.name;
             }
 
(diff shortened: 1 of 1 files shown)


Gengar-i (approved):

src/bundle/ui-dev/src/modules/multi-file-upload/services/multi.file.upload.service.js
```diff
@@ -147,7 +147,7 @@ const prepareStruct = ({ parentInfo, config, languageCode }, data) => {
                 data: data.fileReader.result.replace(/^.*;base64,/, ''),
             };
 
-            if (data.file.type.startsWith('image/')) {
+            if (data.file.type.startsWith('image/') && !data.file.type.startsWith('image/svg')) {
```
dew326: so if someone will change the default MFU config to use the image content type for SVG he will not get the alternative text?

dew326 (changes requested):

src/bundle/ui-dev/src/modules/multi-file-upload/services/multi.file.upload.service.js
mateuszdebinski: in that case, it will not have the alt text added which is set as the filename anyway. However, if it is required, it will occur in error like now. So maybe let's edit the information returned from rest whether the indicated type has the ezimage field (and add alternativeText based on it)?

dew326: Yes, we should have a more reliable way of checking if the alternative text should be added.

src/bundle/ui-dev/src/modules/multi-file-upload/services/multi.file.upload.service.js
```diff
@@ -147,7 +147,9 @@ const prepareStruct = ({ parentInfo, config, languageCode }, data) => {
                 data: data.fileReader.result.replace(/^.*;base64,/, ''),
             };
 
-            if (data.file.type.startsWith('image/')) {
+            const contentType = response.ContentTypeInfoList.ContentType[0];
+
+            if (contentType.imageFields.identifier.includes(mapping.contentFieldIdentifier)) {
```
tischsoic, line 152: Can we assign this condition to a variable for better readability? :) 
```suggestion
            const isImageContentType = contentType.imageFields.identifier.includes(mapping.contentFieldIdentifier)

            if (isImageContentType) {
```

tischsoic (approved):

src/bundle/ui-dev/src/modules/multi-file-upload/services/multi.file.upload.service.js
ViniTou: :+1:, but it seems that `isImageContentType` is not best name, as `ImageContentType` is usually related to `Image` content type, and from what i understand it is more `contentTypeContainsImageFieldType`

dew326 (approved):

mateuszdebinski:
As a new solution for rest has been introduced and the code in admin-ui PR will look completely different, I close it to open a new improved one

PR #2080, opened 2022-11-21
IBX-4130: Optimized subitems fetching in UDW

barw4:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-4130](https://issues.ibexa.co/browse/IBX-4130)
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

The mentioned change will prevent from resolving a `CurrentVersion` object that has resolved all content fields.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/UniversalDiscovery/UniversalDiscoveryProvider.php
+++ b/src/lib/UniversalDiscovery/UniversalDiscoveryProvider.php
@@ -22,6 +22,7 @@ use EzSystems\EzPlatformAdminUi\Permission\LookupLimitationsTransformer;
 use EzSystems\EzPlatformAdminUi\Permission\PermissionCheckerInterface;
 use EzSystems\EzPlatformAdminUi\QueryType\LocationPathQueryType;
 use EzSystems\EzPlatformRest\Output\Visitor;
+use EzSystems\EzPlatformRest\Server\Values\RestLocation;
 use EzSystems\EzPlatformRest\Server\Values\Version;
 
 class UniversalDiscoveryProvider implements Provider
@@ -248,7 +249,14 @@ class UniversalDiscoveryProvider implements Provider
         return [
             'locations' => array_map(
                 function (SearchHit $searchHit) {
-                    return $this->getRestFormat($searchHit->valueObject);
+                    /** @var Location $location */
+                    $location = $searchHit->valueObject;
+                    $restLocation = new RestLocation(
+                        $location,
+                        $this->locationService->getLocationChildCount($location)
+                    );
+
+                    return $this->getRestFormat($restLocation);
                 },
                 $searchResult->searchHits
             ),
--- a/src/lib/UniversalDiscovery/UniversalDiscoveryProvider.php
+++ b/src/lib/UniversalDiscovery/UniversalDiscoveryProvider.php
@@ -249,7 +249,7 @@ class UniversalDiscoveryProvider implements Provider
         return [
             'locations' => array_map(
                 function (SearchHit $searchHit) {
-                    /** @var Location $location */
+                    /** @var \eZ\Publish\API\Repository\Values\Content\Location $location */
                     $location = $searchHit->valueObject;
                     $restLocation = new RestLocation(
                         $location,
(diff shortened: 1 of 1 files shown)


src/lib/UniversalDiscovery/UniversalDiscoveryProvider.php
```diff
@@ -248,7 +249,14 @@ public function getSubitemLocations(
         return [
             'locations' => array_map(
                 function (SearchHit $searchHit) {
-                    return $this->getRestFormat($searchHit->valueObject);
+                    /** @var \eZ\Publish\API\Repository\Values\Content\Location $location */
+                    $location = $searchHit->valueObject;
+                    $restLocation = new RestLocation(
+                        $location,
+                        $this->locationService->getLocationChildCount($location)
```
konradoboza: Have you checked whether necessity of calling `SearchService` for children count in a loop isn't more expensive in the end?

barw4: It definitely isn't as expensive as fetching all fields in a given content. Perhaps I could just put 0 there as this information is not necessary at all 🤔 
//Edit -> that's what I did, seems like we don't really need info on a `childCount` here.

konradoboza: AFAICS there is a dedicated visitor exposing `childCount`, so not really, ref: https://github.com/ezsystems/ezplatform-rest/blob/1.3/src/lib/Server/Output/ValueObjectVisitor/RestLocation.php#L89.

barw4: Yes, I'm aware of the visitor, it will be 0 but why would we need the correct value there? It's not used in subitems list in UDW anywhere.

barw4: Besides, this call was made also in the previous solution (in `ValueObjectVisitor\Location`) so this won't affect performance negatively.

konradoboza: It seems so, yes. I am up for keeping the API call instead of `0` then, performance-wise we should be good here.

barw4: I would say this is arguable, UDW doesn't have to know how many children the current node has, we are not changing it globally but only for the UDW action. Calling it 50 times (as this is the default limit) as you have noticed can be respectfully damaging from the performance side. Hardcoding it to 0 shouldn't hurt us in any way.

konradoboza: If we decide to go this way, we should at least make sure that this property isn't visited in `ibexa/rest` then as it would show wrong data. Not sure if we aren't breaking the BC policy though with that approach.

barw4: I don't think I follow. In other places/calls the value of `childCount` would be properly set. We are changing it locally only for the single UDW action that fetches subitems.

konradoboza: Indeed, it seems this is only used internally, so saving unnecessary `SearchService` calls seems to be a way to go. +1 from my side.

konradoboza (approved):

webhdx (approved):
I'd let myself better explain this change as it may not be that apparent: `UniversalDiscoveryProvider` is first doing Location search using `SearchService` and then convert the results into JSON output compatible with our `ibexa/rest` server. Previously it was converting bare `Location` objects which wasn't correct as REST server uses `RestLocation` wrappers. This is what @barw4 did. I'm not too sure wether we should force `$childCount` to 0 in that case but AFAIR it'd require another database fetch to get this information which of course has performance penalty.

micszo (approved):
QA Approved on Ibexa Commerce 3.3.29-dev.

Tested `starting_location_id` on different levels in tree.

mikadamczyk (approved):

PR #2084, opened 2023-01-27
IBX-4486: Disabled Autosave feature for content creation

barw4:
| Question      | Answer
| ------------- | ---
| Tickets       | https://issues.ibexa.co/browse/IBX-4486
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Autosave shouldn't be enabled for `nodraft` requests.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/Form/Processor/Content/AutosaveProcessor.php
+++ b/src/lib/Form/Processor/Content/AutosaveProcessor.php
@@ -47,7 +47,7 @@ class AutosaveProcessor implements EventSubscriberInterface
         }
 
         $event->setResponse(
-        // Response content is irrelevant as it will be overwritten by ViewRenderer anyway
+            // Response content is irrelevant as it will be overwritten by ViewRenderer anyway
             new Response(null, $statusCode)
         );
     }
--- a/src/lib/Form/Processor/Content/AutosaveProcessor.php
+++ b/src/lib/Form/Processor/Content/AutosaveProcessor.php
@@ -10,6 +10,7 @@ namespace EzSystems\EzPlatformAdminUi\Form\Processor\Content;
 
 use eZ\Publish\API\Repository\Exceptions\Exception as APIException;
 use EzSystems\EzPlatformAdminUi\Event\AutosaveEvents;
+use EzSystems\EzPlatformContentForms\Data\Content\ContentUpdateData;
 use EzSystems\EzPlatformContentForms\Event\FormActionEvent;
 use EzSystems\EzPlatformContentForms\Form\Processor\ContentFormProcessor;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
@@ -35,15 +36,18 @@ class AutosaveProcessor implements EventSubscriberInterface
 
     public function processAutosave(FormActionEvent $event): void
     {
+        $statusCode = Response::HTTP_OK;
+
         try {
-            $this->innerContentFormProcessor->processSaveDraft($event);
-            $statusCode = Response::HTTP_OK;
+            if ($event->getData() instanceof ContentUpdateData) {
+                $this->innerContentFormProcessor->processSaveDraft($event);
+            }
         } catch (APIException $exception) {
             $statusCode = Response::HTTP_BAD_REQUEST;
         }
 
         $event->setResponse(
-            // Response content is irrelevant as it will be overwritten by ViewRenderer anyway
+        // Response content is irrelevant as it will be overwritten by ViewRenderer anyway
             new Response(null, $statusCode)
         );
     }


src/lib/Form/Processor/Content/AutosaveProcessor.php
```diff
@@ -35,9 +36,12 @@ public static function getSubscribedEvents(): array
 
     public function processAutosave(FormActionEvent $event): void
     {
+        $statusCode = Response::HTTP_OK;
+
         try {
-            $this->innerContentFormProcessor->processSaveDraft($event);
-            $statusCode = Response::HTTP_OK;
+            if ($event->getData() instanceof ContentUpdateData) {
```
konradoboza, line 42: Is this a reliable way to detect `/nodraft` requests? Can we have some test coverage to be sure that we didn't disabled autosave for cases that shouldn't be affected by this change?

barw4: @konradoboza I can create some tests, but what other cases do you have in mind that shouldn't be affected by this change? In `processSaveDraft` we explicitly mention only two cases - `ContentCreateData` and `ContentUpdateData` (https://github.com/ezsystems/ezplatform-content-forms/blob/1.3/src/lib/Form/Processor/ContentFormProcessor.php#L83). It's even more reasonable to block other unexpected `Datas` here and leave only the one fitting.

konradoboza: I assume that with this change we don't block autosaving while being on content creation form, right? The naming itself is a bit hard to wrap a head around without context. That's why I asked about some test coverage. Nevertheless, if it isn't the case, I am fine with proposed changes.

barw4: If by "content creation form" you mean creation of content that is in fact updating its first draft then it is not blocking it. That's how Back Office works usually. This PR only affects the "nodraft" content creation when a draft for a given content doesn't exist yet and therefore there is no need to update/autosave it.

konradoboza: >If by content creation form you mean creation of content that is in fact updating its first draft

That was indeed my assumption, but wasn't sure.

konradoboza (approved):

src/lib/Form/Processor/Content/AutosaveProcessor.php
barw4: Just adding to the above comment:
clicking on "+ Create" in Back Office creates a draft and then redirects us to updating such a draft unless this is a `nodraft` URL.

ViniTou: This form has `drafts_enabled` option, wouldnt that be a better way to check it? It just seems it is set wrong for /nodraft action and it is only used in UDW.

@Nattfarinn

Nattfarinn: I guess best would be to make this:  https://github.com/ezsystems/ezplatform-content-forms/blob/1.3/src/lib/Content/View/Filter/ContentCreateViewFilter.php#L132 optional parameter and set it to `false` here: https://github.com/ezsystems/ezplatform-content-forms/blob/1.3/src/lib/Content/View/Filter/ContentCreateViewFilter.php#L87

This should give us reliable information whether drafts are enabled or not.

Nattfarinn (changes requested):

src/lib/Form/Processor/Content/AutosaveProcessor.php
barw4: @ViniTou @Nattfarinn setting it to `false` will break content create forms - https://github.com/ezsystems/ezplatform-content-forms/blob/1.3/src/lib/Form/Type/Content/ContentEditType.php#L46-L65.

In the `create.html.twig` (https://github.com/ezsystems/ezplatform-admin-ui/blob/2.3/src/bundle/Resources/views/themes/admin/content/create/create.html.twig#L29-L30) we are using both cancel and save draft buttons. So perhaps we could introduce another option that is `autosave_enabled`?

ViniTou: yup, for me form option sounds better then `instance of` check.

On the other hand, if it only breaks displaying of buttons, then maybe it is ok. If people wants to have `Save draft` button, when using `/nodraft` path, that contradicts a little to each other but I guess thats another discussion.

barw4: @ViniTou @Nattfarinn Closing this in favor of https://github.com/ezsystems/ezplatform-content-forms/pull/66

PR #2104, opened 2023-06-22
IBX-6017: Avoid loading whole content in the Content Tree

barw4:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-6017](https://issues.ibexa.co/browse/IBX-6017)
| Improvement?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

The idea is that we could implement `loadVersionInfos` PAPI method and get translated name based on this, just like it's done in `loadContentListByContentInfo` - this would have a much higher impact on performance though.

Blackfire profiles are included in the ticket's comments.

Requires:  https://github.com/ezsystems/ezplatform-kernel/pull/375

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/lib/UI/Module/ContentTree/NodeFactory.php
+++ b/src/lib/UI/Module/ContentTree/NodeFactory.php
@@ -79,15 +79,22 @@ final class NodeFactory
     ): Node {
         $uninitializedContentInfoList = [];
         $containerLocations = [];
-        $node = $this->buildNode($location, $uninitializedContentInfoList, $containerLocations, $loadSubtreeRequestNode, $loadChildren, $depth, $sortClause, $sortOrder);
-        $contentById = $this->contentService->loadContentListByContentInfo($uninitializedContentInfoList);
+        $node = $this->buildNode(
+            $location,
+            $uninitializedContentInfoList,
+            $containerLocations,
+            $loadSubtreeRequestNode,
+            $loadChildren,
+            $depth,
+            $sortClause,
+            $sortOrder
+        );
 
         $aggregatedChildrenCount = null;
         if ($this->searchService->supports(SearchService::CAPABILITY_AGGREGATIONS)) {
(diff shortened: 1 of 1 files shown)


webhdx:
BC is kept as long as we still return any name. I vaguely remember why it was done this way but my guess would be that it was required to fetch Content name in current siteaccess prioritized language. With your change you likely return the name in Content's main language. We need to make sure there is no regression here.

barw4:
> BC is kept as long as we still return any name. I vaguely remember why it was done this way but my guess would be that it was required to fetch Content name in current siteaccess prioritized language. With your change you likely return the name in Content's main language. We need to make sure there is no regression here.

@webhdx yes, this is exactly the case, we are prioritizing the first language defined in the current site access. And with the change following I'm returning the name in the main language of content, as you described.

webhdx:
We are talking about regression then. I'd say this improvement in current form is a no go as another customer will report that all of sudden they are not getting translated names. A feature toggle is also a flawed idea because you get performance gain at the cost of missing a feature.

I'd rather explore the possibility of loading `VersionInfo` list as you described. But I personally see implementing node caching would work the best here but we are talking about larger feature than just a quick fix.

barw4:
@webhdx related PR is ready with `versionInfoList` as discussed https://github.com/ezsystems/ezplatform-kernel/pull/375

webhdx (approved):

alongosz (approved):
I'm assuming behat failure is related to missing kernel dependency :)

barw4:
> I'm assuming behat failure is related to missing kernel dependency :)

@alongosz, yes, most errors are coming from the missing method (`loadVersionInfoListByContentInfo`) that was introduced in the kernel package.

konradoboza (approved):

micszo (approved):
Test status in https://github.com/ezsystems/ezplatform-kernel/pull/375.

PR #2106, opened 2023-07-11
IBX-5852: Created a new layout for error pages to prevent leak data

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-5852](https://issues.ibexa.co/browse/IBX-5852)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Changed the layout that error pages use so that by default they use as few of our components as possible and haven't too much data like window.ibexa

#### Checklist:
- [x] Ready for Code Review


src/bundle/Resources/views/themes/admin/ui/layout_error.html.twig
```diff
@@ -0,0 +1,48 @@
...
+    <head>
+        <meta charset="UTF-8" />
+        {% block meta %}
+        {% endblock %}
+        {% if app.request.locale == 'ach_UG' %}
+            <script type="text/javascript">
+                var _jipt = [];
+                _jipt.push(['project', 'ezplatform']);
+            </script>
+            <script type="text/javascript" src="//cdn.crowdin.com/jipt/jipt.js"></script>
```
glye, line 12: Comment on the security hotspot alert for this? https://sonarcloud.io/project/security_hotspots?id=ezsystems_ezplatform-admin-ui&pullRequest=2106&resolved=false&types=SECURITY_HOTSPOT

glye (commented):
Not sure yet, please look into the hotspot alert.

src/bundle/Resources/views/themes/admin/ui/layout_error.html.twig
glye: If this integrity check can be used, it would be good: https://sonarcloud.io/organizations/ezsystems/rules?open=Web%3AS5725&rule_key=Web%3AS5725&tab=how_to_fix

mateuszdebinski: this code was from our default layout: https://github.com/ezsystems/ezplatform-admin-ui/blob/2.3/src/bundle/Resources/views/themes/admin/ui/layout.html.twig#L88

src/bundle/Resources/public/js/scripts/admin.error.page.js
```diff
@@ -0,0 +1,35 @@
...
+
+        const notificationNode = container.querySelector('.alert');
+
+        notificationsContainer.append(notificationNode);
+    };
+
+    Object.entries(notifications).forEach(([label, messages]) => {
+        messages.forEach((message) => addNotification({ detail: { label, message } }));
+    });
+})(window.document);
```
Gengar-i: newline

Gengar-i (approved):

src/bundle/Resources/views/themes/admin/ui/layout_error.html.twig
glye: DO NOT COMMIT, just an example
```suggestion
            <script type="text/javascript" src="//cdn.crowdin.com/version-x.y.z/jipt/jipt.js" integrity="sha384-ye83f7h4...etc"></script>
```
So this alert is not from this PR, but I'm trying to resolve it anyway. The subresource integrity check makes sense here. Making the hash is easy. But we need to lock it to a specific version of the JS code, so it doesn't break on the next update. I haven't yet found how crowdin would specify the version.
https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity#tools_for_generating_sri_hashes

glye: NB: This should not hold up this PR - but please look into it as a follow-up, if merged without a fix for this.

src/bundle/Resources/views/themes/admin/ui/layout_error.html.twig
```diff
@@ -0,0 +1,48 @@
...
+        {% endif %}
+        <title>{% block title %}Ibexa DXP{% endblock %}</title>
+        {{ encore_entry_link_tags('ezplatform-admin-ui-layout-css', null, 'ezplatform') }}
+	    <script src="{{ asset('bundles/ezplatformadminuiassets/vendors/jquery/dist/jquery.min.js') }}"></script>
+	    <script src="{{ asset('bundles/ezplatformadminuiassets/vendors/bootstrap/dist/js/bootstrap.min.js') }}"></script>
```
dew326: indent

src/bundle/Resources/views/themes/admin/ui/layout_error.html.twig
```diff
@@ -0,0 +1,48 @@
...
+
+        <div
+            class="ez-notifications-container"
+            data-notifications="{{ app.flashes|json_encode() }}"
+            data-template="{{ include('@ezdesign/ui/notification.html.twig', {
+                label: '{{ label }}',
+                message: '{{ message }}',
+                badge: '{{ badge }}'
+            })|e('html_attr')  }}"></div>
+        <div class="ez-modal-wrapper"></div>
```
dew326: do we need modal wrapper?

src/bundle/Resources/views/themes/admin/ui/error_page/403.html.twig
```diff
@@ -1,4 +1,4 @@
-{% extends '@ezdesign/ui/layout.html.twig' %}
+{% extends '@ezdesign/ui/layout_error.html.twig' %}
 
 {% block navigation %}{% endblock %}
```
dew326: if this block does not exist in layout_error you probably can remove it from those templates

dew326 (changes requested):

tischsoic (approved):

dew326 (approved):

src/bundle/Resources/views/themes/admin/ui/layout_error.html.twig
mateuszdebinski: @glye I asked Crowdin about this and here is an answer:

"The team told me that we don’t have versions of jipt.js. Bug fixes or new features may be introduced to the file"

glye: Thanks. That's unfortunate. The only way to secure it is to bundle it, then, and ensure we update it during release cycles.

Anyway, this problem should not block this PR since it's pre-existing.

mnocon (approved):
1) I've tested that we're not leaking any information through JavaScript objects
2) Error pages in the Admin panel are working as intended

QA Approved!

PR #2110, opened 2023-08-31
IBX-6451: Calculating the size of the dropdown element has been changed

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-6451](https://issues.ibexa.co/browse/IBX-6451)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

as itemsContainerTop may be too large relative to the remaining height of the bottom element at times. Therefore, a calculation was added on how much is left from the top of the modal to a given element. To check whether this value is greater or smaller than itemsContainerTop, also I did not subtract DROPDOWN_MARGIN from this value, because the margin is the size of the element

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
+++ b/src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
@@ -142,14 +142,9 @@
             }
 
             if (this.itemsContainer.closest(SELECTOR_MODAL)) {
-                const heightModal = this.itemsContainer.closest('.modal-content').getBoundingClientRect().height;
-                const leftToBottom = heightModal - itemsContainerTop;
+                const modalHeight = this.itemsContainer.closest('.modal-content').getBoundingClientRect().height;
 
-                if (leftToBottom < itemsContainerTop) {
-                    return leftToBottom;
-                }
-
-                return itemsContainerTop - DROPDOWN_MARGIN;
+                return modalHeight - itemsContainerTop;
             }
 
             return documentElementHeight - itemsContainerTop - DROPDOWN_MARGIN;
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
```diff
@@ -142,6 +142,13 @@
             }
 
             if (this.itemsContainer.closest(SELECTOR_MODAL)) {
+                const heightModal = this.itemsContainer.closest('.modal-content').getBoundingClientRect().height;
+                const leftToBottom = heightModal - itemsContainerTop;
```
GrabowskiM:
```suggestion
                const modalHeight = this.itemsContainer.closest('.modal-content').getBoundingClientRect().height;
                const leftToBottom = modalHeight - itemsContainerTop;
```

GrabowskiM: Also I'm not perfectly sure about `leftToBottom` variable, it sounds quite vague for me

GrabowskiM (changes requested):

src/bundle/Resources/public/js/scripts/core/custom.dropdown.js
mateuszdebinski: @GrabowskiM maybe heightToBottom?
I have no idea how to name this variable

GrabowskiM (approved):

dew326 (approved):

bogusez (approved):

PR #2116, opened 2024-03-18
IBX-7954: Show error message on empty image asset

barw4:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-7954](https://issues.ibexa.co/browse/IBX-7954)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

In case a chosen related asset has no image instead of JS console error we should show a message.

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
+++ b/src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
@@ -98,6 +98,18 @@
                 return field.fieldDefinitionIdentifier === imageAssetMapping['contentFieldIdentifier'];
             });
 
+            if (imageField.fieldValue === null) {
+                showErrorNotification(
+                    Translator.trans(
+                        /* @Desc("The chosen asset has no image data available.") */ 'ezimageasset.empty_data.message.error',
+                        {},
+                        'fieldtypes_preview'
+                    )
+                );
+
+                return;
+            }
+
             this.updateData(
                 response.ContentInfo.Content._id,
                 response.ContentInfo.Content.TranslatedName,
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js
```diff
@@ -126,6 +126,18 @@
          * @param {Object} image
          */
         updateData(destinationContentId, destinationContentName, destinationLocationId, image) {
+            if (image === null) {
```
GrabowskiM: I have some concers regarding flow - we use this method currently it two places: 

https://github.com/ezsystems/ezplatform-admin-ui/blob/fd71c454ecf2e05225f27e0d3b060d7b5ff485b4/src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js#L101 here probably everything will be ok if image is null

https://github.com/ezsystems/ezplatform-admin-ui/blob/fd71c454ecf2e05225f27e0d3b060d7b5ff485b4/src/bundle/Resources/public/js/scripts/fieldType/ezimageasset.js#L62 - but here should you have `response.value` empty you'll show both error AND success notification.

If it's not possible for line 62 to have empty value and all you want to fix is line 101 (in `loadAsset` method) I think you should patch it directly in `loadAsset` before calling for `updateData`. 

Current flow can be little misleading, as one of possible paths indicates that in can be both success and error (even if it can't be reached in live situation)

OstafinL (approved):

dew326 (approved):

Gengar-i (approved):

GrabowskiM (approved):

bogusez (approved):

PR #2117, opened 2024-03-26
IBX-7983: Handled previewing `ezimage` field with height of 0

barw4:
| Question      | Answer
| ------------- | ---
| Tickets       | [IBX-7983](https://issues.ibexa.co/browse/IBX-7983)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Tests pass?   | yes
| Doc needed?   | no

Division by `0` is impossible, hence, the change.

 `.avif` files in later PHP versions should be handled properly by the `getimagesize` method (since PHP 8.1 RC3), but we still have to handle older files somehow (dimension values are stored in .xmls in a database).

#### Checklist:
- [x] Ready for Code Review

### DIFF
--- a/src/bundle/Resources/views/themes/admin/ui/field_type/preview/content_fields.html.twig
+++ b/src/bundle/Resources/views/themes/admin/ui/field_type/preview/content_fields.html.twig
@@ -433,10 +433,12 @@
                         <td>{{ 'ezimage.master_dimensions'|trans|desc('Master dimensions') }}:</td>
                         <td>{{ 'ezimage.width_and_height'|trans({'%width%': field.value.width, '%height%': field.value.height})|desc('Width: %width%px height: %height%px') }}</td>
                     </tr>
-                    <tr class="ez-field-preview__meta-value-row">
-                        <td>{{ 'ezimage.ratio'|trans|desc('Ratio') }}:</td>
-                        <td>{{ (field.value.width/field.value.height)|round(2) }}</td>
-                    </tr>
+                    {% if (field.value.height is not same as '0') %}
+                        <tr class="ez-field-preview__meta-value-row">
+                            <td>{{ 'ezimage.ratio'|trans|desc('Ratio') }}:</td>
+                            <td>{{ (field.value.width/field.value.height)|round(2) }}</td>
+                        </tr>
+                    {% endif %}
                 </tbody>
             </table>
         </div>
--- a/src/bundle/Resources/views/themes/admin/ui/field_type/preview/content_fields.html.twig
+++ b/src/bundle/Resources/views/themes/admin/ui/field_type/preview/content_fields.html.twig
@@ -433,7 +433,7 @@
                         <td>{{ 'ezimage.master_dimensions'|trans|desc('Master dimensions') }}:</td>
                         <td>{{ 'ezimage.width_and_height'|trans({'%width%': field.value.width, '%height%': field.value.height})|desc('Width: %width%px height: %height%px') }}</td>
                     </tr>
-                    {% if (field.value.height != '0') %}
+                    {% if field.value.height != '0' %}
                         <tr class="ez-field-preview__meta-value-row">
                             <td>{{ 'ezimage.ratio'|trans|desc('Ratio') }}:</td>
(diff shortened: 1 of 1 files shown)


konradoboza (approved):
Is there any chance that height will be e.g. `0.00` or `0` perhaps? I was wondering if any sanitization/casting on the backend side makes sense.

barw4:
@konradoboza I don't think there is a need as those are already parsed as strings https://github.com/ezsystems/ezplatform-kernel/blob/1.3/eZ/Publish/Core/Persistence/Legacy/Content/FieldValue/Converter/ImageConverter.php#L115

src/bundle/Resources/views/themes/admin/ui/field_type/preview/content_fields.html.twig
```diff
@@ -433,10 +433,12 @@
                         <td>{{ 'ezimage.master_dimensions'|trans|desc('Master dimensions') }}:</td>
                         <td>{{ 'ezimage.width_and_height'|trans({'%width%': field.value.width, '%height%': field.value.height})|desc('Width: %width%px height: %height%px') }}</td>
                     </tr>
-                    <tr class="ez-field-preview__meta-value-row">
-                        <td>{{ 'ezimage.ratio'|trans|desc('Ratio') }}:</td>
-                        <td>{{ (field.value.width/field.value.height)|round(2) }}</td>
-                    </tr>
+                    {% if (field.value.height is not same as '0') %}
```
webhdx: Is `field.value.height` a string? Can we avoid string comparison here?

tischsoic: I think that Twig actually casts variables to numbers when using `==`. 
I'm not 100% sure, but e.g `'0.0' == 0` is _true_ in Twig. So maybe we can just do this: 🤔 
```suggestion
                    {% if field.value.height == 0 %}
```

tischsoic: And I wonder, maybe _Ratio_ field should still be visible but its value instead should be set to something like `n/a`? But I don't know whether it would make any sense, just asking :)

Steveb-p: > I think that Twig actually casts variables to numbers when using `==`. I'm not 100% sure, but e.g `'0.0' == 0` is _true_ in Twig.

`is same as` is a strict-type comparison. `==` is a loose comparison.

In general, strict comparison is preferred. In this case it makes sense to make it lose, **if value coming from storage can be of different types**. So it comes down to what we actually can get here as value.

If you can get `"0"`, `0` and `"0.00"`, then a lose comparison approach is somewhat acceptable.

barw4: Resolved in https://github.com/ezsystems/ezplatform-admin-ui/pull/2117/commits/8108b918551f27dcef98453cc8dcafce504ddc3f

src/bundle/Resources/views/themes/admin/ui/field_type/preview/content_fields.html.twig
```diff
@@ -433,10 +433,12 @@
                         <td>{{ 'ezimage.master_dimensions'|trans|desc('Master dimensions') }}:</td>
                         <td>{{ 'ezimage.width_and_height'|trans({'%width%': field.value.width, '%height%': field.value.height})|desc('Width: %width%px height: %height%px') }}</td>
                     </tr>
-                    <tr class="ez-field-preview__meta-value-row">
-                        <td>{{ 'ezimage.ratio'|trans|desc('Ratio') }}:</td>
-                        <td>{{ (field.value.width/field.value.height)|round(2) }}</td>
-                    </tr>
+                    {% if (field.value.height != '0') %}
```
Steveb-p:
```suggestion
                    {% if field.value.height != '0' %}
```

tischsoic (approved):

dew326 (approved):

